PR #61, opened 2022-09-16
Setup linkcars capability, and required decoder.

ice-breaker-tg:
Setup linkcars capability.
Setup decoder for array of links in UCAN.

should the decoder maybe get pulled up to ucanto, or 
should we make a more generic method of caveats that are a 'Array\<T\>'?


Gozala:
close in favor of #81

PR #62, opened 2022-09-16
chore: migarte w3-store from ucanto branch

### LINKED ISSUE #52: migrate other w3 packages here
I think we should migrate remaining w3 packages here as having them in ucanto makes no sense, currently they live at

https://github.com/web3-storage/ucanto/tree/w3/w3/

Gozala:
This just migrates all the packages from https://github.com/web3-storage/ucanto/tree/w3/w3 to here except CLI which got superseeded by https://github.com/web3-storage/w3up-cli

Fixes #52

### DIFF
--- a/.github/release-please-config.json
+++ b/.github/release-please-config.json
@@ -3,6 +3,8 @@
   "packages": {
     "packages/access": {},
     "packages/access-api": {},
-    "packages/wallet": {}
+    "packages/wallet": {},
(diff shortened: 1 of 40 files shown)


Gozala:
@ice-breaker-tg I have moved stuff from ucanto into here so we can iterate over interfaces and capabilities in a single repo

PR #77, opened 2022-09-28
redirect on success

### LINKED ISSUE #71: Done page should be a full branded HTML page
We should congratulate you for joining web3 storage and point you to resources you can use to get started.

https://github.com/web3-storage/w3-protocol/blob/305b2d317ba4b8743a1594e9dbe0d22bac90c229/packages/access-api/src/routes/validate.js#L12

jchris:
this might be all we need to close #71

### DIFF
--- a/packages/access-api/src/routes/validate.js
+++ b/packages/access-api/src/routes/validate.js
@@ -9,7 +9,7 @@ export async function validate(req, env) {
   if (req.query && req.query.ucan) {
     await validations.create(req.query.ucan)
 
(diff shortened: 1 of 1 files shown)


jchris:
connected with Hugo, we should have a page with a message similar to this, so lets figure out how also add some helpful resources and a coupon code to this basic message:

Screen Shot 2022-09-28 at 2 54 10 PM

hugomrdias (approved):

PR #79, opened 2022-09-30
fix store/* capabilities to allow omiting optionals

Gozala:


ice-breaker-tg (approved):
This all looks good to me.

Before deployment, we need to discuss if this will cause issues with current CLI delegations / w3up.

PR #119, opened 2022-10-24
sequence diagrams: registration, upload, and list

jchris:
I think these will require more than one round of feedback, so it probably makes sense to merge them more or less as is, if the file layout makes sense and they should be in this repository.

Then we can do additional refinement in PRs to the markdown files.


jchris (commented):
I think this is ready to merge :)

PR #120, opened 2022-10-25
feat: add IndexedDB Store implementation

alanshaw:
An IndexedDB store implementation that uses RSA keys for the browser.


packages/access/src/stores/store-indexeddb.js
```diff
@@ -0,0 +1,227 @@
...
+
+  /** @type {IDBDatabase|undefined} */
+  #db
+
+  /**
+   * @param {string} dbName
+   * @param {number} [dbVersion]
+   */
+  constructor(dbName, dbVersion) {
+    this.#dbName = dbName
```
Gozala, line 34: Are these widely enough available for us to use ? If I recall correctly they were considered in js-multiformats but then dropped because adoption was not there yet.

alanshaw: I can put underscores...

hugomrdias (approved):
LGTM

packages/access/src/stores/store-indexeddb.js
alanshaw: I think it's fine!

Screenshot 2022-10-26 at 18 19 14

PR #136, opened 2022-11-04
fix: upload/add capability root validation

alanshaw:
@gozala should `root` be the root CID of the data, not a CAR CID? This is how it's implemented AFAIK. Or were you thinking it was the CID of the CAR that has the root data CID?

This PR switches `root` to allow _any_ CID, not just a CAR CID.

Also, for a separate PR, maybe neither of these fields should be optional. Like, an "upload" should always have a data root CID and be present in at least 1 shard?

### DIFF
--- a/packages/access/src/capabilities/upload.js
+++ b/packages/access/src/capabilities/upload.js
@@ -17,16 +17,16 @@ export const upload = any.derive({
 })
 
 // Right now ucanto does not yet has native `*` support, which means
-// `store/add` can not be derived from `*` event though it can be
-// derived from `store/*`. As a workaround we just define base capability
+// `upload/add` can not be derived from `*` event though it can be
+// derived from `upload/*`. As a workaround we just define base capability
 // here so all store capabilities could be derived from either `*` or
-// `store/*`.
+// `upload/*`.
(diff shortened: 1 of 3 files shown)


ice-breaker-tg (approved):
LGTM

We should also look into extending the array (and maybe some other validators) to have minimum length, etc?

Something like, in the todo app example, could be like "item name" as a caveat, and needs to be 
string.minimumLength(4);

Somewhat unrelated, but just should maybe think about it

Gozala (approved):

Gozala:
> @Gozala should root be the root CID of the data, not a CAR CID?

Correct! Thanks for catching and fixing

Gozala:
> Also, for a separate PR, maybe neither of these fields should be optional. Like, an "upload" should always have a data root CID and be present in at least 1 shard?

making root required sounds good! It was only optional because of some bug in ucanto which has been fixed.


As of `shards` I’m hesitant it’s really just a hint and there’s no guarantee about the accuracy. I would say let’s keep it optional at least until we have some validation in place that would reject operation when root isn’t in the shard.


As a side it may be a good idea to have separate field for the root shard so we won’t have to check every single one

PR #139, opened 2022-11-07
description of beta migration

jchris:


packages/access-api/src/utils/html.js
```diff
@@ -86,6 +86,13 @@ export const ValidateEmail = ({ delegation, ucan, qrcode }) => (
       {delegation.capabilities[0].nb.identity.replace('mailto:', '')} was
       confirmed. You may close this window.
     </p>
+    <p>
+      During the beta period, uploads via w3up will only appear via the beta 
+      APIs, and not on the website. Eventually, existing uploads from the old 
+      APIs will be migrated to this API. Please use a unique email address for 
+      beta accounts that you would like to link to NFT.Storage after the beta
```
hugomrdias, line 93: shouldn't this be web3.storage instead of nft.storage ?

hugomrdias (changes requested):

packages/access-api/src/utils/html.js
jchris: this is for NFT.Storage, so people who are looking for free NFT storage know what to do. Maybe the wording needs to be more clear.

hugomrdias: but isn't this true for web3.storage also ?

jchris: basically we will treat all account as web3 storage accounts unless the user tell us they are nft.storage accounts. @dchoi27 mind looking at the verbiage?

jchris:
During the beta period, uploads via w3up will only appear via the beta 
       APIs, and not on the website. Eventually, existing uploads from the old 
       APIs will be migrated to this API. Please use a unique email address for 
       beta accounts that you would like to link to NFT.Storage after the beta period.

packages/access-api/src/utils/html.js
dchoi27: It might be more confusing to the user if we tell them existing web3.storage uploads will be migrated to w3up, rather than w3up uploads getting migrated to web3.storage (from their perspective, it should be one and the same). Let's go with something like:

During the beta period, uploads via w3up will only appear via the beta APIs, and not on the web3.storage or NFT.Storage websites, even if the associated email addresses are the same.

At the end of the beta period, web3.storage and NFT.Storage's upload APIs and existing accounts will be upgraded to use w3up. Uploads via the w3up beta be migrated to web3.storage by default and, if applicable, combined with the relevant existing web3.storage account (based on email address). If you would rather have your beta w3up account be associated with the NFT.Storage product (e.g., you are only storing off-chain NFT data), please email us at `support@nft.storage`. All w3up uploads associated with a given registered email will only be able to be associated with one of either web3.storage or NFT.Storage when these uploads are migrated.

hugomrdias (approved):

hugomrdias:
Text LGTM , i will merge this after making sure ui looks good

PR #148, opened 2022-11-09
feat: add static uploads API

alanshaw:
This PR adds static methods allowing streaming uploads to the service. It is integrated with ucanto 0.9.

Essentially the client is "bring your own agent" - it requires an `issuer` (a `Signer`) and `proofs` (delegated capabilities) to be passed to each method.

The API is oriented around 3 use cases:

1. Simple and easy to use `uploadFile`/`uploadDirectory` methods that return CIDs.
2. More advanced APIs that split out DAG creation, CAR creation, storing and registering uploads.
3. Streaming APIs that have the same split, but cater for uploads of arbitrary size using streams.

Every method has tests and the [README has extensive examples and API reference documentation](https://github.com/web3-storage/w3protocol/blob/777327609cd2c96b2be4e5e8d48e9a45f17b9a7b/packages/upload-client/README.md).

closes https://github.com/web3-storage/w3protocol/issues/150


hugomrdias:
does this closes https://github.com/web3-storage/w3protocol/issues/150 ?

hugomrdias:
you need to add 

```text
"references": [{ "path": "../access" }]
```
to tsconfig.json

packages/upload-client/src/storage.js
```diff
@@ -0,0 +1,134 @@
...
+          audience: serviceDID,
+          with: account,
+          nb: {
+            // @ts-expect-error should allow v0 CIDs!
+            root,
+            shards,
+          },
+        })
+        .execute(conn)
+      if (result?.error === true) throw result
```
vasco-santos, line 60:
```suggestion
      if (result?.error === true) throw result.error
```

I guess this is what we want?

vasco-santos (commented):
This is amazing ❤️  Learned so much

olizilla (commented):
Partial review to get the feedback out. I will review the rest tomorrow.

packages/upload-client/package.json
```diff
@@ -14,20 +14,41 @@
...
-    "build": "tsc --build",
-    "test": "npm run test:node && npm run test:browser",
-    "test:node": "mocha 'test/**/!(*.browser).test.js' -n experimental-vm-modules -n no-warnings",
+    "lint": "eslint '**/*.{js,ts}' && prettier --check '**/*.{js,ts,yml,json}' --ignore-path ../../.gitignore",
+    "build": "run-s build:*",
+    "build:deps": "pnpm -r --filter @web3-storage/access run build",
+    "build:tsc": "tsc --build",
+    "test": "npm-run-all -p -r mock:bucket test:all",
+    "test:all": "run-s test:browser test:node",
+    "test:node": "c8 -r html -r text mocha 'test/**/!(*.browser).test.js' -n experimental-vm-modules -n no-warnings",
```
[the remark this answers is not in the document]
alanshaw: Yeah we can do, I still want the report though.

packages/upload-client/README.md
```diff
@@ -0,0 +1,408 @@
...
+### Step 0
+
+Obtain the invocation configuration. i.e. the issuer (the signing authority) and proofs that the issuer has been delegated the capabilities to store data and register uploads:
+
+```js
+import { Agent } from '@web3-storage/access-client'
+import { add as storeAdd } from '@web3-storage/access-client/capabilities/store'
+import { add as uploadAdd } from '@web3-storage/access-client/capabilities/upload'
+
+const agent = await Agent.create({ store })
```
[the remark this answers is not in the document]
alanshaw: @hugomrdias can the agent create an appropriate store so this isn't mandatory?

packages/upload-client/README.md
```diff
@@ -0,0 +1,408 @@
...
+class ShardingStream extends TransformStream<Block, CARFile>
+```
+
+Shard a set of blocks into a set of CAR files. The last block written to the stream is assumed to be the DAG root and becomes the CAR root CID for the last CAR output.
+
+Note: `CARFile` is just a `Blob` with two extra properties:
+
+```ts
+type CARFile = Blob & { version: 1; roots: CID[] }
+```
```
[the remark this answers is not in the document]
alanshaw: Yeah there's a few other repeated bits of information as well so we should do the heading thing. I don't want to assume folks have read the rest of the document.

packages/upload-client/src/store.js
```diff
@@ -0,0 +1,173 @@
...
+  const conn = options.connection ?? connection
+  const result = await retry(
+    async () => {
+      return await StoreCapabilities.add
+        .invoke({
+          issuer,
+          audience: serviceDID,
+          // @ts-expect-error expects did:${string} but cap with is ${string}:${string}
+          with: capability.with,
+          nb: { link },
```
vasco-santos, line 43: size is now required per https://github.com/web3-storage/w3protocol/releases/tag/access-v5.0.0

packages/upload-client/test/upload.test.js
```diff
@@ -0,0 +1,266 @@
...
+    const service = mockService({
+      upload: {
+        remove(invocation) {
+          assert.equal(invocation.issuer.did(), issuer.did())
+          assert.equal(invocation.capabilities.length, 1)
+          const invCap = invocation.capabilities[0]
+          assert.equal(invCap.can, UploadCapabilities.remove.can)
+          assert.equal(invCap.with, account.did())
+          assert.equal(String(invCap.nb.root), car.roots[0].toString())
+          return null
```
olizilla, line 215: same here, i think we need plan or invocation count assertion for this test to verify that remove was called correctly

packages/upload-client/test/utils.test.js
```diff
@@ -0,0 +1,59 @@
...
+    const proofs = [
+      await StoreCapabilities.add.delegate({
+        issuer,
+        audience: id,
+        with: issuer.did(),
+        expiration: Infinity,
+      }),
+    ]
+
+    assert.throws(() => findCapability(proofs, 'store/add', issuer.did()), {
```
olizilla, line 25:
```suggestion
    // we match on `audience`. Passing in the issuer or any other DID here should fail.
    assert.throws(() => findCapability(proofs, 'store/add', issuer.did()), {
```

packages/upload-client/test/index.test.js
```diff
@@ -1,8 +1,162 @@
...
+describe('uploadFile', () => {
+  it('uploads a file to the service', async () => {
+    const res = {
+      status: 'upload',
+      headers: { 'x-test': 'true' },
+      url: 'http://localhost:9200',
+    }
+
+    const account = await Signer.generate()
+    const issuer = await Signer.generate()
```
olizilla, line 24: I found the test difficult to understand initially... a comment here would help
```suggestion
    const issuer = await Signer.generate() // The "user" that will ask the service to accept the upload
```

we can circle back and rename account -> space later.

olizilla (commented):
Lovely tests.

packages/upload-client/src/utils.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @param {string} can
+ * @param {import('@ucanto/interface').Ability} ability
+ */
+function capabilityMatches(can, ability) {
+  return can === ability
+    ? true
+    : can.endsWith('*') && ability.startsWith(can.split('*')[0])
+}
```
olizilla, line 36:
```suggestion
function capabilityMatches(can, ability) {
  if (can === ability) return true
  if (can === '*/*') return true
  if (can === '*') return true
  if (can.endsWith('*') && ability.startsWith(can.slice(0, -1)) return true
  return false
}
```

- top is `*/*` https://github.com/ucan-wg/spec/blob/main/README.md#52-top
- or is it `*` https://github.com/ucan-wg/spec/blob/main/README.md#23-ability
- `*` could appear anywhere in `can` value. it would be whack, but maybe explicitly chop off the last * when checking for a wildcard

olizilla (commented):
can can

packages/upload-client/src/index.js
```diff
@@ -1,4 +1,99 @@
...
+    UnixFS.createDirectoryEncoderStream(files),
+    options
+  )
+}
+
+/**
+ * @param {import('./types').InvocationConfig} invocationConfig
+ * @param {ReadableStream<import('@ipld/unixfs').Block>} blocks
+ * @param {UploadOptions} [options]
+ * @returns {Promise<import('multiformats').Link<unknown, number, number, import('multiformats').Version>>}
```
olizilla, line 74: plz my eyes. can has @typedef for `import('multiformats').Link<unknown, number, number, import('multiformats').Version>`

packages/upload-client/src/sharding.js
```diff
@@ -0,0 +1,112 @@
+import Queue from 'p-queue'
+import { encode } from './car.js'
+import { add } from './store.js'
+
+// most thing are < 30MB
```
olizilla, line 5: sure

olizilla: > in general, when your object size reaches 100 MB, you should consider using multipart uploads instead of uploading the object in a single operation.
– https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html

packages/upload-client/src/unixfs.js
```diff
@@ -0,0 +1,144 @@
+import * as UnixFS from '@ipld/unixfs'
+import * as raw from 'multiformats/codecs/raw'
+
+const queuingStrategy = UnixFS.withCapacity(1_048_576 * 175)
```
olizilla, line 4: Can has comment on why these numbers?

olizilla (approved):
Beautiful work ✨🎷🐩

packages/upload-client/src/sharding.js
alanshaw: Ok I update to 100

packages/upload-client/test/index.test.js
```diff
@@ -1,8 +1,182 @@
...
+        with: account.did(),
+        expiration: Infinity,
+      }),
+      UploadCapabilities.add.delegate({
+        issuer: account,
+        audience: serviceSigner,
+        with: account.did(),
+        expiration: Infinity,
+      }),
+    ])
```
olizilla, line 42: How does this work? what ties issuer + account together? why is `issuer` allowed to invoke an upload with this proof chain?

Gozala: How does this work? what ties issuer + account together?

I'm not sure I understand what you mean, here also what test does here is hopefully not what we do in the actual client. Idea is that `space` should delegate all of it's capabilities, that is `{ can: *, with: space.did() }` to the service DID so that service in the future could delegate it back to the user.

Problem here is that it delegates just specific capabilities not all of them. That would imply that if user looses key we can delegate back only these two capabilities but nothing else.

> why is `issuer` allowed to invoke an upload with this proof chain?

This is precisely what the [ownership section](https://hackmd.io/Zb7gjpLsQn2w3a3JUvnFcw?view#Ownership) here tries to address.

If `with` DID is the same as `iss` DID you don't need any proof because it's self-evident that you own that resource, otherwise how would you would not be able to issue such UCAN because you would not be able to sign without a private key.

P.S.: If you think there is a good / better place way to capture that let me know

Gozala: If you mean how does space tied to the agent then

1. On space creation:
   a. space delegates `{ can: *, with: space.did() }` to `agent.did()` - So agent can act on user behalf
   b. space delegates `{ can: *, with: space.did() }` to `service.did()` - So service can delegate to users other agent after email verification. Key recovery & rotation is the same flow.
   c. space should delegate to user email as per #165 
2. On login to existing space
   a. User types email address and agent asks service for the delegation to (local) `agent.did()`
   b. Service delegates capabilities from 1. .b to requesting `agent.did()` and sends an email with a link to import it. After clicking it agent imports delegation and now has access to the space.

As an aside we would be able to replace delegation form space -> service, with space to user email address so we never actually have user space capabilities, because with [ucan-mailto](https://github.com/ucan-wg/ucan-mailto/pull/1/files) will be able to do email to did delegations

olizilla: my question was really "why is the issuer allowed to invoke `store/add` on this space when the proof chain the issuer provides doesn't include a delegation from space to issuer".

olizilla: this test passes and i assumed it would fail as it is currently written.

Gozala: > my question was really "why is the issuer allowed to invoke `store/add` on this space when the proof chain the issuer provides doesn't include a delegation from space to issuer".

Got it, does [this comment](https://github.com/web3-storage/w3protocol/pull/148#discussion_r1024296793) answers it / makes sense ?

olizilla: Very helpful, thanks! So if I understand correctly the self-evident proof is a very sensitive bunch o bytes. If it got intercepted any agent could use it... Any agent could upload to thay space if they got hold of such a proof. (Were it to be constructed in the wild)

alanshaw:
```js
StoreCapabilities.add.delegate({
  issuer: account,
  audience: serviceSigner,
  with: account.did(),
  expiration: Infinity,
}),
```

As I understood it when I wrote it, the space (`account` here) is delegating the `store/add` capability. The delegation is valid for use with our service (the audience) and with the passed space (the resource, or "with").

Anyone who recieves this delegation can issue `store/add` invocations for the space, at our service using this delegation as the proof.

So there's 2 questions:

1. Is that the right interpretation of what is being delegated and what is allows?
2. I changed `audience` in this delegation from the service to the agent and were able to successfully call the service with this delegation as proof. Why does that work? I expected ucanto server to reject. Is it a bug?

Gozala: > As I understood it when I wrote it, the space (`account` here) is delegating the `store/add` capability. The delegation is valid for use with our service (the audience) and with the passed space (the resource, or "with").
>
> Anyone who recieves this delegation can issue store/add invocations for the space, at our service using this delegation as the proof.
🎯

> Is that the right interpretation of what is being delegated and what is allows?

Yes

> I changed audience in this delegation from the service to the agent and were able to successfully call the service with this delegation as proof. Why does that work? I expected ucanto server to reject. Is it a bug?

It is correct because when you invoked capability you end up with following delegation chain

```
account -> agent -> service
```

account can self issue because it own's DID, agent can invoke because it has delegation from the account

Gozala: > Very helpful, thanks! So if I understand correctly the self-evident proof is a very sensitive bunch o bytes. If it got intercepted any agent could use it... Any agent could upload to thay space if they got hold of such a proof. (Were it to be constructed in the wild)

No it's not, self-evident delegation from owner is always to some DID and signed by the issuer (which is owner). No one other than DID owner delegated to could invoke that capability, because they would not be able to sign with the private key of that DID. If some other DID uses that delegation it will no longer be a valid chain it will be mole like

```
account -> agent -x- attacker -> service
```

olizilla: i'm still not getting it! In this test we do this:

```js
     const account = await Signer.generate()
     const issuer = await Signer.generate() // The "user" that will ask the service to accept the upload

     const proofs = await Promise.all([
       StoreCapabilities.add.delegate({
         issuer: account,
         audience: serviceSigner,
         with: account.did(),
         expiration: Infinity,
       }),
       UploadCapabilities.add.delegate({
         issuer: account,
         audience: serviceSigner,
         with: account.did(),
         expiration: Infinity,
       }),
     ])

    // ... trim server and conn creation ...

    const dataCID = await uploadFile({ issuer, proofs }, file, { connection } )
```

to my eyes that looks like we set up a proof that says 
- _"dear service, account here. allow store/add, upload/add on account, signed off by yours truly, account"_  ...with no mention of `issuer` the signer created on line 2 in the above snippet.

then it's invoked like
- _"hey service! issuer here! i'd like to store/add this car plz! here is a proof that says that account is cool with folks doing store/add and upload/add on account, so i guess store it in account. They didn't mention me explicitly, but i found this string of bytes accidentally sent to a log file, so i pasted it in here, but we're cool, right? it's all signed off by account! Also, I, issuer, am also obviously me (not account), as i have signed this off as `issuer`."_

olizilla: that maybe just how it works, i'm just trying to clarify my concern around such a self evident proof looking like maybe it is dangerously powerful. Why is a proof chain that never addresses our signer called `issuer` as the `audience` for the capbailities usable by `issuer` to prove they can do a thing?

Gozala: > i'm still not getting it! In this test we do this:
> 
> ```js
>      const account = await Signer.generate()
>      const issuer = await Signer.generate() // The "user" that will ask the service to accept the upload
> 
>      const proofs = await Promise.all([
>        StoreCapabilities.add.delegate({
>          issuer: account,
>          audience: serviceSigner,
>          with: account.did(),
>          expiration: Infinity,
>        }),
>        UploadCapabilities.add.delegate({
>          issuer: account,
>          audience: serviceSigner,
>          with: account.did(),
>          expiration: Infinity,
>        }),
>      ])
> 
>     // ... trim server and conn creation ...
> 
>     const dataCID = await uploadFile({ issuer, proofs }, file, { connection } )
> ```
> 
> to my eyes that looks like we set up a proof that says

I don't know what `uploadFile` does there but assuming it issues some invocation using `proofs` as proofs it should not work. Issuer there should be an `account` or delegations should change to delegate to issuer instead of `serviceSigner`

Gozala: Again it kind of depends what `uploadFile` does it may be using `issuer.did()` in `with` field in which case proofs are pointless and validator won't look at them because `issuer` could issue any capability as long as `with` is `issuer.did()` without proofs

Gozala: > that maybe just how it works, i'm just trying to clarify my concern around such a self evident proof looking like maybe it is dangerously powerful. Why is a proof chain that never addresses our signer called issuer as the audience for the capbailities usable by issuer to prove they can do a thing?

No if what happens under the hood looks roughly like the following, validator would have to say you're not authorized and if it does not I'm very surprised that such bug has escaped:


```ts
{ 
   iss: 'did:key:iss'
   aud: 'did:key:service'
   att: [
       { can: 'store/add', with: 'did:key:acc', link: 'bag...car' },
   ],
  proofs: [
      { iss: 'did:key:acc'
        aud: 'did:key:service'
        att: [{ can: "store/add", with: "did:key:acc" }]
      }
  ]
}
```

Gozala: I see now that what @olizilla was describing is actually what's in this test, I'm surprised it's passing. I have created test in ucanto to check for principal aligment https://github.com/web3-storage/ucanto/pull/134

I'll continue investigating why this isn't failing

PR #149, opened 2022-11-09
feat: account recover with email

hugomrdias:
Theres also a commit with changes to rename access to access-client


packages/access-api/src/service/index.js
```diff
@@ -18,13 +22,50 @@ export function service(ctx) {
     },
 
     account: {
-      info: Server.provide(Account.info, async ({ capability }) => {
+      info: Server.provide(Account.info, async ({ capability, invocation }) => {
         const results = await ctx.kvs.accounts.get(capability.with)
         if (!results) {
-          throw new Failure('Account not found...')
+          return new Failure('Account not found.')
```
alanshaw, line 28: It feels counter-intuitive to return an error...why can we not throw these?

packages/access-api/src/service/index.js
```diff
@@ -46,21 +87,29 @@ export function service(ctx) {
...
             })
             .delegate()
 
           const encoded = await delegationToString(inv)
+          const url = `${ctx.url.protocol}//${ctx.url.host}/validate-email?ucan=${encoded}&mode=recover`
+
           // For testing
           if (ctx.config.ENV === 'test') {
-            return encoded
+            return url
```
alanshaw, line 112: Only return in test?

alanshaw (approved):

packages/access-api/src/service/index.js
hugomrdias: These are known errors that should be returned to the user but not sent to sentry.
Ucanto supports multiple invocations in the same request thats why the responses are always arrays, so some invocations can succeed and some can error thats why we dont throw here.

In ucanto a known error like "account not found" is a valid response, the server types also include those. Any other exception error like "fail to parse json" that we want to send to sentry we should throw from the code.

packages/access-api/src/service/index.js
hugomrdias: yes just for tests so we can process the delegation without dealing with emails.

alanshaw (approved):

PR #156, opened 2022-11-10
feat: add upload-client class

alanshaw:


alanshaw:
I will submit a new PR if/when we need this.

PR #159, opened 2022-11-11
chore: add doc comments to the capabilities

Gozala:
## overview

- Adds bunch of comments to the capabilities
- Makes mandatory fields non-optional (I had to fix bugs in ucanto and update to new version to fix this)

PR #163, opened 2022-11-12
feat!: Update voucher API

Gozala:
## Overview

Originally [voucher spec](https://hackmd.io/@gozala/w3-accounts) used:

1. `nb.of` (named `nb.product` here) to identify a product.
2. `nb.at` (named `nb.service` here) to identify a provider service of the product.
3. `with` field in `voucher/redeem` was basically web3.storage did so it could self-issued.

With this change we simplify things by:

1. Identifying products via did:key
2. Requiring that `vouche/redeem` used product DID in the `with` field, which in turn implies product owner should delegate capability to the validator.

### Background

I realized that we could simply identify products via did:key, which in turns remove need for both `nb.of` and `nb.at` fields. Not only it simplifies things it also means that any one could create a product (not just us but also our users)  and likewise anyone could install any product as long as product owner delegated that capability by whatever means.
- Of course we may not do anything if some account installs some product we're not aware of, but that is ok it's like `upload/list` we just need to save products in the list product list and only change account settings based on products we're aware of.
- In the future we may also create ability to publishing products, e.g PL could create "pl-tier" product which basically would provide accounts with some space while billing "pl-tier" owner.
    

⚠️ Please note that this branch targets #159 because it fixed bunch of problems which I did not want to deal once again. Once #159 lands I'll change target to main branch


Gozala:
I'm also starting to think that maybe we don't need `voucher/redeem` but instead it should be something like `product/install`

packages/access-api/src/service/voucher-claim.js
```diff
@@ -6,29 +6,43 @@ import { delegationToString } from '@web3-storage/access/encoding'
  * @param {import('../bindings').RouteContext} ctx
  */
 export function voucherClaimProvider(ctx) {
+  // Provider should have access to delegated vouchers which it can
+  // redelegate. Currently however we just identify "free-tier" with a
+  // provider DID.
+  const products = new Map([
+    [
+      ctx.signer.did(),
```
alanshaw, line 14: Should we create a separate DID for this? This is the DID of the service no?

alanshaw (changes requested):

packages/access-api/src/service/voucher-claim.js
Gozala: In practice I think we should have DID per product, that said I was not sure what the best way to thread that through, so I thought this was reasonable compromise to get at least discussion started.

Gozala:
After thinking about treading through delegations from voucher -> service I've realized there are bunch of complications:

1. If we rotate service key we need to coordinate all voucher delegations
2. For tests we we'll need to somehow have another set of delegations (because DID in test is different)

Since https://github.com/web3-storage/w3protocol/issues/182 will basically address all the above, I'm going to pause it for now and resume it afterwards

PR #166, opened 2022-11-15
fix: store add capabilities

vasco-santos:
`store/add` invocation should have at bare minimum link required.

However, perhaps we should make size required given we want to use it for the signed url?

BREAKING CHANGE: store/add capabilities now have required properties


alanshaw (approved):
LGTM

Gozala:
This was fixed in https://github.com/web3-storage/w3protocol/pull/159 and should not longer be required

PR #169, opened 2022-11-16
fix: workaround ts bug & generate valid typedefs

Gozala:
This works arounds typescript bug https://github.com/microsoft/TypeScript/issues/51548 which in turn caused https://github.com/web3-storage/ucanto/issues/133 by exporting imported modules so that generated typedefs will have those references

### DIFF
--- a/packages/access/package.json
+++ b/packages/access/package.json
@@ -115,7 +115,8 @@
       }
     },
     "rules": {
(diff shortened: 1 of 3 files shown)


vasco-santos (approved):
LGTM! Thanks @gozala :heart:

PR #171, opened 2022-11-16
fix: workaround for ts bug in upload capabilities

vasco-santos:
Same as https://github.com/web3-storage/w3protocol/pull/169 but for upload capabilities. Did also a quick look at generated types in other capabilities and they look fine.

### DIFF
--- a/packages/access/src/capabilities/upload.js
+++ b/packages/access/src/capabilities/upload.js
@@ -137,3 +137,7 @@ export const list = base.derive({
    */
   derives: equalWith,
 })
+
+// ⚠️ We export imports here so they are not omited in generated typedes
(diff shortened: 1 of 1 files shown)


packages/access/src/capabilities/upload.js
```diff
@@ -137,3 +137,7 @@ export const list = base.derive({
    */
   derives: equalWith,
 })
+
+// ⚠️ We export imports here so they are not omited in generated typedes
```
alanshaw, line 141:
```suggestion
// ⚠️ We export imports here so they are not omited in generated typedefs
```

alanshaw (approved):

Issue #153, opened 2022-11-10
access-client supports `currentAccount` and `getProof` for upload-client


hugomrdias:
- [ ] setCurrentAccount
- [ ] getCurrentAccount
- [ ] getProof(cap, accountDID?)

PR #177, opened 2022-11-16
feat: #153

hugomrdias:
- typedoc support to generate docs for all the clients
- add list of Abilities types
- renamed `any.js` to `wildcard.js`
- added `parseAbility` and `canDelegateAbility` functions to the capabilities `utils.js`
- cli is mostly commented out needs to be updated in a follow PR
- StoreData interface revamped
- added more exports to index.js to help with the generated docs until typedoc supports export maps
- added a bunch of delegation utilities to `delegation.js`
- agent has new methods, `addProof`, `proofs`, `delegations`, `delegationsWithMeta`, `createSpace`, `setCurrentSpace`, `currentSpace`, `currentSpaceWithMeta` and `execute`
- old createAccount renamed to `registerSpace`
- a bunch of jsdocs 

preview docs https://vigorous-chandrasekhar-6b427f.netlify.app/

todo:

- [x] rename caps
- [x] rename account to space in api
- [x] finish setup wrangler migrations for staging and prod (using CI)
- [x] more docs for agent methods 
- [x] feedback
- [x] support multiple invocations to be executed on a connection


packages/access-client/src/agent.js
```diff
@@ -128,67 +126,240 @@ export class Agent {
...
+    }
+
+    await this.store.save(this.data)
+  }
+
+  /**
+   * @param {import('@ucanto/interface').Capability[]} [caps]
+   */
+  async *proofs(caps) {
+    for await (const value of this.proofsWithMeta(caps)) {
```
alanshaw, line 180: Can they please be collected for me? I'm not sure when I would not just collect them all before sending them...

packages/access-client/src/cli/cmd-whoami.js
```diff
@@ -1,33 +1,31 @@
...
+  //   console.log('Delegations received:')
+  //   for (const [key, value] of delegations.receivedByResource) {
+  //     console.log(
+  //       `Resource: ${key}`,
+  //       value.map((cap) => cap.cap.can)
+  //     )
+  //   }
+  // } else {
+  //   console.error(`Run "${NAME} setup" first`)
+  // }
```
alanshaw, line 30: Needs implementation?

packages/access-client/src/cli/index.js
```diff
@@ -81,111 +39,103 @@ prog
...
+    //   try {
+    //     const result = await agent.getAccountInfo(account)
+    //     console.log(result)
+    //   } catch (error_) {
+    //     const error = /** @type {Error} */ (error_)
+    //     console.log(error.message)
+    //   }
+    // } else {
+    //   console.error(`Run "${NAME} setup" first`)
+    // }
```
alanshaw, line 75: Needs implementation?

packages/access-client/src/stores/store-indexeddb.js
```diff
@@ -41,6 +40,10 @@ export class StoreIndexedDB {
     this.#dbStoreName = options.dbStoreName ?? STORE_NAME
   }
 
+  /**
+   *
+   * @returns {Promise<Store>}
```
alanshaw, line 45: If you declare that these instance methods return a `Store` then you lose the ability to use it in specifics. It should adhere to that interface _anyway_, you shouldn't need to explicitly state it.

There's a bunch of these that have been added - do we really need to?

packages/access-client/test/awake.node.test.js
```diff
@@ -41,12 +40,12 @@ describe('awake', function () {
   it('should send msgs', async function () {
     const agent1 = await Agent.create({
       store: await StoreMemory.create(),
-      fetch: globalThis.fetch || fetch,
       url: new URL('http://127.0.0.1:8787'),
     })
+    const account = await agent1.createSpace('responder')
```
alanshaw, line 45: Still referring to "account" here?

packages/access-client/src/cli/cmd-whoami.js
hugomrdias: im gonna rework the cli in a follow up

packages/access-client/src/cli/index.js
hugomrdias: im gonna rework the cli in a follow up

packages/access-client/src/stores/store-indexeddb.js
hugomrdias: This `Store` type is the local one defined at the top of the file, its already specific. If you dont do this you loose the method documentation defined in the types.ts file for the interface.

What are we losing here ?

alanshaw: Oh ok I thought this was referring to the interface!

alanshaw (approved):

PR #181, opened 2022-11-17
feat: explicit resource

alanshaw:
This PR:

1. Makes `with` explicit in the `InvocationConfig` (no longer derived from proofs).
2. Adds the ucanto validator to the test servers and fixes the delegations (which were incorrect, but working in the tests because no validator 🙈).
3. Makes `audience` optional in the `InvocationConfig` and uses it for testing.
4. Renamed `serviceDID` to `servicePrincipal` because it's an object with a `did()` method, not a DID itself.

A lot of this is very mechanical renames, so should be relatively easy to review.


packages/upload-client/README.md
```diff
@@ -15,27 +15,33 @@ npm install @web3-storage/upload-client
...
+import { upload } from '@web3-storage/access/capabilities/upload'
 
-const agent = await Agent.create({ store })
+const agent = await Agent.create()
 
-// Note: you need to create and register an account 1st time:
-// await agent.createAccount('you@youremail.com')
+// Note: you need to create and register a space 1st time:
+// await agent.createSpace()
+// await agent.registerSpace('you@youremail.com')
```
alanshaw, line 33: This is TBC...skating to where the puck will be based on https://github.com/web3-storage/w3protocol/pull/177

vasco-santos (approved):
LGTM
Things are more clear now around spaces and delegations 🎉

PR #184, opened 2022-11-18
feat: add support for list pagination in list capability invocations

vasco-santos:
Adds support for list pagination in `store/list` and `upload/list`.

Relevant notes on decisions made:
- unfortunately, Databases like DynamoDB does not support pagination by just giving `pageNumber` and `size`. Aiming to have something generic (that works in DynamoDB, but also other DBs), chosen approach is to rely on `cursor` and `size` .
- at first, attempted to make `cursor` a Link property of the capability. And it worked well for `store/list`, but then I got to use cases (`upload/list` for instance) where it cannot be a link. This is because Upload Table primary key contains `${dataCid}#${carCid}` to be unique. So, in this case that is the cursor we need to have. There might be other kind of use cases, so having this as string is really the most flexible approach to support pagination.

Note:
- it would be great to get a fix for https://github.com/web3-storage/ucanto/issues/140 before merging this

BREAKING CHANGE: store/list and upload/list types now require nb object with optional properties

### DIFF
--- a/packages/access-client/src/capabilities/store.js
+++ b/packages/access-client/src/capabilities/store.js
@@ -140,6 +140,18 @@ export const list = base.derive({
      * be stored.
      */
     with: URI.match({ protocol: 'did:' }),
+    nb: {
+      /**
+       * Item where a previous list operation stopped, inclusive of the previous
+       * result set. Use this value to start a new operation, inorder to paginate
+       * list.
+       */
+      cursor: Schema.string().optional(),
+      /**
+       * Size of the page being requested to list.
+       */
+      size: Schema.integer().optional(),
+    },
     derives: (claimed, delegated) => {
       if (claimed.with !== delegated.with) {
         return new Failure(
--- a/packages/access-client/src/capabilities/store.js
+++ b/packages/access-client/src/capabilities/store.js
@@ -143,12 +143,12 @@ export const list = base.derive({
     nb: {
       /**
        * Item where a previous list operation stopped, inclusive of the previous
-       * result set. Use this value to start a new operation, inorder to paginate
+       * result set. Use this value to start a new operation, in order to paginate
        * list.
        */
       cursor: Schema.string().optional(),
       /**
-       * Size of the page being requested to list.
+       * Maximum number of items per page.
        */
       size: Schema.integer().optional(),
     },
--- a/packages/access-client/src/capabilities/store.js
+++ b/packages/access-client/src/capabilities/store.js
@@ -142,9 +142,8 @@ export const list = base.derive({
     with: URI.match({ protocol: 'did:' }),
     nb: {
       /**
-       * Item where a previous list operation stopped, inclusive of the previous
-       * result set. Use this value to start a new operation, in order to paginate
-       * list.
+       * A pointer that can be moved back and forth on the list.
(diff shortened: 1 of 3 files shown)


packages/access-client/test/capabilities/upload.test.js
```diff
@@ -328,6 +328,7 @@ describe('upload capabilities', function () {
       audience: w3,
       with: account.did(),
       proofs: [await any],
+      nb: {},
```
vasco-santos, line 331: Types require empty object when all nb properties are optional. need to look

packages/access-client/src/capabilities/upload.js
```diff
@@ -130,6 +130,18 @@ export const list = base.derive({
   to: capability({
     can: 'upload/list',
     with: URI.match({ protocol: 'did:' }),
+    nb: {
+      /**
+       * Item where a previous list operation stopped, inclusive of the previous
+       * result set. Use this value to start a new operation, inorder to paginate
```
alanshaw, line 136:
```suggestion
       * result set. Use this value to start a new operation, in order to paginate
```

packages/access-client/src/capabilities/store.js
```diff
@@ -130,6 +130,18 @@ export const list = base.derive({
      * be stored.
      */
     with: URI.match({ protocol: 'did:' }),
+    nb: {
+      /**
+       * Item where a previous list operation stopped, inclusive of the previous
+       * result set. Use this value to start a new operation, inorder to paginate
```
alanshaw, line 136:
```suggestion
       * result set. Use this value to start a new operation, in order to paginate
```

packages/access-client/src/capabilities/store.js
```diff
@@ -130,6 +130,18 @@ export const list = base.derive({
...
+       * list.
+       */
+      cursor: Schema.string().optional(),
+      /**
+       * Size of the page being requested to list.
```
alanshaw, line 141:
```suggestion
       * Maximum number of items per page.
```

packages/access-client/src/capabilities/upload.js
```diff
@@ -130,6 +130,18 @@ export const list = base.derive({
...
+       * list.
+       */
+      cursor: Schema.string().optional(),
+      /**
+       * Size of the page being requested to list.
```
alanshaw, line 141:
```suggestion
       * Maximum number of items per page.
```

packages/access-client/src/capabilities/store.js
```diff
@@ -130,6 +130,18 @@ export const list = base.derive({
      * be stored.
      */
     with: URI.match({ protocol: 'did:' }),
+    nb: {
+      /**
+       * Item where a previous list operation stopped, inclusive of the previous
```
alanshaw, line 135: This sounds specific to the particular implementation backing this. Maybe this could be more generic?

alanshaw (approved):

packages/access-client/test/capabilities/upload.test.js
hugomrdias: this should not happen theres a lot of infer types in ucanto to mitigate this. Did you open an issue in ucanto ?

PR #185, opened 2022-11-18
feat: allow custom shard size

alanshaw:
Bubbles up the shard size option so `uploadFile` and `uploadDirectory` can set it.

### DIFF
--- a/packages/upload-client/test/index.test.js
+++ b/packages/upload-client/test/index.test.js
@@ -128,8 +128,8 @@ describe('uploadFile', () => {
     ])
 
     const service = mockService({
-      store: { add: () => res },
-      upload: { add: () => null },
+      store: { add: provide(StoreCapabilities.add, () => res) },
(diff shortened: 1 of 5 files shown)


packages/upload-client/test/index.test.js
```diff
@@ -93,6 +93,65 @@ describe('uploadFile', () => {
...
+        issuer: space,
+        audience: agent,
+        with: space.did(),
+        expiration: Infinity,
+      }),
+    ])
+
+    const service = mockService({
+      store: { add: () => res },
+      upload: { add: () => null },
```
vasco-santos, line 127: we should add the server provide to guarantee the validations happen

vasco-santos (commented):
needs to updated based on latest commit merged to guarantee ucanto server does its validations. Other than that looks good

vasco-santos (approved):
LGTM

PR #198, opened 2022-11-22
refactor!: rename callback

alanshaw:
I keep getting this wrong, to the extent that I even documented it this way in the README.

I think `on[noun][verb]` aligns better with DOM event callbacks e.g. `onMouseClick`.


vasco-santos (approved):
LGTM

PR #207, opened 2022-11-23
feat(access-client): cli and recover

hugomrdias:


packages/access-api/scripts/migrate.js
```diff
@@ -14,34 +15,34 @@ const sqliteFormat = {
...
-    }
-  } catch (error) {
-    const err = /** @type {Error} */ (error)
-    // eslint-disable-next-line no-console
-    console.error('D1 Error', {
-      message: err.message,
-      // @ts-ignore
-      cause: err.cause?.message,
-    })
+  const runnedMigrations = /** @type {number} */ (
```
alanshaw, line 28:
```suggestion
  const appliedMigrations = /** @type {number} */ (
```

packages/access-client/src/agent.js
```diff
@@ -25,6 +26,23 @@ import {
 
 const HOST = 'https://access.web3.storage'
 
+/** @type {import('./types').Abilities[]} */
+export const abilitiesAsStrings = [
```
alanshaw, line 30: Aren't _abilities_ always strings?

I think this should be exported from `capabilities` export not here.

packages/access-client/src/cli/index.js
```diff
@@ -31,111 +31,104 @@ prog
...
+      const agent = await Agent.create({
+        store,
+        url,
+      })
+      const space = await selectSpace(agent)
+
+      await agent.setCurrentSpace(space)
+
+      const { audience, expiration, name, type, abilities } =
+        await inquirer.prompt([
```
alanshaw, line 79: Note: an interactive CLI means that folks can't use this for scripting...which I think will be an important use case initially since we won't have implementations in other languages.

packages/access-client/src/types.ts
```diff
@@ -80,8 +80,8 @@ export interface AgentData<T> {
   meta: AgentMeta
   principal: T
   currentSpace?: DID
-  spaces: Map<DID, SpaceMeta>
-  delegations: Map<CIDString, { meta?: DelegationMeta; delegation: Delegation }>
+  spaces: Map<DID | undefined, SpaceMeta>
```
alanshaw, line 83: No...?

alanshaw (changes requested):

packages/access-api/migrations/0002_add_delegation_column.sql
```diff
@@ -0,0 +1,3 @@
+-- Migration number: 0002 	 2022-11-29T14:41:37.991Z
+ALTER TABLE "spaces"
+ADD COLUMN "delegation" text DEFAULT NULL;
```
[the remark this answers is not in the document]
hugomrdias: yeah it was just easier to do it like this, can we change this in a follow PR when we move to providers and accounts ?

packages/access-api/src/service/index.js
```diff
@@ -36,31 +33,33 @@ export function service(ctx) {
...
+                audience: invocation.issuer,
+                issuer: ctx.signer,
+                with: delegation.capabilities[0].with,
+                expiration: Infinity,
+                proofs: [delegation],
+                facts: [metadata],
+              })
 
-            results.push(await delegationToString(del))
+              results.push(await delegationToString(del))
```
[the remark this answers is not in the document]
hugomrdias: created an issue for this https://github.com/web3-storage/w3protocol/issues/240

packages/access-client/src/cli/index.js
hugomrdias: https://github.com/web3-storage/w3protocol/issues/241

packages/access-client/src/types.ts
hugomrdias: just a workaround for TS, sometimes you want to `spaces.get(didThatMayBeUndefined)` and you can't because the type is too strict

package.json
```diff
@@ -42,5 +42,10 @@
     "depcheck": "^1.4.3",
     "typedoc": "^0.23.21",
     "typedoc-plugin-missing-exports": "^1.0.0"
+  },
+  "pnpm": {
+    "overrides": {
+      "@ucanto/principal": "3.0.1"
```
alanshaw, line 48: What does this do please?

packages/access-api/migrations/0002_add_delegation_column.sql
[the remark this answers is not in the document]
hugomrdias: https://github.com/web3-storage/w3protocol/issues/249

package.json
hugomrdias: bah was trying to workaround the ucanto release on friday i will remove it

packages/capabilities/src/capabilities/types.ts
```diff
@@ -29,19 +30,22 @@ export type StoreList = InferInvokedCapability<typeof list>
...
+  UploadAdd['can'],
+  UploadRemove['can'],
+  UploadList['can'],
+  Store['can'],
+  StoreAdd['can'],
+  StoreRemove['can'],
+  StoreList['can'],
+  VoucherClaim['can'],
+  VoucherRedeem['can']
+]
```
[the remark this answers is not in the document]
alanshaw: Ok gotcha!

PR #210, opened 2022-11-24
fix: do not load conf in browser

alanshaw:
Rollup unfortunately does not seem to be able to tree shake this module.


hugomrdias:
ah i see the issue there no export map for the agent and you need to use index.js, i would rather add an export map for the agent and remove the export map for index.js.
The stores can already be imported by `import { store } from @web3-storage/access/stores/store-indexeddb`

alanshaw:
https://github.com/web3-storage/w3protocol/pull/212

PR #216, opened 2022-11-28
fix: round trip delegations in IndexedDB store

alanshaw:
You can't store a CID instance in IndexedDB.

### DIFF
--- a/packages/access-client/src/stores/store-indexeddb.js
+++ b/packages/access-client/src/stores/store-indexeddb.js
@@ -1,6 +1,7 @@
 import { importDAG } from '@ucanto/core/delegation'
 import * as Signer from '@ucanto/principal/rsa'
 import defer from 'p-defer'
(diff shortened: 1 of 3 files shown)


packages/access-client/src/stores/store-indexeddb.js
```diff
@@ -149,25 +150,26 @@ export class StoreIndexedDB {
...
-              delegation: [...value.delegation.export()],
-            },
-          ])
-        }
-        const putReq = store.put({
+        /** @type {import('./types').StoreDataIDB} */
+        const raw = {
           id: DATA_ID,
           meta: data.meta,
+          // @ts-ignore
```
olizilla, line 157: expect so we get the nudge to remove this if the types align in the future?
```suggestion
          // @ts-expect-error
```

olizilla (approved):
👍

vasco-santos (approved):
lgtm

Issue #80, opened 2022-09-30
Move capabilities into own package PLZ


Gozala:
Hey @hugomrdias I would really like to decouple iterations on protocol from iterations on access SDK etc.. can we please just move all the capability definitions into own package e.g. `@web3-storage/protocol` or `@web3-storage/capabilities` and pull those in from wherever we need.

hugomrdias:
we can for sure, do we need to do it now though ? i would like to avoid having everyone using capabilities directly we should have an abstraction like the Agent or something else so we can make changes in the caps and backend without bringing the world

Gozala:
I think it makes sense to have a package with basically protocols, that is capabilities and some markdowns explaining the interaction.

Those could be the pulled in by the agent / client library that our tools and libraries can depend upon. Those also could be pulled in by various service implementations.

Current setup makes iteration on the protocol somewhat coupled with iteration on the implementation and I would like to be able to decouple them a bit more

gobengo:
I'll do this next week. Accidentally created a duplicate. Copying this description from that:

Acceptance Criteria:
* w3protocol monorepo has a new package containing just the capability objects used in the protocols
* clients like access-client import the capabilities from the capabilities package

Ideas
* Should we also split up the capabilities by [service](https://github.com/web3-storage/w3protocol/blob/main/spec/services.md): e.g. accounts, storage, upload ?

Gozala:
> Should we also split up the capabilities by [service](https://github.com/web3-storage/w3protocol/blob/main/spec/services.md): e.g. accounts, storage, upload ?

definitely split into separate modules, not sure separate packages are worth it, probably just going to only increase overhead right now

gobengo:
Reminder to me:

* [ ] ensure docs for `@web3-storage/capabilities` appear here https://web3-storage.github.io/w3protocol/modules.html
* [ ] ensure CI releases to npm
  * [ ] ensure `npm install @web3-storage/capabilities` works once released

gobengo:
@hugomrdias I'm assuming it's a good idea to release this to npm. I think that means I should add to this conditional to also run the release job for `packages/capabilities`? https://github.com/web3-storage/w3protocol/blob/191e2427529e113d67d5a6ec8ab7e7b4368b1c58/.github/workflows/release.yml#L40

gobengo:
After merge
* [ ] ensure CI releases to npm

After release:
[...]

PR #218, opened 2022-11-28
chore: split packages/access-client/src/capabilities to new package @web3-storage/capabilities

gobengo:
Motivation:
* https://github.com/web3-storage/w3protocol/issues/80

To do:
* [x] initial package split
* [x] remove capabilities dir from packages/access-client and fix imports in access-client
* [x] fix imports from other packages that were importing from access-client/capabilities
* [x] consider splitting to `capabilities/{service}/types` not `capabilities/types`
  * decided not to unless requested (makes exports map even more complicated, can do later)
* [x] GH actions should run tests on packages/capabilities
* [x]  .github/workflows/release.yml does npm/docs jobs if paths_released contains packages/capabilities


.github/workflows/capabilities.yml
```diff
@@ -0,0 +1,33 @@
...
+    paths:
+      - 'packages/capabilities/**'
+      - '.github/workflows/capabilities.yml'
+      - 'pnpm-lock.yaml'
+  pull_request:
+    paths:
+      - 'packages/capabilities/**'
+      - '.github/workflows/capabilities.yml'
+      - 'pnpm-lock.yaml'
+jobs:
```
Gozala, line 18: Can you please add the typecheck action, because it reports issues inline with code. I think other actions had them and they don't seem to appear anymore, not sure why they got removed, but I find them to be very helpful.

Here is an example of it
https://github.com/web3-storage/ucanto/blob/18b3fd17e8d7671dc76a238723cb8d524b29cba3/.github/workflows/transport.yml#L22-L48

P.S.: I know build will fail if there are type errors, but it just dumps them into output as opposed to printing them out in the code.

.github/workflows/capabilities.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/checkout@v3
+      - uses: pnpm/action-setup@v2.2.3
+        with:
+          version: 7
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'pnpm'
+      - run: pnpm install
+      - run: pnpm -r --filter @web3-storage/capabilities run build
```
Gozala, line 31: I don't believe we need to run the build, it mostly does typecheck which is better handled by separate job as suggested in the comment above.

packages/access-api/src/kvs/spaces.js
```diff
@@ -22,8 +22,8 @@ export class Spaces {
   }
 
   /**
-   * @param {import('@web3-storage/access/capabilities/types').VoucherRedeem} capability
-   * @param {Ucanto.Invocation<import('@web3-storage/access/capabilities/types').VoucherRedeem>} invocation
+   * @param {import('@web3-storage/capabilities/types').VoucherRedeem} capability
```
Gozala, line 25: Nit: Would be nice if we export all the types from the `@web3-storage/capabilities` directly so we won't have to keep adding `/types` everywhere.

packages/access-api/src/service/index.js
```diff
@@ -1,14 +1,14 @@
...
+import * as Space from '@web3-storage/capabilities/space'
 import { voucherClaimProvider } from './voucher-claim.js'
 import { voucherRedeemProvider } from './voucher-redeem.js'
 import * as DID from '@ipld/dag-ucan/did'
 import {
   delegationToString,
   stringToDelegation,
 } from '@web3-storage/access/encoding'
-import { top } from '@web3-storage/access/capabilities/top'
+import { top } from '@web3-storage/capabilities/top'
```
Gozala, line 11: Nit: I would reexport `top` from each capability group, so you could simply use `Space.top` if needed as opposed to needing to import second module everywhere.

packages/access-client/src/agent.js
```diff
@@ -9,9 +9,9 @@ import * as CBOR from '@ucanto/transport/cbor'
 import * as HTTP from '@ucanto/transport/http'
 import { URI } from '@ucanto/validator'
 import { Peer } from './awake/peer.js'
-import * as Space from './capabilities/space.js'
-import * as Voucher from './capabilities/voucher.js'
-import { top as Top } from './capabilities/top.js'
+import * as Space from '@web3-storage/capabilities/space'
+import * as Voucher from '@web3-storage/capabilities/voucher'
+import { top as Top } from '@web3-storage/capabilities/top'
```
Gozala, line 14:
```suggestion
import { Space, Voucher, Top } from '@web3-storage/capabilities'
```

packages/capabilities/package.json
```diff
@@ -0,0 +1,102 @@
+{
+  "name": "@web3-storage/capabilities",
+  "version": "0.0.0",
+  "description": "Capabilities for interacting with web3.storage",
```
Gozala, line 4: Not sure this is better, but it makes bit more sense to me

```suggestion
  "description": "Capabilities provided by web3.storage",
```

packages/capabilities/src/index.js
```diff
@@ -0,0 +1,5 @@
+export * as Space from './capabilities/space.js'
+export * as Top from './capabilities/top.js'
+export * as Store from './capabilities/store.js'
+export * as Upload from './capabilities/upload.js'
+export * as Voucher from './capabilities/voucher.js'
```
Gozala, line 5: If you add black `types.js` you'd be able to do following to export all the types along with it.

```suggestion
export * as Voucher from './capabilities/voucher.js'

export * from './capabilities/type.js'
```

Gozala (approved):
Most comments are suggestions, feel free to address or disregard. I don't think we need another review cycle even if you choose to address them.

.github/workflows/capabilities.yml
gobengo: good idea, did it

.github/workflows/capabilities.yml
gobengo: removed it when adding typecheck

packages/access-api/src/service/index.js
gobengo: did it

packages/capabilities/src/index.js
gobengo: If I add `types.js` next to `types.ts` it doesn't seem to work. we can try as followup but I think it might be trickier than that

packages/access-api/src/kvs/spaces.js
gobengo: I think it might not be supported in our current setup due to how we mix .js files and .ts file. can figure it out later but I think I tried it yesterday and settled for this to get it working.

hugomrdias (commented):
made a PR with fixes for my inline comments https://github.com/web3-storage/w3protocol/pull/239

.github/workflows/capabilities.yml
```diff
@@ -0,0 +1,44 @@
...
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        pnpm-version:
+          - 7
+        node-version:
+          - 18
+        package:
+          - capabilities
```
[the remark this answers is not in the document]
gobengo: it's what @Gozala had done in the code snippet he linked me to in his review :)

.github/workflows/capabilities.yml
```diff
@@ -0,0 +1,44 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: 'pnpm'
+      - run: pnpm install
+      - name: Typecheck
+        uses: gozala/typescript-error-reporter-action@v1.0.8
+        with:
+          project: packages/${{matrix.package}}/tsconfig.json
+      - run: pnpm -r --filter @web3-storage/${{matrix.package}} run lint
```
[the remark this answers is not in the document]
gobengo: ok I removed it based on [this](https://github.com/web3-storage/w3protocol/pull/218#discussion_r1036402150)

.github/workflows/upload-client.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           node-version: 18
           cache: 'pnpm'
       - run: pnpm install
-      - run: pnpm run build
+      - run: pnpm -r --filter @web3-storage/upload-client run build
```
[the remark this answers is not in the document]
hugomrdias: no all of it is in https://github.com/web3-storage/w3protocol/pull/239

PR #223, opened 2022-11-30
chore: add depcheck to all packages and update deps

hugomrdias:
I was concerned about missing deps but turns out we dont have any, bundlephobia cannot handle export maps so lots of ipld and ucanto stuff fail to be resolved because they dont have a `main` in package.json.

- update typescript to latest 0.9.4


alanshaw (approved):

vasco-santos (approved):
nice 🌟

PR #228, opened 2022-12-01
refactor!: access-client store decoupling

alanshaw:
Goals:
1. Decouple store from agent
2. Simplify agent creation
3. Agent governs data format not store
4. Initialization of agent data, not store
5. DRY initialization in agent, not repeated in each store impl

~~See: https://gist.github.com/alanshaw/ea4bd2b0ab215ade696eac1300be577d~~ outdated

```js
// In regular use:
const store = new StoreIndexedDB()
const data = await store.load()
const agent = data
  ? Agent.from(data, { store })
  : await Agent.create({}, { store })

// Then StoreMemory can be deleted, since the AgentData already stores all data in
// memory. It's equivalent to:
const agent = await Agent.create()
```

### DIFF
--- a/.github/release-please-config.json
+++ b/.github/release-please-config.json
@@ -4,7 +4,6 @@
     "packages/access-client": {},
     "packages/access-api": {},
     "packages/capabilities": {},
-    "packages/upload-client": {},
-    "packages/wallet": {}
+    "packages/upload-client": {}
   }
 }
--- a/.github/release-please-manifest.json
+++ b/.github/release-please-manifest.json
@@ -1,5 +1,4 @@
 {
-  "packages/wallet": "1.0.0",
   "packages/access-client": "7.0.2",
   "packages/access-api": "3.0.0",
   "packages/capabilities": "1.0.0",
--- a/.github/workflows/manual.yml
+++ b/.github/workflows/manual.yml
@@ -11,7 +11,6 @@ on:
           - access-api
           - upload-client
           - access-client
-          - wallet
           - docs
       environment:
         description: 'Environment to deploy'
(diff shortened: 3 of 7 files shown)


gobengo:
fyi I made a draft PR that you may wanna merge into this one and it might help a little bit with resolving conflicts with main
https://github.com/web3-storage/w3protocol/pull/244

packages/access-client/src/agent-data.js
```diff
@@ -0,0 +1,160 @@
...
+    return storedData
+      ? AgentData.fromExport(storedData)
+      : await AgentData.create(options.initialData, {
+          store: {
+            save: async (d) => {
+              await store.save(d)
+            },
+          },
+        })
+  }
```
hugomrdias, line 67: if we dont have a method to test if the store was initialized do we always call this method ? if yes we gonna need to always pass on a Signer just in case its empty ?

hugomrdias (changes requested):
in general looks good but now the user needs to choose store driver plus signer type maybe we can find a easier solution

packages/access-client/src/agent-data.js
alanshaw: > if we dont have a method to test if the store was initialized do we always call this method ?

Yes, this is probably what you use 99% of the time. Essentially `store.load()` is the init check (which is what is happening above).

> if yes we gonna need to always pass on a Signer just in case its empty ?

No because `AgentData.create` generates one for you. I understand that's not great if you want to use a different key type from what `AgentData` gives you by default...and in that situation you'd either always pass it (as you say), or do what this method is doing in your application code. I don't know how to make that a better experience other than picking a better default in the browser.

packages/access-client/src/agent.js
```diff
@@ -22,97 +22,63 @@ import {
...
+    this.connection =
+      options.connection ??
+      connection({
+        principal: options.servicePrincipal,
+        url: this.url,
+      })
+    this.data = data
 
     // private
-    this.#fetch = opts.fetch
```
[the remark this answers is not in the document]
hugomrdias: yes sure

packages/access-client/src/agent-data.js
hugomrdias: can we make the store.load() do store.open() ?

hugomrdias: humm so user needs to know about:
- Agent
- AgentData
- StoreDriver
- Signer

We can leave Signer in the options for flexibility, but right now BrowserDriver kinda requires RSASigner.

can we reduce this to:
- Agent
- Driver
- Signer (this would be optional, either from the driver or from the user overriding the driver)

Most uses would just need Agent and Driver, i think we can do this by just hiding AgentData inside Agent.

hugomrdias:
```js

const data = driver.load()

if(data) {
  agent.from(data, driver)
}else{
  agent.create(driver)
}
```

something like this ?

alanshaw: > ```js
> const data = driver.load()
> 
> if(data) {
>   agent.from(data, driver)
> }else{
>   agent.create(driver, { 
>     principal: await EdSigner.generate() // if we want to overriding Driver.generatePrincipal() 
>   })
> }
> ```
> 
> something like this ?

yes we can do that

hugomrdias (changes requested):
LGTM, left one inline comment. 

Maybe we could change the naming store to driver and hardcode the data type instead of using generics. Avoid having `/** @type {StoreConf<import('../types').AgentDataExport>} */` everywhere. These two are non blocking feel free to ignore.

hugomrdias (approved):

PR #229, opened 2022-12-01
feat!: upload/list items have a shards array property

olizilla:
- update tests and types to handle upload/list items with a shards array property.
- update upload/add fn to return an UploadAddResponse as shards out may be more than shards in.
- update README

see also: https://github.com/web3-storage/upload-api/pull/56

### DIFF
--- a/packages/upload-client/README.md
+++ b/packages/upload-client/README.md
@@ -347,7 +347,7 @@ function add(
   root: CID,
   shards: CID[],
   options: { retries?: number; signal?: AbortSignal } = {}
-): Promise<void>
+): Promise<UploadAddResponse>
 ```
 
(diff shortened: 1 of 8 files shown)


alanshaw (approved):
LGTM

olizilla:
@alanshaw plz to eyes minor change to this PR since your review https://github.com/web3-storage/w3protocol/pull/229/commits/d25f1b38b23aa3507f61e210eda431b5c97f55b9

`shards` are optional on upload/add invocation and response and updates the index.js test to echo back the root and shards sent in most cases. I made the first test check the dataCID and carCID against an expected car as it's the simplest one, and it seems reasonable to have a more concrete expectation there.... i then reaslised i didn't want to express any particular guarnatee about the expected CIDs for cars and roots for the other tests and figured out I could echo back the invocation `nb` (with an existence check to appease tsc).

If you prefer I can strip out the expected CAR from the first test here or leave it if useful.

vasco-santos (approved):
LGTM

Issue #237, opened 2022-12-01
access service & client should use `did:web:web3.storage`


gobengo:
Motivation:
* towards https://github.com/web3-storage/w3protocol/issues/183
  * does the 'access' part

Parts
* [x] upgrade to @ucanto/{interface,principal} 4.0.0 which is required to use non-key DIDs https://github.com/web3-storage/w3protocol/pull/238
* [x] update access-api to create a ucanto server with a principal whose id is `did:{newServiceDid}`
  * https://github.com/web3-storage/w3protocol/pull/275
* [x] update access-client connection factory to use `did:{newServiceDid}` as default principal
  * or at least make the new well-known principal id/key values be easy to import so e.g. [w3ui doesn't have to hardcode them](https://github.com/web3-storage/w3ui/blob/dc2139f5e00c9195c480ce5c98a78b4296713ac7/packages/uploader-core/src/index.ts#L71)
* [ ] `test-ucanto-connection` dev tool in access-client cli to test the ucanto connection by sending an invocation to the expected id
* [ ] update cli `test-ucanto-connection` to send invocation to new access id (i.e. `did:{newServiceDid}`)

hugomrdias:
is this the same https://github.com/web3-storage/w3protocol/issues/182 ? can we add this to the track issue ?

gobengo:
@hugomrdias it's part of that, but just the part for access-client and access-api. that ticket is for 'all the services' afaict.

> can we add this to the track issue ?

sure idc. I put it in the equivalent milestone so it appears in the gh project board

hugomrdias:
why did:web instead of did:dns ? @gobengo @Gozala

gobengo:
> why did:web instead of did:dns ? @gobengo @Gozala

@hugomrdias only prior discussion [was here](https://github.com/web3-storage/specs/pull/12#discussion_r1033968381).

In short, it's because did:web is easier to implement and try things out with.
* implementable using only a web server, which access-api and upload-api already are
  * did:dns impl would beg question of how to manage the dns rules, which depends on how we operate our dns server and whether we outsource it e.g. to cloudflare or route53
* [`did:web` spec](https://github.com/w3c-ccg/did-method-web/) is maintained by members of w3c ccg, which is a group we can go to for advice while implementing. [`did:dns` spec](https://danubetech.github.io/did-method-dns/) maintained by one company atm (even though it's a rad company :P).
[...]

PR #238, opened 2022-12-02
feat!: upgrade to `@ucanto/{interface,principal}`@^4.0.0

gobengo:
Motivation:
* part of https://github.com/web3-storage/w3protocol/issues/237
* we need this version of ucanto libs because they include updated types to handle non-key-dids and dids with more than one key

Parts
* [x] build/test works in packages/access-client
* [x] build/test works in packages/upload-client
  * [x] [blocked by needed ucanto release](https://github.com/web3-storage/ucanto/pull/163)
* [x] build/test works in packages/access-api
  * [x] [blocked by needed ucanto release](https://github.com/web3-storage/ucanto/pull/163)


packages/access-client/src/stores/types.ts
```diff
@@ -5,8 +5,8 @@ import {
   DelegationMeta,
   SpaceMeta,
 } from '../types.js'
-import { RSASigner } from '@ucanto/principal/rsa'
 import { SignerArchive, DID } from '@ucanto/interface'
+import { RSA } from '@ucanto/principal/.'
```
Gozala, line 9: Whaaat ? Is this same as

```suggestion
import { RSA } from '@ucanto/principal'
```

And if so why not use the above one ?

packages/access-client/src/stores/types.ts
```diff
@@ -5,8 +5,8 @@ import {
   DelegationMeta,
   SpaceMeta,
 } from '../types.js'
-import { RSASigner } from '@ucanto/principal/rsa'
 import { SignerArchive, DID } from '@ucanto/interface'
```
Gozala, line 8:
```suggestion
import { SignerArchive, DID, DIDKey } from '@ucanto/interface'
```

We have this one now so you can type less

packages/access-client/src/stores/types.ts
```diff
@@ -50,7 +50,7 @@ export interface IStore<T> {
 export interface StoreDataIDB {
   id: number
   meta: AgentMeta
-  principal: SignerArchive<RSASigner>
+  principal: SignerArchive<DID<'key'>, typeof RSA.signatureCode>
```
Gozala, line 53:
```suggestion
  principal: SignerArchive<DIDKey, typeof RSA.signatureCode>
```

packages/access-client/test/stores/store-indexeddb.browser.test.js
```diff
@@ -12,8 +12,11 @@ describe('IndexedDB store', () => {
     // principal private key is not extractable
     const archive = data.principal.toArchive()
     assert(!(archive instanceof Uint8Array))
-    assert(archive.key instanceof CryptoKey)
-    assert.equal(archive.key.extractable, false)
+    const keys = [...Object.entries(archive.keys)]
+    assert.equal(keys.length, 1)
+    const firstKey = keys[0][1]
+    assert(firstKey instanceof CryptoKey)
+    assert.equal(firstKey.extractable, false)
```
Gozala, line 19:
```suggestion
    const [[did, key], ...keys] = [...Object.entries(archive.keys)]
    assert.deepEqual(keys, [])
    assert(key instanceof CryptoKey)
    assert.equal(key.extractable, false)
```

I think this is easier to follow

packages/capabilities/src/voucher.js
```diff
@@ -31,7 +31,7 @@ export const Identity = URI.match({ protocol: 'mailto:' })
 /**
  * Services are identified using did:key identifier.
  */
-export const Service = DID.match({ method: 'key' })
+export const Service = DID.match({})
```
Gozala, line 34:
```suggestion
export const Service = DID.match({ method: 'web' })
```

Should the above be ^^ instead ? Also comment need to be updated

Gozala (approved):
Provided some suggestions, but it looks good to me.

Please make sure to use feat!: in the merge commit to make a major version bump

packages/access-client/test/stores/store-indexeddb.browser.test.js
gobengo: I think it would throw an unexpected error if keys was length 0, whereas the way I did it will throw an AssertionError. either is fine w/ me though.

packages/capabilities/src/voucher.js
gobengo: I intended the type to allow any did. specifically I wanted to at least allow both `did:key` and `did:web`.
Does it matter much right now? we can narrow the type later if needed.... it seems like at this capabilities layer it shouldn't care which did method?

Issue #236, opened 2022-12-01
Handle archive format change in clients


Gozala:
Recent changes https://github.com/web3-storage/ucanto/pull/149 in ucanto change an archive format. This has implications on the clients who will try to load signer from the previously stored archive in IndexedDB.

Previous structure was:

```ts
export type SignerArchive<Self extends Signer = Signer> =
  | ByteView<Self>
  | SignerInfo<Self>

export interface SignerInfo<Self extends Signer = Signer> {
  readonly did: ReturnType<Self['did']>
  readonly key: CryptoKey
}
```

New structure is:

```ts
/**
 * {@link Signer} keys and it's DID that can be used for persist and restore
 * signer across sessions.
 */
export interface SignerArchive<
  ID extends DID = DID,
  Alg extends SigAlg = SigAlg
> {
  /**
   * [DID Subject](https://www.w3.org/TR/did-core/#did-subject) for this
   * signer.
   */
  id: ID

  /**
   * Set of private keys this signer uses keyed by corresponding [did:key][].
   *
   * ⚠️ At the moment signers only support single key use case, however we may
   * change that in the future, which is why data model is forward designed to
   * support multiple keys.
   *
   * [did:key]:https://w3c-ccg.github.io/did-method-key/
   */
  keys: { [Key: DIDKey]: KeyArchive<Alg> }
}

/**
 * Represents a private key which will be in `CryptoKey` format if it is
 * non-extractable or is byte encoded when extractable.
 */
export type KeyArchive<Alg extends SigAlg = SigAlg> =
  | CryptoKey
  | ByteView<SignerKey<Alg> & CryptoKey>
```

If we ignore the possibility that keys may be in binary as opposed to `CryptoKey` upgrading old archive to new will boil down to:

```ts
const upgradeArchive = ({ did, key }) => ({ id: did, keys: {[did]:key} })
```

I think clients should probably add IndexDB migration thing to migrate all records to new format to avoid breakage

Gozala:
@alanshaw the issue we discussed earlier in the call

Gozala:
I could add `.fromLegacy(archive)` to the ucanto, but I think doing it inside the indexdb migration would be a better option

alanshaw:
There's no code in production that uses this yet so it's safe to change.

Gozala:
@alanshaw so should I just close this issue or do we want to still add something to the indexdb migration

PR #246, opened 2022-12-03
feat!: upgrade @ucanto/* and @ipld/dag-ucan major versions

gobengo:
Motivation:
* second attempt at this https://github.com/web3-storage/w3protocol/pull/238
  * after first was reverted https://github.com/web3-storage/w3protocol/pull/245

Blockers
* Wait for these two to land first (@hugomrdias says so)
  * [ ] @hugomrdias https://github.com/web3-storage/w3protocol/pull/207
  * [ ] @alanshaw https://github.com/web3-storage/w3protocol/pull/228


alanshaw (approved):
LGTM

Gozala (changes requested):
This looks great! We do however should address #236. I believe relevant code lives in https://github.com/web3-storage/w3protocol/blob/main/packages/access-client/src/drivers/indexeddb.js

I'm ok with doing it as a followup change, but it does seem like they should go together though.

Gozala (approved):
Talked to @gobengo & I think it makes sense to tackle db migration as followup and probably someone who's more familiar with code there.

PR #260, opened 2022-12-06
refactor: sync delegations method

### LINKED ISSUE #219: Remove side effect from access client agent delegations method
Removing the expired delegations when fetching the list of delegations makes this method async and forces a whole bunch of other methods to be async that don't need to be. Please can we remove the side effect?

alanshaw:
By removing the side effect many methods that read data in memory no longer have to be async. I think this makes the API simplier.

I've added a _method_ that can be called to clean up expired delegations, which shifts the responsibility onto the user. Alternatively we could move the clean up into the methods that write delegations? Open to other suggestions.

resolves #219

### DIFF
--- a/packages/access-client/src/agent.js
+++ b/packages/access-client/src/agent.js
@@ -194,9 +194,15 @@ export class Agent {
    * @param {import('@ucanto/interface').Capability[]} [caps] - Capabilities to filter by. Empty or undefined caps with return all the proofs.
    */
   proofs(caps) {
-    return this.#delegations(caps)
-      .filter((v) => v.delegation.audience.did() === this.issuer.did())
-      .map((v) => v.delegation)
+    const arr = []
+
+    for (const value of this.#delegations(caps)) {
+      if (value.delegation.audience.did() === this.issuer.did()) {
+        arr.push(value.delegation)
+      }
+    }
+
+    return arr
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/access-client/src/agent.js
```diff
@@ -182,40 +193,30 @@ export class Agent {
...
-      if (value.delegation.audience.did() === this.issuer.did()) {
-        arr.push(value.delegation)
-      }
-    }
-
-    return arr
+  proofs(caps) {
+    return this.#delegations(caps)
+      .filter((v) => v.delegation.audience.did() === this.issuer.did())
+      .map((v) => v.delegation)
```
hugomrdias, line 199: can we just remove the `await` word and keep the for of ?

packages/access-client/src/agent.js
```diff
@@ -149,27 +149,38 @@ export class Agent {
...
-        // delete any expired delegation
+      }
+    }
+    return values
+  }
+
+  /**
+   * Clean up any expired delegations.
+   */
+  async removeExpiredDelegations() {
```
hugomrdias, line 181: can we call this methods somewhere ? like `fromStore` or any other place thats already async ?

hugomrdias (changes requested):

packages/access-client/src/agent.js
alanshaw: I've added to `addProof` and `delegate` for your approval. Makes sense to me if we're mutating delegations then we can remove some old ones.

hugomrdias (approved):
LGTM

PR #275, opened 2022-12-08
feat: access-api uses DID env variable when building its ucanto server id

gobengo:
Motivation:
* https://github.com/web3-storage/w3protocol/issues/237

Notes
* This is intended to be safe to merge/deploy as is. The new functionality can be opted-into via env vars
* The ucanto server will only be constructed with an opts.id that is a signer `withDID(...)` iff `process.env.DID` is set to a DID
  * If `process.env.DID` is unset, the ucanto server id signer will have a `did()` corresponding to `config.PRIVATE_KEY` (it will be a `did:key` of the corresponding public key)
* Before this PR, many tests of the access-api worker were configured implicitly via dotenv with whatever the `/.env.local` file happened to contain. I added a worker test that asserts functionality that is dependent on the details of the env vars, so I made an affordance for [tests to be able to configure the access-api worker with environment variables defined in the test case itself](https://github.com/web3-storage/w3protocol/pull/275/files#diff-e20ffc550d006747790e7b67da280446c1cd1d2d4f72ccb5df04f62cbb6423daR149).

### DIFF
--- a/packages/access-api/src/config.js
+++ b/packages/access-api/src/config.js
@@ -32,11 +32,14 @@ export function loadConfig(env) {
     }
   }
 
+  const DID = env.DID
+  const PRIVATE_KEY = vars.PRIVATE_KEY
+  const signer = configureSigner({ DID, PRIVATE_KEY })
   return {
     DEBUG: boolValue(vars.DEBUG),
     ENV: parseRuntimeEnv(vars.ENV),
 
-    PRIVATE_KEY: vars.PRIVATE_KEY,
+    PRIVATE_KEY,
     POSTMARK_TOKEN: vars.POSTMARK_TOKEN,
     SENTRY_DSN: vars.SENTRY_DSN,
     LOGTAIL_TOKEN: vars.LOGTAIL_TOKEN,
@@ -52,7 +55,8 @@ export function loadConfig(env) {
     // eslint-disable-next-line no-undef
     COMMITHASH: ACCOUNT_COMMITHASH,
 
-    DID: env.DID,
+    DID,
+    signer,
 
     // bindings
     METRICS:
--- a/packages/access-api/src/config.js
+++ b/packages/access-api/src/config.js
@@ -39,7 +39,6 @@ export function loadConfig(env) {
     DEBUG: boolValue(vars.DEBUG),
     ENV: parseRuntimeEnv(vars.ENV),
 
-    PRIVATE_KEY,
     POSTMARK_TOKEN: vars.POSTMARK_TOKEN,
     SENTRY_DSN: vars.SENTRY_DSN,
     LOGTAIL_TOKEN: vars.LOGTAIL_TOKEN,
@@ -55,7 +54,6 @@ export function loadConfig(env) {
     // eslint-disable-next-line no-undef
     COMMITHASH: ACCOUNT_COMMITHASH,
 
-    DID,
     signer,
 
     // bindings
--- a/packages/access-api/src/config.js
+++ b/packages/access-api/src/config.js
@@ -1,3 +1,4 @@
+import { DID } from '@ucanto/validator'
 import { Signer } from '@ucanto/principal/ed25519'
 
 /**
@@ -119,13 +120,14 @@ export function createAnalyticsEngine() {
  */
 export function configureSigner(config) {
   const signer = Signer.parse(config.PRIVATE_KEY)
-  if (config.DID) {
-    if (!isDID(config.DID)) {
-      throw new Error(`Invalid DID: ${config.DID}`)
-    }
-    return signer.withDID(config.DID)
+  const did = config.DID
+  if (!did) {
+    return signer
   }
-  return signer
+  if (!isDID(did)) {
+    throw new Error(`Invalid DID: ${did}`)
+  }
+  return signer.withDID(did)
 }
 
 /**
@@ -136,9 +138,8 @@ export function configureSigner(config) {
  * @returns {object is `did:${string}:${string}`}
  */
 function isDID(object) {
-  if (typeof object !== 'string') return false
-  const parts = object.split(':')
-  if (parts.length <= 2) return false
-  if (parts[0] !== 'did') return false
-  return true
+  try {
+    return Boolean(DID.match({}).from(object))
+  } catch {}
+  return false
 }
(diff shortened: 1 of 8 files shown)


packages/access-api/src/config.js
```diff
@@ -105,3 +109,36 @@ export function createAnalyticsEngine() {
...
+ * Given a config, return a ucanto Signer object representing the service
+ *
+ * @param {object} config
+ * @param {string} [config.DID] - public identifier of the running service. e.g. a did:key or a did:web
+ * @param {string} config.PRIVATE_KEY - multiformats private key of primary signing key
+ */
+export function configureSigner(config) {
+  const signer = Signer.parse(config.PRIVATE_KEY)
+  if (config.DID) {
+    if (!isDID(config.DID)) {
```
Gozala, line 123: I think it may be better to use [existing functionality](https://github.com/web3-storage/ucanto/blob/main/packages/validator/src/schema/did.js) for this, for example this will parse and throw if it's invalid:

```ts
DID.match({ method: 'web' }).from(config.DID)
```

It will also take care of refusing `did:key` which would be invalid in this case.

Gozala (approved):
Looks good to me! Thanks

packages/access-api/src/config.js
```diff
@@ -136,9 +138,8 @@ export function configureSigner(config) {
  * @returns {object is `did:${string}:${string}`}
  */
 function isDID(object) {
-  if (typeof object !== 'string') return false
-  const parts = object.split(':')
-  if (parts.length <= 2) return false
-  if (parts[0] !== 'did') return false
-  return true
+  try {
+    return Boolean(DID.match({}).from(object))
```
Gozala, line 142: What’s the point of catching here to then throw ? You can simply just do .from instead of whole isDID thing.

Alternatively there’s also .is method which you could instead which will return boolean

packages/access-api/src/utils/context.js
```diff
@@ -42,12 +41,12 @@ export function getContext(request, env, ctx) {
     env: config.ENV,
   })
 
-  const keypair = Signer.parse(config.PRIVATE_KEY)
+  const signer = configureSigner(config)
```
hugomrdias, line 44: can you do this inside the `loadConfig` ?

packages/access-api/src/utils/context.js
```diff
@@ -42,12 +41,12 @@ export function getContext(request, env, ctx) {
...
   const url = new URL(request.url)
   const db = new D1QB(config.DB)
   return {
     log,
-    signer: keypair,
+    signer,
```
hugomrdias, line 49:
```suggestion
    signer: Signer.parse(config.PRIVATE_KEY).withDID(config.ENV === 'production' ? 'did:web:ucan.web3.storage': `did:web:ucan-${config.ENV}.web3.storage`),
```
can't we just do this ? any error will be caught at dev time or CI.

packages/access-api/test/ucan.test.js
```diff
@@ -144,6 +144,31 @@ describe('ucan', function () {
     t.deepEqual(rsp, ['test pass'])
   })
 
+  test('should support ucan invoking to a did:web aud', async function () {
+    const serviceDidWeb = 'did:web:web3.storage'
+    const { mf, issuer, service } = await context({
+      environment: {
+        ...process.env,
+        PRIVATE_KEY:
+          'MgCYWjE6vp0cn3amPan2xPO+f6EZ3I+KwuN1w2vx57vpJ9O0Bn4ci4jn8itwc121ujm7lDHkCW24LuKfZwIdmsifVysY=',
```
hugomrdias, line 153: why do you need a new key here ?

hugomrdias (changes requested):

packages/access-api/src/utils/context.js
gobengo: 👍 https://github.com/web3-storage/w3protocol/pull/275/commits/b9533c92d0b8080f9f6d14c240a89f370d6ad7f7

packages/access-api/src/utils/context.js
gobengo: I want the advertised DID to come from env variable. A '.web3.storage' domain won't be the best choice 100% of the times, e.g. in localdev or testing various scenarios. env variable gives me flexibility to get something working. once something works/tests end to end, can narrow down the configurability.

packages/access-api/test/ucan.test.js
gobengo: I don't. Good catch. 🙏  https://github.com/web3-storage/w3protocol/pull/275/commits/d20433140493ecf5408c0e6edb49ce654ec01f12

packages/access-api/src/config.js
gobengo: 👍 https://github.com/web3-storage/w3protocol/pull/275/commits/5184a554ee80f29ddb74ddd9c78e3b883c4c8119

packages/access-api/src/config.js
```diff
@@ -49,6 +55,9 @@ export function loadConfig(env) {
     // eslint-disable-next-line no-undef
     COMMITHASH: ACCOUNT_COMMITHASH,
 
+    DID,
+    signer,
```
hugomrdias, line 59: can we get the did:key and did:web from the signer instance ?

packages/access-api/src/config.js
```diff
@@ -29,11 +32,14 @@ export function loadConfig(env) {
...
 
+  const DID = env.DID
+  const PRIVATE_KEY = vars.PRIVATE_KEY
+  const signer = configureSigner({ DID, PRIVATE_KEY })
   return {
     DEBUG: boolValue(vars.DEBUG),
     ENV: parseRuntimeEnv(vars.ENV),
 
-    PRIVATE_KEY: vars.PRIVATE_KEY,
+    PRIVATE_KEY,
```
hugomrdias, line 42: we probably dont need this one anymore right ?

hugomrdias (changes requested):

packages/access-api/src/config.js
gobengo: https://github.com/web3-storage/w3protocol/pull/275/commits/852951f932d97fc651c08258fdc6f4a7b15ce658

hugomrdias (approved):
LGTM

PR #276, opened 2022-12-09
feat: sync encode/decode delegations

hugomrdias:


packages/access-client/src/encoding.js
```diff
@@ -13,89 +13,77 @@
...
   }
-  await writer.close()
-
-  const bytes = await collection
-
-  return u8.toString(bytes, encoding)
-}
+  const headerLength = CarBufferWriter.estimateHeaderLength(roots.length)
+  const writer = CarBufferWriter.createWriter(
+    new ArrayBuffer(headerLength + byteLength),
```
alanshaw, line 55: This is kinda weird seeing as we have all the blocks already. Would be nicer if it just collected all the buffers and joined them together when we call `close()`

Issue #302, opened 2022-12-13
voucher/redeem breaks when env.DID is set, because it issues a UCAN as a did:web, which we don't yet know how to verify signature of


gobengo:
Problem
* https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/src/service/voucher-claim.js#L12
* that line issues a ucan as did:web
* later the ucan signatures can't be resolved because verification code doesn't know how to verify signatures for a `did:web`

What to change:
* No longer do `.withDID(....)` on the underlying `ctx.signer`. I think we may not need to have the 'signer' know about a DID other than its `did:key`.
* All we actually need is for the ucanto server id to have the right did
  * so just use an explicit object for that [here](https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/src/routes/raw.js#L11) instead of passing `id: signer`

gobengo:
gonna fix this way https://github.com/web3-storage/w3protocol/pull/312

PR #303, opened 2022-12-13
fix: access-api ctx.signer no longer uses env.DID. instead env.DID is only used for ucanto server id

gobengo:
Motivation:
* https://github.com/web3-storage/w3protocol/issues/302


Gozala (approved):
Looks good to me! Mind also creating an issue to revert this after we propagate https://github.com/web3-storage/ucanto/pull/168

gobengo:
ty @Gozala I made this issue https://github.com/web3-storage/w3protocol/issues/304

PR #309, opened 2022-12-14
feat: include ucanto server principal did as 'aud' key in /version endpoint

gobengo:
Motivation:
* make it easy to discover the ucanto server principal
* unlike [reverted PR](https://github.com/web3-storage/w3protocol/pull/305), don't change `did` property
  * because at least `@web3-storage/access` cli expects to be able to parse that value as a Verifier, which it can only do when the `did` is a `did:key`


alanshaw (approved):
Makes sense - thanks for fixing, reverting and refixing :)

Issue #316, opened 2022-12-14
Have an endpoint where we print out did:key along with did:web for the API


Gozala:
Now that we use did:web it's not always obvious which did:key's are used, we should print out both.

gobengo:
done
https://access.web3.storage/version
https://staging.up.web3.storage/version

PR #317, opened 2022-12-14
feat(access-api): access-api version endpoint has publicKey property

gobengo:
...which is always the underlying did:key publicKey, even if signer.did() is a did:web

Motivation
* https://github.com/web3-storage/w3protocol/issues/316
* historically there has been a `did` property, which is (historically) always a `did:key` (so a public key)
* now we deal with non-key dids like `did:web`. So the did may not be identical to the public key (you have to resolve a `did:web` to determine its pubkeys)
* we should have a property that is explicitly only ever a public key (serialized as a `did:key` string)
  * e.g. for code like this to rely on: https://github.com/web3-storage/w3protocol/blob/main/packages/access-client/src/cli/utils.js#L40 (which I think will break when version endpoint's `did` property is a `did:web`)

Rants
* try to avoid calling things 'did' when you mean 'public key'. Or assuming that things called 'did' will always be a `did:key`
* we often say 'did' or 'did:key' when we mean 'keypair', which is confusing because:
  * 1. a 'did' may have zero keypairs
  * 2. a 'did:key', on its own, has no private key. You can't make a 'signer' from only a `did:key` string.

### DIFF
--- a/packages/access-api/src/bindings.d.ts
+++ b/packages/access-api/src/bindings.d.ts
@@ -5,6 +5,7 @@ import { Email } from './utils/email.js'
 import { Spaces } from './models/spaces.js'
 import { Validations } from './models/validations.js'
 import { loadConfig } from './config.js'
+import { Signer as EdSigner } from '@ucanto/principal/ed25519'
 
 export {}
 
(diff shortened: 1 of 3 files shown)


packages/access-api/src/bindings.d.ts
```diff
@@ -42,7 +43,7 @@ export interface Env {
 
 export interface RouteContext {
   log: Logging
-  signer: Signer
```
gobengo, line 45: This `Signer` didn't resolve to anything before in my `tsc` setup, so was treated as `any`

alanshaw (approved):

Issue #259, opened 2022-12-06
validation flow should show which DIDs are being delegated access


heyjay44:
To update design related to https://github.com/web3-storage/w3protocol/pull/257

ayenisholah:
Hello, I would like to help with this, please could you add more description.

jchris:
When a user signs up for a new account in w3ui, they are prompted to click the link in their email. Currently the email is generic and the same across all invocations. In the mature product, we need to put the "choose which space to run this app in" workflow in place. 

Until then we can do intelligent defaults, but we are constrained in that we want people to be able to use w3ui without any other dependencies or API keys.

So the default flow will look like a user creating a new space (local private key) and delegating themselves (email key) access to all the capabilities on it. The agent used by the app code will get whatever access the user gets at this point. So that the app can upload and list.

By default we want apps to work without the dev needing an API key, so in most cases we'll be saying "user device agent is granting user email key capabilities to manage this newly provisioned user owned space"

However we will want to give users a shortcut for choosing the correct space when they launch the app again on a new device. So maybe we want to have some affinity for the site origin.

jchris:
Maybe useful to start reading here https://github.com/web3-storage/specs/blob/main/w3-account.md

jchris:
related to #291

Issue #291, opened 2022-12-13
add redirect URL to validation request UCAN


jchris:
This will be in a `fact` field.

We also need to add type validation in the fact fields we use.

This enables apps to redirect back to themselves.

* Voucher claim needs to have the redirect URL in the fact, and the voucher redeem in the email also need to have it.

Gozala:
Any reason why it should be in `fct` as opposed to inside an `nb`  instead ? I would like to make case for later because given the way https://github.com/ucan-wg/invocation/pull/1 is shaping up it's not clear if we'll have a same `fct` for a whole batch of invocations or a single one making things trickier.

It's also `fct` is an array so it would require additional signaling which would complicate matter even more

jchris:
As mentioned in #318, we should also put the space nickname in the protocol so it can be in the email

PR #318, opened 2022-12-15
wip: first pass at including space did info in verification email

travis:
First pass at including space DID info in verification email, per #259:

Screen Shot 2022-12-14 at 4 29 20 PM

I think this is unsatisfactory - DIDs aren't designed to be shown to humans and people will likely mostly ignore them. There are a number of potential solutions:

1) Just be ok with showing DIDs to users for now and figure out a more user-friendly solution later. This is probably ok - users who care can look at the DID of the space they just created and then eyeball it against the one in the email. 
2) Use Gravatars of DIDs in the email. We'd need to massage the messaging to make this easy for users to understand, but @alanshaw demonstrated how it can be done in https://github.com/web3-storage/w3ui/pull/128
3) Show the name of the space the user just created. A couple problems with this: 
  a) I don't think we save the name outside of the user's browser at the moment (tho could be wrong.. hope I'm wrong?!) 
  b) users might have multiple spaces with the same name, so this may not be unambiguous

Opening this draft pull request to kick off discussion and design of how we actually want this to work. Please weigh in!

### DIFF
--- a/packages/access-api/src/service/voucher-claim.js
+++ b/packages/access-api/src/service/voucher-claim.js
@@ -45,6 +45,7 @@ export function voucherClaimProvider(ctx) {
     await ctx.email.sendValidation({
       to: capability.nb.identity.replace('mailto:', ''),
       url,
+      space: inv.capabilities[0].nb.space,
     })
   })
 }
--- a/packages/access-api/src/utils/email.js
+++ b/packages/access-api/src/utils/email.js
@@ -15,7 +15,7 @@ export class Email {
   /**
    * Send validation email with ucan to register
    *
-   * @param {{ to: string; url: string }} opts
+   * @param {{ to: string; url: string, space: string }} opts
    */
   async sendValidation(opts) {
     const rsp = await fetch('https://api.postmarkapp.com/email/withTemplate', {
@@ -30,6 +30,7 @@ export class Email {
           product_name: 'Web3 Storage',
           email: opts.to,
           action_url: opts.url,
+          space_did: opts.space,
         },
       }),
     })


jchris:
I think we should coalesce around gravitas for consistency. Maybe we can also pass the space nickname from browser via UCAN through the protocol like https://github.com/web3-storage/w3protocol/issues/291

Gozala:
Few notes:

1. There is only accept option and nothing about the case when you have not requested access. We should at least tell user not to click it or better yet have a deny button. Often they tend to be red buttons or red text so to bring the attention.
2. This made me realize that auth the way we do it now over websocket, is not ideal because:
     - Attacker may attempt to get access to user space by making them click the validate button.
     - With redirect link back to site we can avoid this problem, because even if user clicks the link they authorize agent in the target site as opposed to whoever issued request.
     - This also makes me wonder if attackers may fabricate their own emails with URLs to their site to gain access to the account
3. I realize this designed for voucher stuff and not [accounts](https://github.com/web3-storage/specs/blob/main/w3-account.md) but perhaps it's a good idea to consider later as that's where we going to go next.
   - There we don't have a notion of registering space
   - Instead you authorize agent to act on behalf of your account
   - DID would be a agent DID not the space DID
4. I think some visual like gravatar would be a lot more helpful that printing out DID
   - Most gravatars sadly look lame, but [urbit designed some remarkable beautiful ones](https://urbit.org/blog/creating-sigils) I wonder if we could do something along the same lines

travis:
oo for posterity, I found the quote I was looking for that forms the basis of my feeling that showing the DID to the user is bad - I don't think there's any disagreement, but worth thinking about in this context: https://github.com/cwebber/rebooting-the-web-of-trust-spring2018/blob/petnames/draft-documents/petnames.md

hugomrdias:
I think we should revisit this after or during implementation of providers and accounts in our current phase.

travis:
closing in favor of the work in https://github.com/web3-storage/w3protocol/pull/399

PR #319, opened 2022-12-15
fix: use nullish coalescing for audience

alanshaw:
When explicitly passing `null` a default value is not set.


alanshaw:
LGTM 🙈

PR #323, opened 2022-12-16
fix: better upload-api service types

alanshaw:


packages/upload-client/src/types.ts
```diff
@@ -31,7 +31,7 @@ export interface Service {
   store: {
     add: ServiceMethod<StoreAdd, StoreAddResponse, never>
     list: ServiceMethod<StoreList, ListResponse<StoreListResult>, never>
-    remove: ServiceMethod<StoreRemove, null, never>
+    remove: ServiceMethod<StoreRemove, void, never>
```
Gozala, line 34: I think we actually can not use `void` because IPLD can not encode `undefined`, which is why it was `null`. That said I think it's best to define such returns as `{}` as it allows future extensions in backwards compatible way.

```suggestion
    remove: ServiceMethod<StoreRemove, {}, never>
```

packages/upload-client/test/store.test.js
```diff
@@ -516,7 +530,6 @@ describe('Store.remove', () => {
           assert.equal(invCap.can, StoreCapabilities.remove.can)
           assert.equal(invCap.with, space.did())
           assert.equal(String(invCap.nb?.link), car.cid.toString())
```
Gozala, line 532:
```suggestion
          assert.equal(String(invCap.nb?.link), car.cid.toString())
          return {}
```

Gozala (approved):
Looks good, provided some minor suggestions

Issue #325, opened 2023-01-04
create single api endpoint for all capabilities


gobengo:

gobengo:
this is a little related to https://github.com/web3-storage/w3protocol/issues/182

gobengo:
#status you can now issue `store/list` invocations against access-api
[test here](https://github.com/gobengo/monorepo/blob/997cf85472a2f887d5a3ac60dd1127fa6f98de5a/experiments/w3protocol-test/src/w3protocol.test.ts#L16)

`store/add` or other upload-api service methods that validate ucan proofs error like this:
<details>
<summary>error from <a href="https://github.com/gobengo/monorepo/pull/7/files#diff-706e9ee739c12b0b8ca50cd72c2c6708c04c0bf40184239b5b94756a5708c0ceR99">testing store/add against staging access-api using upload-client</a></summary>
<pre>
    # error uploading file
    # error cause {
    #   name: 'Error',
    #   error: true,
    #   stack: 'Error: did:web:staging.web3.storage is not delegated capability store/add on did:key:z6MkoeDLAk1UP48Dc2FT3JdD9G8vB4vnEaeFQr5QVSVf2SJv\\n' +
    #     '    at null.<anonymous> (/upload-api/service/store/add.js:36:16)\\n' +
    #     '    at processTicksAndRejections (node:internal/process/task_queues:96:5)\\n' +
    #     '    at null.invoke2 (/node_modules/@ucanto/server/src/server.js:129:14)\\n' +
    #     '    at null.execute2 (/node_modules/@ucanto/server/src/server.js:90:18)\\n' +
    #     '    at null.handle (/node_modules/@ucanto/server/src/server.js:71:18)\\n' +
    #     '    at null.ucanInvocationRouter (/upload-api/functions/ucan-invocation-router.js:87:20)\\n' +
    #     '    at Runtime.handler (/node_modules/src/awslambda.ts:332:1)',
    #   message: 'did:web:staging.web3.storage is not delegated capability store/add on did:key:z6MkoeDLAk1UP48Dc2FT3JdD9G8vB4vnEaeFQr5QVSVf2SJv'
    # }
    # Subtest: w3protocol-test can upload file
</pre>
</details>

Getting that to work will require https://github.com/web3-storage/ucanto/issues/195

gobengo:
#status I will update it to use new ucanto functionality to do the proxying without issuing a new delegation - https://github.com/web3-storage/w3protocol/issues/383

[...]

PR #328, opened 2023-01-06
feat: unify ucan api endpoints

gobengo:
Motivation:
* https://github.com/web3-storage/w3protocol/issues/325

Parts
* [x] poc proxying to access-api
* [x] poc proxying to upload-api
* [x] make upstreams and route function configurable
* [x] staging deploy to cf workers
* [ ] test [dev url](https://refs-heads-325-unify-ucan-api-endpoints.protocol-labs.workers.dev/) with access-client
* [ ] test [dev url](https://refs-heads-325-unify-ucan-api-endpoints.protocol-labs.workers.dev/) with upload-client
* [ ] script to run tests against an arbitrary url
  * [ ] run the script in github workflow after deploying to cf workers
* [ ] better GET / response than 'hello world' (e.g. explain what you're looking at)
* [ ] test multiple invocations to same service in same ucanto request
* [ ] test multiple invocations to different services in same ucanto request
  * needs to split into multiple upstream requests and await response

### DIFF
--- a/packages/upload-client/src/types.ts
+++ b/packages/upload-client/src/types.ts
@@ -31,7 +31,7 @@ export interface Service {
   store: {
     add: ServiceMethod<StoreAdd, StoreAddResponse, never>
     list: ServiceMethod<StoreList, ListResponse<StoreListResult>, never>
-    remove: ServiceMethod<StoreRemove, null, never>
+    remove: ServiceMethod<StoreRemove, {}, never>
   }
   upload: {
     add: ServiceMethod<UploadAdd, UploadAddResponse, never>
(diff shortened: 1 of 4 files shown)


packages/api-gateway/src/index.js
```diff
@@ -0,0 +1,202 @@
...
+   * @returns {Promise<ucanto.Invocation>}
+   */
+  fromRequest: async (request) => {
+    const invocations = await carTransport.decode({
+      headers: Object.fromEntries(request.headers.entries()),
+      body: request.body
+        ? await toBuffer(toIt(request.body))
+        : Uint8Array.from([]),
+    })
+    if (invocations.length !== 1) {
```
Gozala, line 71: We should requests to carry a single invocations, it's been designed to send batches of invocations in a single request.

Also I think you don't need to do most of these, you could simply define a server just like you usually would except don't wrap handler in [`provider`](https://github.com/web3-storage/ucanto/blob/main/packages/server/src/handler.js) that way handler will be called without performing any kind of validation and you'll be able to execute it on desired connection.

https://github.com/web3-storage/w3protocol/blob/4cfe312c9501c2c595a2c7f3fe768d58be1dbc7b/packages/access-api/src/service/index.js#L24-L31

gobengo:
lots of ugly commits for what was ultimately a silly mistake in my github workflow yaml. My bad.

Good news is now commits to api-gateway will get a dev deployment, e.g. https://refs-heads-325-unify-ucan-api-endpoints.protocol-labs.workers.dev/

tomorrow I will test it using access-client and upload-client

gobengo:
Status
* approach changed after [decision 2](https://github.com/web3-storage/w3protocol/issues/325#decision-2-websockets) to not have a new api-gateway package as I was making here, but instead to just add functionality into access-api

PR #329, opened 2023-01-06
feat: add uploadCAR function

alanshaw:
The PR adds an `uploadCAR` function to the client.

It differs from `Store.add` because it automatically shards the CAR file, calls `Store.add` for each shard and finally registers an upload (using `Upload.add`). This is very similar to `uploadFile` or `uploadDirectory` except it skips the UnixFS encoding step.

Once merged, I'll propogate this to the https://github.com/web3-storage/w3cli so that @lanzafame can upload BIG CARs and have them automatically sharded.

### DIFF
--- a/packages/upload-client/README.md
+++ b/packages/upload-client/README.md
@@ -175,6 +175,7 @@ function uploadDirectory(
     signal?: AbortSignal
     onShardStored?: ShardStoredCallback
     shardSize?: number
+    concurrentRequests?: number
   } = {}
 ): Promise<CID>
 ```
@@ -196,6 +197,7 @@ function uploadFile(
     signal?: AbortSignal
     onShardStored?: ShardStoredCallback
     shardSize?: number
+    concurrentRequests?: number
   } = {}
 ): Promise<CID>
 ```
@@ -217,6 +219,7 @@ function uploadCAR(
     signal?: AbortSignal
     onShardStored?: ShardStoredCallback
     shardSize?: number
+    concurrentRequests?: number
   } = {}
 ): Promise<void>
 ```
--- a/packages/upload-client/README.md
+++ b/packages/upload-client/README.md
@@ -139,6 +139,8 @@ await Upload.add(conf, rootCID, carCIDs)
 - [API](#api)
   - [`uploadDirectory`](#uploaddirectory)
   - [`uploadFile`](#uploadfile)
+  - [`uploadCAR`](#uploadcar)
+  - [`CAR.BlockStream`](#carblockstream)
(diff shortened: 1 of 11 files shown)


packages/upload-client/README.md
```diff
@@ -204,6 +206,35 @@ Required delegated capability proofs: `store/add`, `upload/add`
...
+### `uploadCAR`
+
+```ts
+function uploadCAR(
+  conf: InvocationConfig,
+  car: Blob,
+  options: {
+    retries?: number
+    signal?: AbortSignal
+    onShardStored?: ShardStoredCallback
```
vasco-santos, line 218: should we also add concurrent requests?

vasco-santos (approved):
LGTM

packages/upload-client/src/index.js
```diff
@@ -68,6 +68,39 @@ export async function uploadDirectory(conf, files, options = {}) {
...
+ * The `proofs` are a set of capability delegations that prove the issuer
+ * has the capability to perform the action.
+ *
+ * The issuer needs the `store/add` and `upload/add` delegated capability.
+ * @param {import('./types').BlobLike} car CAR file.
+ * @param {import('./types').UploadOptions} [options]
+ */
+export async function uploadCAR(conf, car, options = {}) {
+  const blocks = new CAR.BlockStream(car)
+  options.rootCID = options.rootCID ?? (await blocks.getRoots())[0]
```
olizilla, line 100: if a CAR contains multiple roots and all the blocks for them, we could store the CAR and then create an upload for each root in the CAR. Minor tho... could be just be raised as an issue for a subsequent PR.

olizilla (approved):
Nice!

packages/upload-client/README.md
```diff
@@ -253,7 +284,7 @@ function add(
 ): Promise<CID>
 ```
 
-Store a CAR file to the service.
+Store a CAR file to the service. Returns the CID of the CAR file stored.
```
gobengo, line 287: Is the returned CID the CID of the input car file? Or might it be a different CID due to sharding? If the latter, this might be more accurate (but idk?)
```suggestion
Store a CAR file to the service. Returns the CID of the sharded CAR file.
```

alanshaw: `Store.add` does not shard, it stores whatever CAR file you pass it. This is part of the reason for the `uploadCAR` function.

packages/upload-client/src/index.js
alanshaw: https://github.com/web3-storage/w3protocol/issues/335

PR #330, opened 2023-01-10
chore: remove unnecessary step from api-gateway workflow

gobengo:
Motivation:
* testing workflow for when this source branch is deleted after PR merge (it should call `wrangler delete`)


gobengo:
It didn't run, I think because github might require the workflow with `on: delete` to be in the main branch.

Issue #331, opened 2023-01-10
make it so access-api can proxy store and upload invocations to upload-api


gobengo:
Motivation:
* https://github.com/web3-storage/w3protocol/issues/325
* so there is at least one URL that can handle all ucanto invocations + access-client websocket

PR #334, opened 2023-01-11
feat: access-api forwards store/ and upload/ invocations to upload-api

gobengo:
Motivation:
* https://github.com/web3-storage/w3protocol/issues/331

### DIFF
--- a/packages/access-api/src/config.js
+++ b/packages/access-api/src/config.js
@@ -1,7 +1,7 @@
 import { Signer } from '@ucanto/principal/ed25519'
-import * as DID from '@ipld/dag-ucan/did'
 // eslint-disable-next-line no-unused-vars
-import * as dagUcan from '@ipld/dag-ucan'
+import { UCAN } from '@ucanto/interface'
+import { DID } from '@ucanto/core'
 
 /**
  * Loads configuration variables from the global environment and returns a JS object
@@ -132,9 +132,9 @@ export function configureSigner(config) {
 }
 
 /**
- * @template {dagUcan.DID} ConfigDID
- * @template {dagUcan.DID} [ID=dagUcan.DID]
- * @template {dagUcan.SigAlg} [Alg=dagUcan.SigAlg]
+ * @template {UCAN.DID} ConfigDID
+ * @template {UCAN.DID} [ID=UCAN.DID]
+ * @template {UCAN.SigAlg} [Alg=UCAN.SigAlg]
  * @param {object} config
  * @param {ConfigDID} [config.DID] - public DID for the service
  * @param {import('@ucanto/interface').Verifier<ID,Alg>} verifier
--- a/packages/access-api/src/config.js
+++ b/packages/access-api/src/config.js
@@ -1,6 +1,6 @@
 import { Signer } from '@ucanto/principal/ed25519'
 // eslint-disable-next-line no-unused-vars
-import { UCAN } from '@ucanto/interface'
+import * as UCAN from '@ucanto/interface'
 import { DID } from '@ucanto/core'
 
 /**
--- a/packages/access-api/src/config.js
+++ b/packages/access-api/src/config.js
@@ -133,12 +133,11 @@ export function configureSigner(config) {
 
 /**
  * @template {UCAN.DID} ConfigDID
- * @template {UCAN.DID} [ID=UCAN.DID]
  * @template {UCAN.SigAlg} [Alg=UCAN.SigAlg]
  * @param {object} config
  * @param {ConfigDID} [config.DID] - public DID for the service
- * @param {import('@ucanto/interface').Verifier<ID,Alg>} verifier
- * @returns {import('@ucanto/interface').Verifier<ConfigDID|ID,Alg>}
+ * @param {import('@ucanto/interface').Verifier<ConfigDID,Alg>} verifier
+ * @returns {import('@ucanto/interface').Verifier<ConfigDID,Alg>}
  */
 export function configureVerifier(config, verifier) {
   if (config.DID) {
--- a/packages/access-api/src/config.js
+++ b/packages/access-api/src/config.js
@@ -1,5 +1,7 @@
 import { Signer } from '@ucanto/principal/ed25519'
 import * as DID from '@ipld/dag-ucan/did'
+// eslint-disable-next-line no-unused-vars
+import * as dagUcan from '@ipld/dag-ucan'
 
 /**
  * Loads configuration variables from the global environment and returns a JS object
@@ -127,4 +129,20 @@ export function configureSigner(config) {
     return signer.withDID(DID.parse(config.DID).did())
   }
   return signer
+}
+
+/**
+ * @template {dagUcan.DID} ConfigDID
+ * @template {dagUcan.DID} [ID=dagUcan.DID]
+ * @template {dagUcan.SigAlg} [Alg=dagUcan.SigAlg]
+ * @param {object} config
+ * @param {ConfigDID} [config.DID] - public DID for the service
+ * @param {import('@ucanto/interface').Verifier<ID,Alg>} verifier
+ * @returns {import('@ucanto/interface').Verifier<ConfigDID|ID,Alg>}
+ */
+export function configureVerifier(config, verifier) {
+  if (config.DID) {
+    return verifier.withDID(DID.parse(config.DID).did())
+  }
+  return verifier
 }
--- a/packages/access-api/src/config.js
+++ b/packages/access-api/src/config.js
@@ -66,6 +66,9 @@ export function loadConfig(env) {
     SPACES: env.SPACES,
     VALIDATIONS: env.VALIDATIONS,
     DB: /** @type {D1Database} */ (env.__D1_BETA__),
+
+    UPLOAD_API_URL: env.UPLOAD_API_URL,
+    UPLOAD_API_URL_STAGING: env.UPLOAD_API_URL_STAGING,
   }
 }
 
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -1,4 +1,4 @@
-import * as DID from '@ipld/dag-ucan/did'
+import * as ucanto from '@ucanto/core'
 import * as Server from '@ucanto/server'
 import { Failure } from '@ucanto/server'
 import * as Space from '@web3-storage/capabilities/space'
@@ -102,7 +102,7 @@ export function service(ctx) {
           const inv = await Space.recover
             .invoke({
               issuer: ctx.signer,
-              audience: DID.parse(capability.with),
+              audience: ucanto.DID.parse(capability.with),
               with: ctx.signer.did(),
               lifetimeInSeconds: 60 * 10,
               nb: {
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -9,7 +9,7 @@ import {
 } from '@web3-storage/access/encoding'
 import { voucherClaimProvider } from './voucher-claim.js'
 import { voucherRedeemProvider } from './voucher-redeem.js'
-import { UploadApiProxyService } from './upload-api.js'
+import { UploadApiProxyService } from './upload-api-proxy.js'
 
 /**
  * @param {import('../bindings').RouteContext} ctx
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -9,17 +9,20 @@ import {
 } from '@web3-storage/access/encoding'
 import { voucherClaimProvider } from './voucher-claim.js'
 import { voucherRedeemProvider } from './voucher-redeem.js'
-import { UploadApiProxyService } from './upload-api-proxy.js'
+import * as uploadApi from './upload-api-proxy.js'
 
 /**
  * @param {import('../bindings').RouteContext} ctx
- * @returns {import('@web3-storage/access/types').Service}
+ * @returns {
+ * & import('@web3-storage/access/types').Service
+ * & { store: uploadApi.StoreServiceInferred }
+ * & { upload: uploadApi.UploadServiceInferred }
+ * }
  */
 export function service(ctx) {
   return {
-    ...new UploadApiProxyService({
-      fetch: globalThis.fetch,
-    }),
+    store: uploadApi.createStoreProxy(ctx),
+    upload: uploadApi.createUploadProxy(ctx),
 
     voucher: {
       claim: voucherClaimProvider(ctx),
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -9,6 +9,7 @@ import {
 } from '@web3-storage/access/encoding'
 import { voucherClaimProvider } from './voucher-claim.js'
 import { voucherRedeemProvider } from './voucher-redeem.js'
+import { UploadApiProxyService } from './upload-api.js'
 
 /**
  * @param {import('../bindings').RouteContext} ctx
@@ -16,6 +17,10 @@ import { voucherRedeemProvider } from './voucher-redeem.js'
  */
 export function service(ctx) {
   return {
+    ...UploadApiProxyService.forSigner({
+      signer: ctx.signer,
+      fetch: globalThis.fetch,
+    }),
     voucher: {
       claim: voucherClaimProvider(ctx),
       redeem: voucherRedeemProvider(ctx),
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -17,10 +17,10 @@ import { UploadApiProxyService } from './upload-api-proxy.js'
  */
 export function service(ctx) {
   return {
-    ...UploadApiProxyService.forSigner({
-      signer: ctx.signer,
+    ...UploadApiProxyService.create({
       fetch: globalThis.fetch,
     }),
+
     voucher: {
       claim: voucherClaimProvider(ctx),
       redeem: voucherRedeemProvider(ctx),
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -17,7 +17,7 @@ import { UploadApiProxyService } from './upload-api-proxy.js'
  */
 export function service(ctx) {
   return {
-    ...UploadApiProxyService.create({
+    ...new UploadApiProxyService({
       fetch: globalThis.fetch,
     }),
 
--- a/packages/access-api/src/service/upload-api-proxy.js
+++ b/packages/access-api/src/service/upload-api-proxy.js
@@ -14,47 +14,17 @@ import * as Ucanto from '@ucanto/interface'
 import { createProxyHandler } from '../ucanto/proxy.js'
 
 /**
- * @template {Ucanto.Capability} C
- * @template [Success=unknown]
- * @template {{ error: true }} [Failure={error:true}]
- * @callback InvocationResponder
- * @param {Ucanto.Invocation<C>} invocationIn
- * @param {Ucanto.InvocationContext} context
- * @returns {Promise<Ucanto.Result<Success, Failure>>}
- */
-
-/**
- * Select from T the property names whose values are of type V
- *
- * @template T
- * @template V
- * @typedef { { [K in keyof T]-?: T[K] extends V ? K : never }[keyof T] } KeysWithValue
- */
-
-/**
- * Select from T the entries where the vlaue is of type V
- *
- * @template T
- * @template V
- * @typedef { { [K in KeysWithValue<T,V>]: T[K] } } OnlyValuesOfType
- */
-
-/**
- * @template {Record<string, unknown>} S
- * @typedef { { [K in KeysWithValue<S, Ucanto.TheCapabilityParser<any>>]: InvocationResponder<Ucanto.InferInvokedCapability<S[K] extends Ucanto.TheCapabilityParser<infer M> ? S[K] : never>> } } ModuleService
- */
-
-/**
- * @typedef {ModuleService<Omit<Store, 'store'>>} StoreServiceInferred
- * @typedef {ModuleService<Upload>} UploadServiceInferred
+ * @typedef {import('../ucanto/types.js').InferService<Omit<Store, 'store'>>} StoreServiceInferred
+ * @typedef {import('../ucanto/types.js').InferService<Upload>} UploadServiceInferred
  */
 
 /**
  * @template {string|number|symbol} M
+ * @template {Ucanto.ConnectionView<any>} [Connection=Ucanto.ConnectionView<any>]
  * @param {object} options
  * @param {Ucanto.Signer} [options.signer]
  * @param {Array<M>} options.methods
- * @param {Pick<Map<dagUcan.DID, Ucanto.ConnectionView<Record<M, any>>>, 'get'>} options.connections
+ * @param {{ default: Connection } & Record<Ucanto.UCAN.DID, Connection>} options.connections
  */
 function createProxyService(options) {
   const handleInvocation = createProxyHandler(options)
@@ -116,20 +86,6 @@ export const uploadApiAudienceToUrl = (() => {
   return object
 })()
 
-// /**
-//  * @param {object} options
-//  * @param {Ucanto.Signer} [options.signer]
-//  * @param {typeof globalThis.fetch} options.fetch
-//  * @param {Record<Ucanto.UCAN.DID, URL>} [options.audienceToUrl]
-//  * @returns
-//  */
-// export const createStoreService = (options) => {
-//   return createProxyService({
-//   connections:
-//   methods: ['list', 'add', 'remove', 'store'],
-// })
-// }
-
 export class UploadApiProxyService {
   /** @type {StoreServiceInferred} */
   store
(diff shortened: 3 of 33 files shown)


packages/access-api/test/store-list.js
```diff
@@ -0,0 +1,197 @@
...
+      service: serviceSigner,
+      conn,
+    } = await context({
+      environment: {
+        ...process.env,
+        // this emulates the configuration for deployed environments,
+        // which will allow the access-api ucanto server to accept
+        // invocations where aud=web3storageDid
+        DID: web3storageDid,
+        PRIVATE_KEY: privateKeyFromEnv ?? process.env.PRIVATE_KEY,
```
gobengo, line 79: some of this extra stuff is in here because I wanted to support running the tests with the production private key, which works with a non-error result. But if you don't pass that in, the assertions shift slightly to ensuring that it's an error result of the kind that would happen only if the server signer is wrong

Note how this test was run with `WEB3_STORAGE_PRIVATE_KEY`:
```
bengo@bengo ~/protocol.ai/w3protocol/packages/access-api ⚡  DID="$PROD_DID_KEY" WEB3_STORAGE_PRIVATE_KEY="$PROD_PRIVATE_KEY" pnpm test -- -f 'upload-api'

> @web3-storage/access-api@4.2.0 test /Users/bengo/protocol.ai/w3protocol/packages/access-api
> pnpm build && mocha --bail --timeout 10s -n no-warnings -n experimental-vm-modules "-f" "upload-api"


> @web3-storage/access-api@4.2.0 build /Users/bengo/protocol.ai/w3protocol/packages/access-api
> scripts/cli.js build



  proxy store/list invocations to upload-api
    ✔ forwards store/list invocations with aud=did:key (648ms)
    ✔ forwards invocations with aud=did:web:web3.storage (460ms)
    ✔ errors when a bad delegation is given as proof (320ms)

  Store.all
    ✔ proxies store/add to upload-api (331ms)
    ✔ proxies store/remove to upload-api (306ms)
    ✔ proxies store/list to upload-api (385ms)

  Upload.all
    ✔ proxies upload/add to upload-api (303ms)
    ✔ proxies upload/remove to upload-api (287ms)
    ✔ proxies upload/list to upload-api (342ms)


  9 passing (3s)

File: packages/access-api/src/service/upload-api-proxy.js [hunk 2]
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  /**
+   * @param {object} options
+   * @param {iucanto.Signer} [options.signer]
+   * @param {typeof globalThis.fetch} options.fetch
+   */
+  static create(options) {
+    // eslint-disable-next-line unicorn/no-array-reduce
+    const audienceToUrl = Object.values(this.#environments).reduce(
```
Gozala, line 220: Do we ever need to have multiple connections in practice ? I'd expect we use one in one staging and the other in production, if so it seems like there's bunch of unnecessary code to be maintained with added complexity costs.

As a side note we should probably add code coverage which would decentivize adding code which isn't used.

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,266 @@
...
+ * @typedef UcantoHttpConnectionOptions
+ * @property {dagUcan.DID} audience
+ * @property {typeof globalThis.fetch} options.fetch
+ * @property {URL} options.url
+ */
+
+/**
+ * @implements {Pick<Map<dagUcan.DID, iucanto.Connection<any>>, 'get'>}
+ */
+class AudienceConnections {
```
Gozala, line 137: Nit: This whole class seems like incidental complexity, things would be a whole lot simpler with just `config` object and static function for `get` method.

P.S.: I know that's not a popular opinion, yet I like to share it in a hope to bring peers to a true side :P

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+ * Return whether the provided stack trace string appears to be generated
+ * by a deployed upload-api.
+ * Heuristics:
+ * * stack trace files paths will start with `file:///var/task/upload-api` because of how the lambda environment is working
+ *
+ * @param {string} stack
+ */
+export function isUploadApiStack(stack) {
```
Gozala, line 264: Nit: Having this here makes me uncomfortable, seems to me that it belongs in tests so it's not used for other things.

packages/access-api/test/helpers/context.js
```diff
@@ -70,7 +73,7 @@ export async function context(options) {
   return {
     mf,
     conn: connection({
-      principal,
+      principal: servicePrincipal,
```
Gozala, line 76: Nit: Can you please delete `principal` created on line 53 then ?

packages/access-api/test/upload-api-proxy.test.js
```diff
@@ -0,0 +1,115 @@
...
+import * as Store from '@web3-storage/capabilities/store'
+import * as Upload from '@web3-storage/capabilities/upload'
+import * as dagUcanDid from '@ipld/dag-ucan/did'
+import { context } from './helpers/context.js'
+import * as ucanto from '@ucanto/core'
+// eslint-disable-next-line no-unused-vars
+import * as iucanto from '@ucanto/interface'
+import { isUploadApiStack } from '../src/service/upload-api-proxy.js'
+
+describe('parserCapabilities', function () {
```
Gozala, line 11: These tests seem redundant to me, they basically test that `parserAbilities` works by comparing to the list of abilities expected. It seems to me that just using a set of `expectedCans` in the next tests would be good enough and more practical.

Only thing we can catch with the whole `parserAbilities` dance is new capabilities added, yet there's no guarantee that `parserAbilities` will catch those. If do want to go extra mile and catch those cases, I would suggest using `can` field of capability parsers (they should have it although it's not part of interface), in other words you could derive expected cans like:

```ts
Object.values(Upload).map(cap => cap.can || '').filter(can => !can.includes('|'))
```

packages/access-api/test/store-list.js
```diff
@@ -0,0 +1,197 @@
...
+    )
+    const listInvocation = Store.list.invoke({
+      issuer,
+      audience: service,
+      proofs: [spaceCreation.delegation],
+      with: spaceCreation.space.did(),
+      nb: {},
+    })
+    const result = await listInvocation.execute(
+      // cast to `any` only because this `conn` uses Service type from access-client.
```
Gozala, line 97: Nit: We should probably change `access-client` as well so it's aware of `store` and ` upload` capabilities now. Can be a followup though.

Gozala (approved):
Added bunch of Nits, which you can completely ignore or address before or after landing. Only thing I do wish is that we had a test(s) where we spin up local `store` and `upload` services and route to those in test so that we have passing cases that don't require actual w3up service and all worked in offline mode.

That said it's probably tricky to do it without moving bunch of code from w3infra here so I think it's good enough for now.

Gozala:
> That said it's probably tricky to do it without moving bunch of code from w3infra here so I think it's good enough for now.

Created this issue to that effect https://github.com/web3-storage/w3infra/issues/127

gobengo:
cc @hugomrdias

packages/access-api/src/config.js
```diff
@@ -1,5 +1,7 @@
 import { Signer } from '@ucanto/principal/ed25519'
 import * as DID from '@ipld/dag-ucan/did'
+// eslint-disable-next-line no-unused-vars
+import * as dagUcan from '@ipld/dag-ucan'
```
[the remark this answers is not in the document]
gobengo: Thanks, I updated to that, and will prefer importing from ucanto libs in the future instead of dag-ucan.
```
import { UCAN } from '@ucanto/interface';
import { DID } from '@ucanto/core';
```

packages/access-api/src/service/index.js
```diff
@@ -9,13 +9,18 @@ import {
...
 import { voucherRedeemProvider } from './voucher-redeem.js'
+import { UploadApiProxyService } from './upload-api-proxy.js'
 
 /**
  * @param {import('../bindings').RouteContext} ctx
  * @returns {import('@web3-storage/access/types').Service}
  */
 export function service(ctx) {
   return {
+    ...UploadApiProxyService.create({
```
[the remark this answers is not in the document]
gobengo: i agree and will do as fast follow https://github.com/web3-storage/w3protocol/issues/338

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,266 @@
...
+// eslint-disable-next-line no-unused-vars
+import * as dagUcan from '@ipld/dag-ucan'
+import * as dagUcanDid from '@ipld/dag-ucan/did'
+import * as HTTP from '@ucanto/transport/http'
+// eslint-disable-next-line no-unused-vars
+import * as Store from '@web3-storage/capabilities/store'
+// eslint-disable-next-line no-unused-vars
+import * as Upload from '@web3-storage/capabilities/upload'
+// eslint-disable-next-line no-unused-vars
+import * as iucanto from '@ucanto/interface'
```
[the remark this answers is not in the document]
gobengo: I agree consistency is good and this name is meh (in at least one file I had `import * as ucanto from "@ucanto/core'` and `import * as iucanto from "@ucanto/interface"`).

`API` makes sense to me when I'm readying `@ucanto/` libs. Here I might assume it was a w3protocol API. Also I noticed that other files in this monorepo [call it `Ucanto`](https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/src/utils/d1.js#L3), which makes sense to me and I will update to that for consistency within w3protocol

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,266 @@
...
+ * @param {Pick<Map<dagUcan.DID, iucanto.ConnectionView<any>>, 'get'>} options.connections
+ * @param {iucanto.Signer} [options.signer]
+ * @returns {InvocationResponder<C, Success, Failure>}
+ */
+function createInvocationResponder(options) {
+  /**
+   * @template {import('@ucanto/interface').Capability} Capability
+   * @param {iucanto.Invocation<Capability>} invocationIn
+   * @param {iucanto.InvocationContext} context
+   * @returns {Promise<iucanto.Result<any, { error: true }>>}
```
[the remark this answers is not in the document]
gobengo: great find. I agree on avoiding `any`.

I can't remove this `@returns` while also appeasing the `@returns {InvocationResponder<C, Success, Failure>}` on `createInvocationResponder` itself. Because that's generic, it's probalby fine/useful to have a generic `@returns` right?

I will update it so this inner `@returns` is `Promise<Ucanto.Result<Success, Failure>>`, which works and is much more generic.

if you prefer to remove the (now generic) `@returns` alltogether, say so and I'll do it here, or feel free to just remove it as fast follow.

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,266 @@
...
+    }
+    // eslint-disable-next-line unicorn/prefer-logical-operator-over-ternary
+    const proxyInvocationIssuer = options.signer
+      ? // this results in a forwarded invocation, but the upstream will reject the signature
+        // created using options.signer unless options.signer signs w/ the same private key as the original issuer
+        // and it'd be nice to not even have to pass around `options.signer`
+        options.signer
+      : // this works, but involves lying about the issuer type (it wants a Signer but context.id is only a Verifier)
+        // @Gozala can we make it so `iucanto.InvocationOptions['issuer']` can be a Verifier and not just Signer?
+        /** @type {ed25519.Signer.Signer} */ (context.id)
```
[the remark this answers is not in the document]
gobengo: you're right. `Ucanto.Signer` works and is more generic.

packages/access-api/test/helpers/context.js
gobengo: It's used in the return value [on line 82](https://github.com/web3-storage/w3protocol/pull/334/files#diff-dd020d0285601ca47fabbd295beaa017b1415699c44343d1c183ef9f849341c8R82).

Before I touched this, it would use the same generated signer both for the connection servicePrincipal as well as the returned `issuer`. afaict that would mean most tests invocations before had identical `iss` and `aud`, which isn't necessarily wrong, but seems rare.

now it uses both, but it **does** use both so I believe I should not delete the var called `principal`, unless I'm missing something and you think I should return a different value for `context.issuer`

packages/access-api/test/store-list.js
gobengo: I thought about that but wasn't sure since I haven't hacked much in `access-client` yet. Let's do as followup as-needed

packages/access-api/test/upload-api-proxy.test.js
gobengo: That approach kind of works, but I found it hard to make into a well-typed reusable function (replace `Upload` with a parameter of what type?).

I removed the tests though, which aren't adding much.

packages/access-api/src/config.js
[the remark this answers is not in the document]
gobengo: https://github.com/web3-storage/w3protocol/pull/334/commits/fcc80ea334dd9aaacb94c454caa7351543a5dc6e
https://github.com/web3-storage/w3protocol/pull/334/commits/9559d0e13445a96be3009a641f35658eb3fed0e1

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,266 @@
+import * as Client from '@ucanto/client'
+import * as CAR from '@ucanto/transport/car'
+import * as CBOR from '@ucanto/transport/cbor'
+// eslint-disable-next-line no-unused-vars
+import * as dagUcan from '@ipld/dag-ucan'
+import * as dagUcanDid from '@ipld/dag-ucan/did'
```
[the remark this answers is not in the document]
gobengo: https://github.com/web3-storage/w3protocol/pull/334/commits/9559d0e13445a96be3009a641f35658eb3fed0e1#diff-969dad085be9e4a6140a50e5f0142a0e6f106b12a5c3375f5857006fbe71dab8

packages/access-api/src/service/upload-api-proxy.js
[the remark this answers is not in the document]
gobengo: https://github.com/web3-storage/w3protocol/pull/334/commits/9559d0e13445a96be3009a641f35658eb3fed0e1

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,266 @@
...
+      )
+    }
+    // eslint-disable-next-line unicorn/prefer-logical-operator-over-ternary
+    const proxyInvocationIssuer = options.signer
+      ? // this results in a forwarded invocation, but the upstream will reject the signature
+        // created using options.signer unless options.signer signs w/ the same private key as the original issuer
+        // and it'd be nice to not even have to pass around `options.signer`
+        options.signer
+      : // this works, but involves lying about the issuer type (it wants a Signer but context.id is only a Verifier)
+        // @Gozala can we make it so `iucanto.InvocationOptions['issuer']` can be a Verifier and not just Signer?
```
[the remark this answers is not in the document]
gobengo: https://github.com/web3-storage/w3protocol/pull/334/commits/b9c26441cacb88b8533fbea5f70bc7e75d9af5bd

packages/access-api/src/service/upload-api-proxy.js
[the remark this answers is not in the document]
gobengo: https://github.com/web3-storage/w3protocol/pull/334/commits/b9a3ed8496639920e98a3f0da3f2fbba0c8c7404

packages/access-api/src/service/upload-api-proxy.js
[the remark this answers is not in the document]
gobengo: https://github.com/web3-storage/w3protocol/pull/334/commits/b9a3ed8496639920e98a3f0da3f2fbba0c8c7404#diff-969dad085be9e4a6140a50e5f0142a0e6f106b12a5c3375f5857006fbe71dab8R53

packages/access-api/src/service/upload-api-proxy.js
gobengo: > Do we ever need to have multiple connections in practice ?

No. Though we could also ask "Do we ever need a unified access-api + upload-api single environment service in practice?". Not yet, but I think we both believe it will make working with the total API more intuitive (i.e. api user won't have to worry about which service to invoke).

When we talked I like your vision of 'If you send an invocation to any API, it should try to route it to a place that can handle it'. i.e. unifying api across services. Doing it this way also unifies across environments. I don't think it's strictly necessary, but I do believe it will make working with the overall platform slightly more intuitive. (i.e. api user won't have to worry about **which environment's** 'unified' api to invoke).

I definitely agree code coverage could be useful. https://github.com/web3-storage/w3protocol/issues/339#issue-1531511259

packages/access-api/src/service/upload-api-proxy.js
gobengo: Is this like what you had in mind? https://github.com/web3-storage/w3protocol/pull/334/commits/b46dd85598c427e2a5b0e86fba2f9cb76342f336

> P.S.: I know that's not a popular opinion, yet I like to share it in a hope to bring peers to a true side :P

Eh, I don't disagree necessarily. good suggestion.

packages/access-api/src/service/upload-api-proxy.js
gobengo: https://github.com/web3-storage/w3protocol/pull/334/commits/49144adb8f52904af020a3b3db096b42d69e3f65

packages/access-api/test/upload-api-proxy.test.js
gobengo: https://github.com/web3-storage/w3protocol/pull/334/commits/51794b0a3ffdb6a544f2f4b66c431ecb5aec5126

packages/access-api/test/store-list.js
Gozala: https://github.com/web3-storage/w3protocol/issues/340

packages/access-api/src/service/upload-api-proxy.js
Gozala: Not exactly, sorry for not been very clear. In linked version you still have `audienceConnections` which has `get` function. When I criticize classes what I mean is:

Separate data part which is basically a state from the behavior which are all the methods

That way it does not matter how you constructed state, you can just pass it into a function and it will compute whatever. Separating data and behavior de-tangles two allowing you to store data in e.g. indexDB and then apply function after reading it from indexDB without having to built a class first. Not the case here but complex class heirerchies also often entangle initialization order.

I think there are only two valid reasons to resort to a classes with methods:

1. When you have to deal with open-ended variant space. In other sometimes you have many different state options but only one of them will be used in practice. In this scenario dispatching from one function introduces unnecessary coordination and defining interface is better option.
2. When your methods are simply a convenience sugar for the static functions. I do it all the time to allow method chaining and remove need for importing lot of dependencies. In this case they don't tangle anything they are nothing but a sugar


I would argue use of classes with methods for any other reason just introduces incidental complexity and gets between you and data nothing else.

packages/access-api/test/helpers/context.js
Gozala: Thanks for clarification I did miss it on line 82 and yeah I would expect `iss === aud` to be an unrealistic scenario so having them differ is good 👍

packages/access-api/test/upload-api-proxy.test.js
Gozala: > That approach kind of works, but I found it hard to make into a well-typed reusable function (replace Upload with a parameter of what type?).

Yeah because `can` isn't in types.

I'm mostly concerned that some assumbtions `parserAbilities` makes may not stand long term. It's fine though, we can deal with that when it happens.

packages/access-api/src/service/upload-api-proxy.js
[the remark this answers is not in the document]
Gozala: > API makes sense to me when I'm readying @ucanto/ libs. Here I might assume it was a w3protocol API.

Fare enough, in those cases I often have `type.ts` in the lib that just does `export * from  "@ucanto/interface"` and whatever other things it depends on. Sometimes namespacing if there are conflicts

packages/access-api/src/service/upload-api-proxy.js
gobengo: @Gozala ah I understand what you mean now.

> Separating data and behavior de-tangles two allowing you to store data in e.g. indexDB and then apply function after reading it from indexDB without having to built a class first.

def have run into that. That helped me understand.

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * @protected
+   * @param {StoreService} store
+   * @param {UploadService} upload
+   */
+  constructor(store, upload) {
+    this.store = store
+    this.upload = upload
+  }
+}
```
hugomrdias, line 188: having an hard time following this

cant we just to the following ? all the types should match and its easier to understand
```js

/**
 * @param {object} options
 * @param {typeof globalThis.fetch} options.fetch
 */
function createUploadProxy(options) {
  /** @type {Record<Ucanto.DID, Ucanto.ConnectionView<import('@web3-storage/access/types').Service>>} */
  const connections = {
      'did:web:web3.storage': createUcantoHttpConnection({
        audience: 'did:web:web3.storage',
        url: new URL('https://up.web3.storage'),
        fetch: options.fetch
      }),
     'did:web:staging.web3.storage': createUcantoHttpConnection({
        audience: 'did:web:staging.web3.storage',
        url: new URL('https://staging.up.web3.storage'),
        fetch: options.fetch
      })
    }
  

  return {
    store: {
      add: Server.provide(Store.add, async ({capability, invocation}) => {
        const conn = connections[invocation.audience.did()]
        return conn.execute(invocation)
      })
    },
    .... // all the other handlers
  }
}
```

hugomrdias (changes requested):

packages/access-api/src/config.js
```diff
@@ -128,3 +130,19 @@ export function configureSigner(config) {
   }
   return signer
 }
+
+/**
+ * @template {UCAN.DID} ConfigDID
+ * @template {UCAN.DID} [ID=UCAN.DID]
```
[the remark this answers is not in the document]
gobengo: There is none. Great catch :) ty

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,188 @@
...
+ * @property {InvocationResponder<Ucanto.InferInvokedCapability<typeof Store.list>>} list
+ * @property {InvocationResponder<Ucanto.InferInvokedCapability<typeof Store.remove>>} remove
+ */
+
+/**
+ * @typedef UploadService
+ * @property {InvocationResponder<Ucanto.InferInvokedCapability<typeof Upload.add>>} add
+ * @property {InvocationResponder<Ucanto.InferInvokedCapability<typeof Upload.list>>} list
+ * @property {InvocationResponder<Ucanto.InferInvokedCapability<typeof Upload.remove>>} remove
+ */
```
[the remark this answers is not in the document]
gobengo: Thanks. I found a way of inferring from the Service import itself. https://github.com/web3-storage/w3protocol/pull/334/commits/08981d82feddc90c7ad8c3026525aaa09298451f#diff-969dad085be9e4a6140a50e5f0142a0e6f106b12a5c3375f5857006fbe71dab8R17

packages/access-api/src/config.js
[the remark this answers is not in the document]
gobengo: https://github.com/web3-storage/w3protocol/pull/334/commits/8b03b1bdfc93d1c3b7996c91bc2d298208a07c4b

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * @template {Ucanto.Capability} C
+ * @template [Success=unknown]
+ * @template {{ error: true }} [Failure={error:true}]
+ * @callback InvocationResponder
+ * @param {Ucanto.Invocation<C>} invocationIn
+ * @param {Ucanto.InvocationContext} context
+ * @returns {Promise<Ucanto.Result<Success, Failure>>}
+ */
```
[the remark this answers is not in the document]
gobengo: The only difference is the defaults on Success and Failure. I don't need those so I removed this. Good find. I didn't know all the types we had at our disposal but after this PR I've definitely gotten more familiar with them.
https://github.com/web3-storage/w3protocol/pull/334/commits/fa1245effa62da4c2a129b231b8e5fcf88008227

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,188 @@
...
+ * @property {InvocationResponder<Ucanto.InferInvokedCapability<typeof Upload.add>>} add
+ * @property {InvocationResponder<Ucanto.InferInvokedCapability<typeof Upload.list>>} list
+ * @property {InvocationResponder<Ucanto.InferInvokedCapability<typeof Upload.remove>>} remove
+ */
+
+/**
+ * @template {Record<string, any>} T
+ * @param {object} options
+ * @param {Ucanto.Signer} [options.signer]
+ * @param {Pick<Map<dagUcan.DID, Ucanto.ConnectionView<T>>, 'get'>} options.connections
```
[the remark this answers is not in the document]
gobengo: it was `{ get: (audience: Ucanto.UCAN.DID) => ConnectionView<T> }`

I got rid of this in favor of a record/obj with no functions (which iirc Irakli had also preferred). https://github.com/web3-storage/w3protocol/pull/334/commits/08981d82feddc90c7ad8c3026525aaa09298451f#diff-5fb7be1f94b33171518ab23f095f24d6e1e025282c99e8131c58e9db6c2caa4eL17

packages/access-api/src/service/upload-api-proxy.js
gobengo: I iterated towards this in two ways:
* `connections` is now just an object like you illustrate here no `get` method
* unbundling the class with store + upload into separate functions that build each

I don't believe that the way you're doing `conn.execute(invocation)` will actually work. I wish it did. I tried that first. but had to try some various things to avoid making ucanto changes.

@Gozala and I discussed a few ways that ucanto could make this easier, but also decided to do something that works with ucanto as is, ship that, and decde after that whether to push some of those tools (that are now in `access-api/src/ucanto/proxy.js`) down into ucanto itself.

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -0,0 +1,188 @@
...
+import * as CAR from '@ucanto/transport/car'
+import * as CBOR from '@ucanto/transport/cbor'
+// eslint-disable-next-line no-unused-vars
+import * as dagUcan from '@ipld/dag-ucan'
+import { DID } from '@ucanto/core'
+import * as HTTP from '@ucanto/transport/http'
+// eslint-disable-next-line no-unused-vars
+import * as Store from '@web3-storage/capabilities/store'
+// eslint-disable-next-line no-unused-vars
+import * as Upload from '@web3-storage/capabilities/upload'
```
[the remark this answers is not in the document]
gobengo: i got rid of them and will avoid it in the future too

packages/access-api/src/service/upload-api-proxy.js
hugomrdias: why does it not work ? this was the promise of "lets make a single endpoint because ucanto makes this super simple" this is not super simple

hugomrdias:
i was hoping for another round of review

PR #337, opened 2023-01-12
chore: apply most of recommended eslintConfig to package.json

natevw:
The README currently instructs developers to manually apply the `eslintConfig` to a locally-modified package.json. That seems a bit of a headache to then leave out of every subsequent commit and it's unclear the advantage of not just applying it in the repo?

In this commit I've applied what is labelled as the **non-optional** parts to simply be in the package.json without needing further intervention.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,17 @@
     "typescript": "4.9.4",
     "wrangler": "^2.6.2"
   },
+  "eslintConfig": {
+    "extends": [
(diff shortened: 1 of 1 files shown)


hugomrdias:
that for any new packages created in this repo, the root package.json should not have this configuration

natevw:
> that is for any new packages created in this repo, the root package.json should not have this configuration

ah okay! then let's close this PR, right?

PR #354, opened 2023-01-17
feat: optional override for Postmark email From: field

natevw:
This change allows an optional `POSTMARK_SENDER` env variable to be set alongside the existing `POSTMARK_TOKEN` in case the configure token's account is not approved for sending from the default (main/real) domain. Without this the "From: …" sender email address is hardcoded at quite a low-level which makes it difficult for code re-use.

With this change, I'm enabled to set up a local environment and test from my *own* Postmark account without needing production keys. (Since I don't own or control the domain, sending an email from `anything@dag.house` gets rejected by the Postmark server with an error if requested via my own API token.)

I have avoided touching most of the many places where environment variables are handled because most of those deal in terms of **required** env. This new config variable is optional: available for those who need it, but if not set, the email logic simply stays defaulted to what was the hardcoded sender.


alanshaw (approved):

alanshaw:
@natevw would you mind fixing the lint error reported by CI 🙏?

alanshaw:
> Secret `CF_TOKEN` is required, but not provided while calling.

Looks like we need to update the cloudflare token?

PR #357, opened 2023-01-18
changes to how upload-api proxying works

gobengo:
* route `aud=did:web:staging.web3.storage` invocations to https://staging.up.web3.storage even if `env.UPLOAD_API_URL_STAGING` isn't set
  * This will make it easier to iterate and get this working against staging upload-api without having to use production secrets

### DIFF
--- a/packages/access-api/src/service/upload-api-proxy.js
+++ b/packages/access-api/src/service/upload-api-proxy.js
@@ -84,12 +84,10 @@ function getDefaultConnections(options) {
       ...(uploadApi.production && { url: uploadApi.production }),
       fetch,
     }),
-    ...(uploadApi.staging && {
-      [uploadApiEnvironments.staging.audience]: createUcantoHttpConnection({
-        ...uploadApiEnvironments.staging,
-        url: uploadApi.staging,
-        fetch,
-      }),
+    [uploadApiEnvironments.staging.audience]: createUcantoHttpConnection({
+      ...uploadApiEnvironments.staging,
+      url: uploadApi.staging ?? uploadApiEnvironments.staging.url,
+      fetch,
     }),
   }
 }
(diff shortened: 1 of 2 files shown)


packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -84,12 +84,10 @@ function getDefaultConnections(options) {
       ...(uploadApi.production && { url: uploadApi.production }),
       fetch,
     }),
-    ...(uploadApi.staging && {
-      [uploadApiEnvironments.staging.audience]: createUcantoHttpConnection({
-        ...uploadApiEnvironments.staging,
-        url: uploadApi.staging,
-        fetch,
-      }),
+    [uploadApiEnvironments.staging.audience]: createUcantoHttpConnection({
```
Gozala, line 87: would not this fail if `uploadApi.staging` is falsy ?

packages/access-api/src/service/upload-api-proxy.js
```diff
@@ -84,12 +84,10 @@ function getDefaultConnections(options) {
...
+      ...uploadApiEnvironments.staging,
+      url: uploadApi.staging ?? uploadApiEnvironments.staging.url,
```
Gozala, line 89:
```suggestion
      url: uploadApi.staging ?? uploadApiEnvironments.staging.url,
```

Dependency between two variable here is confusing, should they both be `uploadApi` or `uploadApiEnvironments` instead ?

gobengo: `uploadApi.staging` is the url comes from options and might be falsy. `uploadApiEnvironments.staging.url` will always be defined and is used as the default.

Gozala (approved):

packages/access-api/src/service/upload-api-proxy.js
gobengo: I don't think it will, because in that case it will default to using values from `uploadApiEnvironments.staging` (which is always defined in this file)

PR #362, opened 2023-01-18
chore: update deps and fix problems

hugomrdias:
- updates prettier and hd-scripts deps and fixes all the new errors created by the update


hugomrdias:
Its passing now can you review again pls?

vasco-santos (approved):
LGTM

Issue #363, opened 2023-01-18
access-api: invoking store/list in production errors with `HandlerExecutionError` + `HTTPError`


gobengo:
Motivation:
* testing #325 

Context:
* after #361 , my w3protocol-test can successfully send `can=store/list aud=did:web:staging.web3.storage` invocations through access-api staging
* but it can't do the same against prod 

Problem
<ul>
<li>
Error message <a href="https://github.com/gobengo/monorepo/actions/runs/3952405802/jobs/6767435499#step:6:22">seen in github actions run</a>
<details>
<pre>
# testing aud=did:web:web3.storage url=https://up.web3.storage/
# testing aud=did:web:web3.storage url=https://access.web3.storage/
# unexpected result from store/list invocation aud=did:web:web3.storage url=https://access.web3.storage/ {
#   name: 'HandlerExecutionError',
#   cause: {
#     url: 'https://up.web3.storage/',
#     name: 'HTTPError',
#     stack: 'HTTPError: HTTP Request failed\\n' +
#       '    at lN.throw (d1-beta-facade.entry.js:14585:11)\\n' +
#       '    at dN.request (d1-beta-facade.entry.js:14579:135)\\n' +
#       '    at async WN (d1-beta-facade.entry.js:15030:45)\\n' +
#       '    at async Object.<anonymous> (d1-beta-facade.entry.js:15037:132)\\n' +
#       '    at async IB (d1-beta-facade.entry.js:14077:12)\\n' +
#       '    at async AB (d1-beta-facade.entry.js:14065:12)\\n' +
#       '    at async TB (d1-beta-facade.entry.js:14059:42)\\n' +
#       '    at async PO (d1-beta-facade.entry.js:15097:11)\\n' +
#       '    at async ed.fetch (d1-beta-facade.entry.js:9434:17)\\n' +
#       '    at async Object.fetch (d1-beta-facade.entry.js:20183:13)',
#     status: 530,
#     message: 'HTTP Request failed',
#     statusText: ''
#   },
#   error: true,
#   stack: 'HandlerExecutionError: service handler {can: "store/list"} error: HTTP Request failed\\n' +
#     '    at IB (d1-beta-facade.entry.js:14079:13)\\n' +
#     '    at async AB (d1-beta-facade.entry.js:14065:12)\\n' +
#     '    at async TB (d1-beta-facade.entry.js:14059:42)\\n' +
#     '    at async PO (d1-beta-facade.entry.js:15097:11)\\n' +
[...]

PR #364, opened 2023-01-18
feat: access-api createStoreProxy uses patched fetch that will log any non-ok responses (to debug)

gobengo:
…that logs unexpected responses with status=530

Motivation:
* gather more info for [this error happening in production](https://github.com/web3-storage/w3protocol/issues/363#issuecomment-1396114241)

Plan
* gather info, then revert this PR


yusefnapora (approved):
This seems like a good (temporary) hack, given that the error only seems to happen in prod. Hopefully comparing the output between prod & staging will flush out something...

PR #375, opened 2023-01-19
chore: add examples & context to capabilities package readme

### LINKED ISSUE #369: docs: add examples & context to capabilities readme
I want to update the capabilities package readme to:

- [x] include examples of calling `invoke` and `delegate` on the exported capability objects.
- [x] explain the inferred concrete types (`TheCapabilityParser`, etc) and what's meaningful to an end user
- [x] link to the capabilities.md doc for more info

yusefnapora:
This closes #369 by adding some examples of using the capability objects exposed by the capabilities package and explaining what some of the fancy type signatures are for.

### DIFF
--- a/packages/capabilities/readme.md
+++ b/packages/capabilities/readme.md
@@ -92,7 +92,7 @@ The above would result in an object similar to the following:
 
 #### `invoke`
 
-The `invoke` method returns an "invocation" of the capability, which can be executed against a ucanto service.
+The `invoke` method returns an [invocation](https://github.com/ucan-wg/spec/#29-invocation) of the capability, which can be executed against a ucanto service.
 
 Like `create`, `invoke` accepts `with` and `nb` fields, and the inputs must match the constraints in the capability definition.
 
--- a/packages/capabilities/readme.md
+++ b/packages/capabilities/readme.md
@@ -112,12 +112,13 @@ interface UCANOptions {
 }
 ```
 
-In the example below, we assume that the `loadSigningKey` function returns a valid `Signer` object.
+In the example below, we're generating a new `Signer` to act as the issuer of the invocation using the `@ucanto/principal/ed25519` package. Note that in a real application, the service would likely reject an invocation from this signer, as it does not have any delegated permissions. See the [access client package](https://github.com/web3-storage/w3protocol/tree/main/packages/access-client) for more about key management and delegation in practice.
 
 ```ts
 import * as DID from '@ipld/dag-ucan/did'
+import * as ed25519 from '@ucanto/principal/ed25519'
 
-const issuer = loadSigningKey()
+const issuer = await ed25519.generate()
 const audience = DID.parse('did:web:web3.storage')
 
 const invocation = Store.add.invoke({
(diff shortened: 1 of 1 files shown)


packages/capabilities/readme.md
```diff
@@ -25,3 +33,126 @@ import * as Voucher from '@web3-storage/capabilities/voucher'
...
+        origin: Store.Schema.Schema<...>;
+    }>;
+}, CapabilityMatch<...> | DerivedMatch<...>>>
+```
+
+While this is a fairly complex type signature, most of the types exist to support the mechanics of the capability parser and can generally be ignored when using the capabilities. The most interesting part as a user is the definition in the `DerivedMatch` type constraint, which shows the inferred ability and the constraints upon the resource URI and the caveats. In the example above, the `can` field shows that this capability's ability is `"store/add"`, its resource URI (the `with` field) must have the `"did:"` scheme, and there are three caveats defined in the `nb` field: `link`, `size`, and `origin`, each of which have constraints on their allowed values.
+
+### Using the exported capabilities
+
+The capability object exposes three methods via the `TheCapabilityParser` interface: `create`, `invoke`, and `delegate`.
```
gobengo, line 62: this could be good docs to have in ucanto itself (too) https://github.com/web3-storage/ucanto/

packages/capabilities/readme.md
```diff
@@ -25,3 +33,126 @@ import * as Voucher from '@web3-storage/capabilities/voucher'
...
+  nb: {
+    link: 'bagbaieraspawtgooy5lptr7loyd3fxjsrgkamre3y6au3ga4df5bkhrxdkmq',
+    size: 20,
+  }
+}
+```
+
+#### `invoke`
+
+The `invoke` method returns an "invocation" of the capability, which can be executed against a ucanto service.
```
gobengo, line 95:
```suggestion
The `invoke` method returns an [invocation](https://github.com/ucan-wg/spec/#29-invocation) of the capability, which can be executed against a ucanto service.
```

packages/capabilities/readme.md
```diff
@@ -25,3 +33,126 @@ import * as Voucher from '@web3-storage/capabilities/voucher'
...
+  proofs?: Proof[]
+}
+```
+
+In the example below, we assume that the `loadSigningKey` function returns a valid `Signer` object.
+
+```ts
+import * as DID from '@ipld/dag-ucan/did'
+
+const issuer = loadSigningKey()
```
gobengo, line 120: IMO it could be illustrative to show a working version of this
```
import * as ed25519 from "@ucanto/principal/ed25519";
const issuer = await ed25519.generate();
```

packages/capabilities/readme.md
```diff
@@ -25,3 +33,126 @@ import * as Voucher from '@web3-storage/capabilities/voucher'
...
+
+```ts
+const result = await invocation.execute(serviceConnection)
+```
+
+#### `delegate`
+
+The `delegate` method allows you to create a ucanto `Delegation`, which allows another principal to invoke the capability.
+
+`delegate` accepts the same input as `invoke`, however the `nb` field is optional. If `nb` is present, the values provided will act as constraints on the invocations that can be made using the delegation. For example, creating a `store/add` delegation with the `size` caveat set to `1048576` would limit invocations made using the delegation to uploads of no more than 1MiB.
```
gobengo, line 144: > For example, creating a `store/add` delegation with the `size` caveat set to `1048576` would limit invocations made using the delegation to uploads of no more than 1MiB.

Love calling out the use cases like this. I think over time it may be worth having explicit examples of how to include that `size` caveat using the `delegate` method.

Another use case I anticipate is "How can I delegate the ability to store a specific CID"?

gobengo (approved):

hugomrdias (approved):
LGTM awesome explanations 😍

PR #380, opened 2023-01-20
feat: add /reproduce-cloudflare-error route to access-api

gobengo:
Motivation:
* cloudflare support asked for a URL to reproduce the undelrying cause of #363 
* this adds a route just for that. once they use it to reproduce and diagnose, we remove it


hugomrdias (approved):
LGTM 
we could change the worker from using custom domain binding to the old url pattern binding to check if that fixes the problem.

PR #387, opened 2023-01-23
feat(capabilities): implement access/authorize and ./update caps

### LINKED ISSUE #385: Implement access/authorize cap
Spec: https://github.com/web3-storage/specs/blob/main/w3-account.md#accessauthorize

- `access/authorize` adds a row in DELEGATIONS with `./update`
- no sql queries needed to handle
- sends email to the corresponding `did:mailto` requesting authorization to link that email address to the invocations iss `did:key`

hugomrdias:
closes #385

### DIFF
--- a/packages/capabilities/src/access.js
+++ b/packages/capabilities/src/access.js
@@ -1,0 +1,71 @@
+/**
+ * Access Capabilities
+ *
+ * These can be imported directly with:
+ * ```js
+ * import * as Access from '@web3-storage/capabilities/access'
+ * ```
+ *
+ * @module
+ */
+import { capability, URI, DID } from '@ucanto/validator'
+// @ts-ignore
+// eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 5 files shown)


Gozala (approved):
Looks good to me! 

Fact that `./update` is in `access/*` made me bit uncomfortable, don't know why, perhaps because it feels less special. That said I don't have any better ideas where to put it so this seems fine to me.

packages/capabilities/src/access.js
```diff
@@ -0,0 +1,102 @@
...
+    }],
+    exp: null
+    sig: "..."
+  }
+ * ```
+ */
+export const session = capability({
+  can: './update',
+  // Should be web3.storage DID
+  with: URI.match({ protocol: 'did:' }),
```
Gozala, line 97: Should this be `DID.match({ method: 'web' })` instead or perhaps even a `Schema.literal('did:web:web3.storage')` ?

PR #390, opened 2023-01-23
feat: update access-api ucanto proxy to not need a signer

gobengo:
… and not sign proxyInvocation. using features coming in ucanto 4.2.0

Motivation:
* https://github.com/web3-storage/w3protocol/issues/325
* simplify access-api ucanto proxy using features added to ucanto in https://github.com/web3-storage/ucanto/pull/199
  * previously, the technique used to proxy the invocation was to issue a new invocation (i.e. `proxyInvocation`) in the proxy server, and then send that to the upstream. This had at least two limitations:
    1. required the proxy server to be configured with a `options.signer` to sign the `proxyInvocation`
    2. for functional use in access-api and proxying upload-api, this proxy `options.signer` also had to be configured pretty much identically to the ucanto verifier with same did on the upstream, including requiring both to have the same private key
  * now
    * you don't need an `options.signer` at all! so you definitely don't need one creating signatures with the same private key as the upstream

Steps
* [x] release ucanto 4.2.0 https://github.com/web3-storage/ucanto/pull/200
* [x] update this source branch package.json + pnpm locks to upgrade ucanto to 4.2.0
* [x] ensure `tsc` + tests pass here


gobengo:
note: this is a superset of https://github.com/web3-storage/w3protocol/pull/405 , which only does the upgrade to ucanto 4.2.3 that this depends on.

packages/access-api/src/ucanto/proxy.js
```diff
@@ -47,35 +47,23 @@ function defaultCatchInvocationError(error) {
...
+    const { connections, catchInvocationError = defaultCatchInvocationError } =
+      options
+    const connection =
+      connections[invocation.audience.did()] ?? connections.default
     try {
       const [result] = await Client.execute(
-        [proxyInvocation],
-        /** @type {Client.ConnectionView<any>} */ (connection)
+        [await invocation.delegate()],
+        connection
```
Gozala, line 66: I would expect you to be able to just pass invocation without having to call .delegate()

Gozala (approved):

packages/access-api/src/ucanto/proxy.js
gobengo: This was required, but mostly because of the types of how `import('./types.js').ClientCodec` is defined (using other types from `@ucanto` that got updated the new release).

hugomrdias (approved):
LGTM can we remove the connections code in a follow up PR ? we should just need one single connection to upload api for these proxied invocations right ?

gobengo:
> can we remove the connections code in a follow up PR ? we should just need one single connection to upload api for these proxied invocations right ?

@hugomrdias Yes. I made this to echo back what I think you're asking for. https://github.com/web3-storage/w3protocol/issues/406
Let's align over there

Issue #382, opened 2023-01-20
access-api should respond to `space/info` invocations for space dids not in its db with a well-defined error result type


gobengo:
Motivation:
* enable clients to better distinguish when a space is not found by `space/info` handler. Right now they would have to parse an error message. With this they can just look at `result.status === 404` to check for this case.
* enable upload-api to distinguish when this happens https://github.com/web3-storage/w3infra/issues/134 (upload-api calls `space/info` inside its `verifyInvocation` helper)

Unblocks
* part of implementation of https://github.com/web3-storage/w3infra/issues/134

PR #391, opened 2023-01-24
feat: access-api handling store/info for space not in db returns failure with name

gobengo:
Motivation:
* implement https://github.com/web3-storage/w3protocol/issues/382
  * tl;dr `space/info` now has explicit failure type for case where the ucans are valid, but the space DID is not in the db. that way when upload-api invokes `space/info` as part of its `verifyInvocation` logic, it can distinguish between an invalid invocation and a not-found space id

Unblocks:
* this test passing https://github.com/gobengo/w3protocol-test/pull/1
* rest of https://github.com/web3-storage/w3infra/issues/134#user-content-implementation

### DIFF
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -33,11 +33,13 @@ export function service(ctx) {
       info: Server.provide(Space.info, async ({ capability, invocation }) => {
         const results = await ctx.models.spaces.get(capability.with)
         if (!results) {
-          return {
+          /** @type {import('@web3-storage/access/types').SpaceUnknown} */
+          const spaceUnknownFailure = {
             error: true,
-            status: 404,
-            message: 'Space not found.',
+            name: 'SpaceUnknown',
+            message: `Space not found.`,
           }
+          return spaceUnknownFailure
         }
         return results
       }),
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -33,7 +33,11 @@ export function service(ctx) {
       info: Server.provide(Space.info, async ({ capability, invocation }) => {
         const results = await ctx.models.spaces.get(capability.with)
         if (!results) {
-          return new Failure('Space not found.')
+          return {
+            error: true,
+            status: 404,
+            message: 'Space not found.',
+          }
         }
         return results
       }),
--- a/packages/access-client/src/errors.ts
+++ b/packages/access-client/src/errors.ts
@@ -1,0 +1,10 @@
+/**
+ * Indicates failure executing ability that requires access to a space that is not well-known enough to be handled.
+ * e.g. it's a space that's never been seen before,
+ * or it's a seen space that hasn't been fully registered such that the service can serve info about the space.
+ */
+export interface SpaceUnknown {
+  error: true
+  message: string
+  name: 'SpaceUnknown'
+}
--- a/packages/access-client/src/errors.ts
+++ b/packages/access-client/src/errors.ts
@@ -1,9 +1,11 @@
+import * as Ucanto from '@ucanto/interface'
+
 /**
  * Indicates failure executing ability that requires access to a space that is not well-known enough to be handled.
  * e.g. it's a space that's never been seen before,
  * or it's a seen space that hasn't been fully registered such that the service can serve info about the space.
  */
-export interface SpaceUnknown {
+export interface SpaceUnknown extends Ucanto.Failure {
   error: true
   message: string
   name: 'SpaceUnknown'
--- a/packages/access-client/src/types.ts
+++ b/packages/access-client/src/types.ts
@@ -35,9 +35,11 @@ import type {
 import type { SetRequired } from 'type-fest'
 import { Driver } from './drivers/types.js'
 import type { ColumnType, Selectable } from 'kysely'
+import { SpaceUnknown } from './errors.js'
 
 // export other types
 export * from '@web3-storage/capabilities/types'
+export * from './errors.js'
 
 /**
  * D1 Types
@@ -61,13 +63,6 @@ export interface SpaceTableMetadata {
 }
 
 /**
- * Indicates failure executing ability that requires access to a space that cannot be found by the handler
- */
-export type SpaceNotFoundFailure = Failure & {
-  status: 404
-}
-
-/**
  * Access api service definition type
  */
 export interface Service {
@@ -83,7 +78,7 @@ export interface Service {
     info: ServiceMethod<
       SpaceInfo,
       Selectable<SpaceTable>,
-      Failure | SpaceNotFoundFailure
+      Failure | SpaceUnknown
     >
     'recover-validation': ServiceMethod<
       SpaceRecoverValidation,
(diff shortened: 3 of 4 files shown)


packages/access-api/src/service/index.js
```diff
@@ -33,7 +33,11 @@ export function service(ctx) {
       info: Server.provide(Space.info, async ({ capability, invocation }) => {
         const results = await ctx.models.spaces.get(capability.with)
         if (!results) {
-          return new Failure('Space not found.')
+          return {
+            error: true,
+            status: 404,
+            message: 'Space not found.',
+          }
```
hugomrdias, line 40: we probably should start a error.js file like this https://github.com/web3-storage/ucanto/blob/main/packages/validator/src/error.js and a error.ts for error types in the client

hugomrdias: this is gonna return a http 200 so i dont think we need the status in there, just a normal `SpaceNotFound extends Failure` with `name` and `message` props.

packages/access-client/src/types.ts
```diff
@@ -60,6 +60,13 @@ export interface SpaceTableMetadata {
   agent: AgentMeta
 }
 
+/**
+ * Indicates failure executing ability that requires access to a space that cannot be found by the handler
+ */
+export type SpaceNotFoundFailure = Failure & {
+  status: 404
+}
+
```
hugomrdias, line 69: maybe create a new errors.ts file exported from types.ts with something like https://github.com/web3-storage/ucanto/blob/main/packages/interface/src/capability.ts#L413-L416

hugomrdias (changes requested):

packages/access-api/src/service/index.js
gobengo: The status is in there as the indication of what kind of error this is. I don't think we should use just a specific 'message' value as the only discriminant of this error case because doing that would make it so that changing the error message string would be a backward-incompatible change from the perspective of any client trying to detect this error case. I'm intentionally trying to add a specific non-`message` [error code](https://en.wikipedia.org/wiki/Error_code) that clients can use to detect this case.

packages/access-client/src/types.ts
```diff
@@ -60,6 +60,13 @@ export interface SpaceTableMetadata {
   agent: AgentMeta
 }
 
+/**
+ * Indicates failure executing ability that requires access to a space that cannot be found by the handler
+ */
+export type SpaceNotFoundFailure = Failure & {
+  status: 404
```
gobengo, line 67: change from status int to
```
name: 'SpaceNotFound'
```

Gozala, line 67:
```suggestion
  name: 'UnknownSpaceDID`
```

I would say ☝️, which is I think more accurate than not found. If we want to be future compatible I would go as far as do this instead.

```suggestion
  name: 'NoStorageProvider'
```

Gozala (changes requested):
Provided feedback. I don't think you need another review from me.

packages/access-client/src/types.ts
Gozala: For what it's worth I don't mind `status` along with the `name` to categorize concrete errors into a groups with HTTP semantics, I just think it's not replacement for concrete errors.

packages/access-api/src/service/index.js
hugomrdias: exactly, non-http transport should not need to understand http codes. For errors IMO either use the name as the identifier or a code nodejs style `error.code = ERR_SPACE_NOT_FOUND`

```js
// similar to ucanto
if(error.name === 'SpaceNotFound'){}

// similar to nodejs
class SpaceNotFoundError{}
SpaceNotFoundError.code = 'ERR_SPACE_NOT_FOUND'

if(error.code === SpaceNotFoundError.code) {}
```

gobengo:
also relevant: https://github.com/web3-storage/specs/issues/18

packages/access-api/src/service/index.js
gobengo: @hugomrdias Thanks, when I first responded I missed that you were encouraging a `name` and thought you were just saying to cut `status` which would make it the same as generic `Failure`. my bad

packages/access-client/src/types.ts
gobengo: I moved the failure type to an `errors.ts` file as you suggested https://github.com/web3-storage/w3protocol/pull/391/commits/8f0e5e342e858f25eb832bde4be2ea90a07175ae#diff-f4e938a2d743868395a4bb96adeecd10ce306ac53100e1e978df8d40ac33d4ef

packages/access-client/src/types.ts
gobengo: I changed it to `SpaceUnknown`

gobengo: I got rid of `status` and used `name` of `UnknownSpace` in https://github.com/web3-storage/w3protocol/pull/391/commits/8f0e5e342e858f25eb832bde4be2ea90a07175ae#diff-f4e938a2d743868395a4bb96adeecd10ce306ac53100e1e978df8d40ac33d4ef

packages/access-api/src/service/index.js
```diff
@@ -33,7 +33,11 @@ export function service(ctx) {
       info: Server.provide(Space.info, async ({ capability, invocation }) => {
         const results = await ctx.models.spaces.get(capability.with)
         if (!results) {
-          return new Failure('Space not found.')
+          return {
+            error: true,
+            status: 404,
+            message: 'Space not found.',
```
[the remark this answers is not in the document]
gobengo: I agree it would be good to have actionable feedback in the error message. I defer to @hugomrdias on whether this is the best advice to give (accept the suggestion) or whether to deal with that in separate PR.

packages/access-client/src/errors.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Indicates failure executing ability that requires access to a space that is not well-known enough to be handled.
+ * e.g. it's a space that's never been seen before,
+ * or it's a seen space that hasn't been fully registered such that the service can serve info about the space.
+ */
+export interface SpaceUnknown {
```
hugomrdias, line 6: we probably should extends `Failure` here same as other errors in ucanto

gobengo:
I updated the change here based on review of @hugomrdias and @Gozala . Now the [union tag](https://en.wikipedia.org/wiki/Tagged_union) to detect this failure type is `name` equaling `SpaceUnknown` instead of anything with a `status` property. This should still be enough for the use case at hand, which is `upload-api` having enough info on the result of `space/info` that it can distinguish in `verifiyInvocation` between an invalid invocation and a valid invocation to get info for an unknown space that `access-api` does not want to (or is unable to) serve info for.

hugomrdias (approved):
LGTM, left a comment inline but not blocking

packages/access-client/src/errors.ts
gobengo: I had omitted it on purpose so that access-client types didn't need to depend on ucanto (and we'd just rely on `tsc` checking the structural types), but don't have strong opinions on it either way, so I added the `extends` clause https://github.com/web3-storage/w3protocol/pull/391/commits/0b63c2e4bfd5890d7455042464bc6fd65459ffa7 .

PR #392, opened 2023-01-25
feat: add support for access/authorize and update

### LINKED ISSUE #386: Implement access/authorize email flow in access-api
Spec: https://github.com/web3-storage/specs/blob/main/w3-account.md#accessauthorize
Notes: https://www.notion.so/w3-provider-account-4fa2b6511e57467783d8ee7434f14c66

```mermaid
erDiagram
	ACCOUNT }o--|| SESSION: has
	ACCOUNT {
		string did PK
		string name
	}
	SESSION {
		string agent-did PK "we should only have one session per agent?"
    JSON agent-metadata
		string location
		string account-did FK "this could be just DELEGATIONS.audience"
    cid delegation-cid FK
  }
  ACCOUNT }o--|| DELEGATION: has
	DELEGATION {
		cid delegation-cid PK
		bytes delegation
    did audience "did:mailto:alice@web.mail"
    date expiration "expiration in the ucan"
  }
```

hugomrdias:
closes #386

```mermaid
sequenceDiagram
	autonumber
	actor Alice
  participant Web3
  Alice->>+Web3: access/authorize
  par Web3 responds to Alice
      Web3->>Alice: Ok
  and Web3 sends Alice email
      Web3--)-Alice: ./update
  end
	Alice --) Web3: send DID (ws)
	Web3 -->> Web3: Polls KV for DID
	Alice --) Alice: Click email
	Alice ->>+ Web3: ./update
	Web3 -->> Web3: saves ./update
	Note right of Web3: KV put [AliceDID, ./update, 5mins]
	Web3 ->>-Alice: HTML with ./update
	Note right of Web3: 5. finds Alice DID
	Web3 --) Alice: sends ./update (ws from 4.)
```

### DIFF
--- a/packages/access-api/migrations/0004_add_accounts_table.sql
+++ b/packages/access-api/migrations/0004_add_accounts_table.sql
@@ -1,0 +1,21 @@
+-- Migration number: 0004 	 2023-01-24T15:09:12.316Z
+CREATE TABLE
+    IF NOT EXISTS accounts (
+        did TEXT NOT NULL PRIMARY KEY,
+        inserted_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
+        updated_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
+        UNIQUE (did)
+    );
+
+CREATE TABLE
+    IF NOT EXISTS delegations (
+        cid TEXT NOT NULL PRIMARY KEY,
+        bytes BLOB NOT NULL,
+        audience TEXT NOT NULL,
+        issuer TEXT NOT NULL,
+        expiration TEXT,
+        inserted_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
+        updated_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
+        UNIQUE (cid),
+        FOREIGN KEY (audience) REFERENCES accounts (did) ON UPDATE CASCADE ON DELETE CASCADE
+    );
--- a/packages/access-api/src/models/accounts.js
+++ b/packages/access-api/src/models/accounts.js
@@ -1,4 +1,3 @@
-// @ts-ignore
 // eslint-disable-next-line no-unused-vars
 import * as Ucanto from '@ucanto/interface'
 import {
--- a/packages/access-api/src/models/accounts.js
+++ b/packages/access-api/src/models/accounts.js
@@ -24,7 +24,7 @@ export class Accounts {
     /** @type {GenericPlugin<DelegationRecord>} */
     const objectPlugin = new GenericPlugin({
       // eslint-disable-next-line unicorn/no-null
-      expiration: (v) => (typeof v === 'string' ? new Date(v) : null),
+      expires_at: (v) => (typeof v === 'string' ? new Date(v) : null),
       inserted_at: (v) => new Date(v),
       updated_at: (v) => new Date(v),
     })
@@ -63,7 +63,7 @@ export class Accounts {
         audience: del.audience.did(),
         issuer: del.issuer.did(),
         bytes: delegationsToBytes([del]),
-        expiration: expirationToDate(del.expiration),
+        expires_at: expirationToDate(del.expiration),
       })
       .onConflict((oc) => oc.column('cid').doNothing())
       .returningAll()
@@ -103,9 +103,9 @@ export class Accounts {
       .where('delegations.cid', '=', cid)
       .where((qb) =>
         qb
-          .where('delegations.expiration', '>=', new Date())
+          .where('delegations.expires_at', '>=', new Date())
           // eslint-disable-next-line unicorn/no-null
-          .orWhere('delegations.expiration', 'is', null)
+          .orWhere('delegations.expires_at', 'is', null)
       )
       .executeTakeFirst()
   }
--- a/packages/access-api/src/models/accounts.js
+++ b/packages/access-api/src/models/accounts.js
@@ -24,7 +24,8 @@ export class Accounts {
   constructor(d1) {
     /** @type {GenericPlugin<DelegationRecord>} */
     const objectPlugin = new GenericPlugin({
-      expiration: (v) => new Date(v),
+      // eslint-disable-next-line unicorn/no-null
+      expiration: (v) => (typeof v === 'string' ? new Date(v) : null),
       inserted_at: (v) => new Date(v),
       updated_at: (v) => new Date(v),
     })
--- a/packages/access-api/src/models/accounts.js
+++ b/packages/access-api/src/models/accounts.js
@@ -1,0 +1,112 @@
+// @ts-ignore
+// eslint-disable-next-line no-unused-vars
+import * as Ucanto from '@ucanto/interface'
+import {
+  delegationsToBytes,
+  expirationToDate,
+} from '@web3-storage/access/encoding'
+import { Kysely } from 'kysely'
+import { D1Dialect } from 'kysely-d1'
+import { GenericPlugin } from '../utils/d1.js'
+
+/**
+ * @typedef {import('@web3-storage/access/src/types.js').DelegationRecord} DelegationRecord
+ */
+
+/**
+ * Accounts
+ */
+export class Accounts {
+  /**
+   *
+   * @param {D1Database} d1
+   */
+  constructor(d1) {
+    /** @type {GenericPlugin<DelegationRecord>} */
+    const objectPlugin = new GenericPlugin({
+      expiration: (v) => new Date(v),
+      inserted_at: (v) => new Date(v),
+      updated_at: (v) => new Date(v),
+    })
+    this.d1 = /** @type {Kysely<import('../bindings').D1Schema>} */ (
+      new Kysely({
+        dialect: new D1Dialect({ database: d1 }),
+        plugins: [objectPlugin],
+      })
+    )
+  }
+
+  /**
+   * @param {Ucanto.URI<"did:">} did
+   */
+  async create(did) {
+    const result = await this.d1
+      .insertInto('accounts')
+      .values({
+        did,
+      })
+      .onConflict((oc) => oc.column('did').doNothing())
+      .returning('accounts.did')
+      .execute()
+    return { data: result }
+  }
+
+  /**
+   *
+   * @param {Ucanto.Delegation} del
+   */
+  async addDelegation(del) {
+    const result = await this.d1
+      .insertInto('delegations')
+      .values({
+        cid: del.cid.toV1().toString(),
+        audience: del.audience.did(),
+        issuer: del.issuer.did(),
+        bytes: delegationsToBytes([del]),
+        expiration: expirationToDate(del.expiration),
+      })
+      .onConflict((oc) => oc.column('cid').doNothing())
+      .returningAll()
+      .executeTakeFirst()
+    return result
+  }
+
+  /**
+   * @param {Ucanto.URI<"did:">} did
+   */
+  async get(did) {
+    return await this.d1
+      .selectFrom('accounts')
+      .selectAll()
+      .where('accounts.did', '=', did)
+      .executeTakeFirst()
+  }
+
+  /**
+   * @param {Ucanto.URI<"did:">} did
+   */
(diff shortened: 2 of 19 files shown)


alanshaw (changes requested):
🚀🚀🚀

packages/access-client/src/types.ts
```diff
@@ -64,6 +86,13 @@ export interface SpaceTableMetadata {
  * Access api service definition type
  */
 export interface Service {
+  access: {
+    authorize: ServiceMethod<
+      AccessAuthorize,
+      EncodedDelegation<[AccessSession]> | undefined,
+      Failure
+    >
```
alanshaw, line 94: This doesn't return anything.

```suggestion
    authorize: ServiceMethod<AccessAuthorize, undefined, Failure>
```

packages/access-api/migrations/0004_add_accounts_table.sql
```diff
@@ -0,0 +1,21 @@
...
+    IF NOT EXISTS delegations (
+        cid TEXT NOT NULL PRIMARY KEY,
+        bytes BLOB NOT NULL,
+        audience TEXT NOT NULL,
+        issuer TEXT NOT NULL,
+        expiration TEXT,
+        inserted_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
+        updated_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
+        UNIQUE (cid),
+        FOREIGN KEY (audience) REFERENCES accounts (did) ON UPDATE CASCADE ON DELETE CASCADE
```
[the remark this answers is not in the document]
hugomrdias: If we change something in accounts, delegation stays in sync. Its just a good default so we dont store dangling delegations because these are not issued by us. 
Updates any change to PK/FK pair (should never happen) and deletes delegations if we delete account.

packages/access-api/src/models/accounts.js
```diff
@@ -0,0 +1,112 @@
...
+   * @param {string} cid
+   */
+  async getDelegationsByCid(cid) {
+    return await this.d1
+      .selectFrom('delegations')
+      .selectAll()
+      .where('delegations.cid', '=', cid)
+      .where((qb) =>
+        qb
+          .where('delegations.expiration', '>=', new Date())
```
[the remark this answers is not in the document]
hugomrdias: https://github.com/web3-storage/w3protocol/blob/9dfdb6a648a1ef0929385e452c5589039a5227ed/packages/access-api/src/utils/d1.js#L35

packages/access-api/src/models/accounts.js
```diff
@@ -0,0 +1,112 @@
...
+ */
+export class Accounts {
+  /**
+   *
+   * @param {D1Database} d1
+   */
+  constructor(d1) {
+    /** @type {GenericPlugin<DelegationRecord>} */
+    const objectPlugin = new GenericPlugin({
+      expiration: (v) => new Date(v),
```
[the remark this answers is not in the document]
hugomrdias: nice catch

packages/access-client/src/types.ts
hugomrdias: it does for tests

packages/access-api/migrations/0004_add_accounts_table.sql
```diff
@@ -0,0 +1,21 @@
+-- Migration number: 0004 	 2023-01-24T15:09:12.316Z
+CREATE TABLE
+    IF NOT EXISTS accounts (
+        did TEXT NOT NULL PRIMARY KEY,
+        inserted_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
+        updated_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
```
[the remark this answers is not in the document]
hugomrdias: not really just a standard columns lol maybe in the future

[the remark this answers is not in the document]
hugomrdias: do you prefer that i remove both updated_at ?

packages/access-api/src/models/accounts.js
```diff
@@ -0,0 +1,113 @@
+// @ts-ignore
+// eslint-disable-next-line no-unused-vars
+import * as Ucanto from '@ucanto/interface'
```
vasco-santos, line 3: should we import types needed like below DelegationRecord instead of importing directly in source and ignoring eslint error?

packages/access-api/src/models/spaces.js
```diff
@@ -17,8 +19,17 @@ export class Spaces {
...
+    const objectPlugin = new GenericPlugin({
+      metadata: (v) => JSON.parse(v),
+      inserted_at: (v) => new Date(v),
+      updated_at: (v) => new Date(v),
+    })
     this.d1 = /** @type {Kysely<import('../bindings').D1Schema>} */ (
-      new Kysely({ dialect: new D1Dialect({ database: d1 }) })
+      new Kysely({
+        dialect: new D1Dialect({ database: d1 }),
+        plugins: [objectPlugin],
```
vasco-santos, line 31: nice we can use plugins here directly 👌🏼

packages/access-api/src/utils/html.js
```diff
@@ -1,4 +1,5 @@
 import { render } from 'preact-render-to-string'
+// @ts-ignore
 // eslint-disable-next-line no-unused-vars
 import * as Ucanto from '@ucanto/interface'
```
vasco-santos, line 4: not needed anymore right?

vasco-santos (approved):
LGTM
Left some non blocking feedback inline

packages/access-api/migrations/0004_add_accounts_table.sql
```diff
@@ -0,0 +1,21 @@
...
+        UNIQUE (did)
+    );
+
+CREATE TABLE
+    IF NOT EXISTS delegations (
+        cid TEXT NOT NULL PRIMARY KEY,
+        bytes BLOB NOT NULL,
+        audience TEXT NOT NULL,
+        issuer TEXT NOT NULL,
+        expiration TEXT,
```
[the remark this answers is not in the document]
gobengo: It appears that sqlite doesn't support full ISO-8601, only a subset of its string syntax, so we might get more sqlite compatability by storing as numbers https://www.sqlite.org/lang_datefunc.html

packages/access-api/src/models/accounts.js
gobengo: perhaps we could disable `no-unused-vars` since it is not ts-aware and instead only use `@typescript-eslint/no-unused-vars`. https://stackoverflow.com/a/61555310

packages/access-api/src/models/accounts.js
```diff
@@ -0,0 +1,113 @@
...
+   */
+  async getDelegationsByCid(cid) {
+    return await this.d1
+      .selectFrom('delegations')
+      .selectAll()
+      .where('delegations.cid', '=', cid)
+      .where((qb) =>
+        qb
+          .where('delegations.expiration', '>=', new Date())
+          // eslint-disable-next-line unicorn/no-null
```
gobengo, line 108: I've found this lint rule not so helpful. +1 to disabling it if you aren't attached to it.

packages/access-api/src/routes/validate-email.js
```diff
@@ -25,7 +31,11 @@ export async function validateEmail(req, env) {
       return new HtmlResponse(
         (
           <ValidateEmail
-            delegation={delegation}
+            email={delegation.capabilities[0].nb.identity.replace(
```
gobengo, line 34: It could be good to define a utility function for extracting the email from the ability, because this logic using `String#replace` won't work for `did:mailto` using the syntax we changed to once realizing `@` was not a valid DID character.

gobengo: and then it could be used on line 84 below as well

packages/capabilities/src/access.js
```diff
@@ -0,0 +1,102 @@
...
+      with: "did:web:web3.storage",
+      can: "./update",
+      nb: { key: "did:key:zAgent" }
+    }],
+    exp: null
+    sig: "..."
+  }
+ * ```
+ */
+export const session = capability({
```
gobengo, line 94: when I saw this used elsewhere, I assumed it was a new `access/session` `can` value and was confused at first. Now I understand.

It seems to me that since this is a special 'can' value within ucanto atm, we should make it so you can import it from ucanto. e.g. we could [export this](https://github.com/web3-storage/ucanto/blob/cb1a3bca9d51c95f369f28f1c2be46ee213ed053/packages/validator/src/lib.js#L517). agree @Gozala ?

packages/access-api/src/models/accounts.js
```diff
@@ -0,0 +1,113 @@
...
+      .returning('accounts.did')
+      .execute()
+    return { data: result }
+  }
+
+  /**
+   *
+   * @param {Ucanto.Delegation} del
+   */
+  async addDelegation(del) {
```
gobengo, line 59: it would be nice to allow optionally providing a transaction object here and in other ORM-calling methods https://koskimas.github.io/kysely/classes/Transaction.html

packages/access-api/src/routes/validate-email.js
gobengo: after reading more, I also don't really understand where `nb.identity` is coming from or what type it would be (I don't see `nb.identity` in the access capabilities). If I'm reading right, it might be type `any` when the `replace` method is called on it here. We might avoid runtime bugs by clarifying what the return type of `env.models.validations.put` is expecting (e.g. specific `Capability` types within the delegation, not generic allowing any `Capabilities` tuple) and/or doing more explicit parsing of `req.query.ucan` instead of using the type assertion on line 26

gobengo: (it would be good to add a test for what should happen if someone tries to inject malicious `req.query.ucan` and IMO it's best to ensure we don't create db records without heavy parsing)

packages/access-api/src/service/access-authorize.js
```diff
@@ -0,0 +1,45 @@
...
+export function accessAuthorizeProvider(ctx) {
+  return Server.provide(
+    Access.authorize,
+    async ({ capability, invocation }) => {
+      const session = await Access.session
+        .invoke({
+          issuer: ctx.signer,
+          audience: DID.parse(capability.nb.as),
+          with: ctx.signer.did(),
+          lifetimeInSeconds: 86_400 * 7, // 7 days
```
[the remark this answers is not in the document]
gobengo: also might be good to use an absolute datetime here instead of a duration (so that the expiry datetime is unambiguous)

packages/access-api/src/models/accounts.js
hugomrdias: there no support for transactions in kysely-d1 yet

packages/access-api/src/models/accounts.js
hugomrdias: @vasco-santos the interface package is made with an empty .js file to enable imports like this, because generic types in jsdoc typedefs are a pain.

@gobengo that could work and it looks cleaner

packages/access-api/src/models/accounts.js
hugomrdias: i very much like this rule for these reasons https://github.com/sindresorhus/meta/discussions/7 but we can discuss this in a follow up issue/pr ofc

packages/access-api/src/routes/validate-email.js
hugomrdias: this is from the old recover cap it will go away soon, regarding ucan validation i agree we should validate the caps for the specific mode and the signature. 


https://github.com/web3-storage/w3protocol/issues/411

packages/access-api/src/service/access-authorize.js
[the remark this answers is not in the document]
hugomrdias: https://github.com/web3-storage/specs/issues/43

gobengo (approved):

packages/access-api/src/models/accounts.js
gobengo: fwiw what I proposed before doesn't work, but this does https://stackoverflow.com/a/66261985

gobengo: https://github.com/web3-storage/w3protocol/pull/409/files#r1092603863

PR #397, opened 2023-01-26
fix: make tests use did:web everywhere

hugomrdias:


packages/access-api/src/config.js
```diff
@@ -1,4 +1,4 @@
-import { Signer } from '@ucanto/principal/ed25519'
+// @ts-ignore
```
vasco-santos, line 1: can we use `ts-expect-error` instead? It is easier to flag when it gets fixed

vasco-santos (approved):
LGTM

Issue #348, opened 2023-01-16
Research and avoid SafeLinks and other auto open email client stuff


hugomrdias:
More context here https://purrfect-tracker-45c.notion.site/Email-auth-flow-d33f0715024a47c18a8114ba284e9c07

- [x] Make the current response page run javascript that triggers the websocket, this way only http client that can run javascript will start the websocket to send the delegation back automatically

The email html source in the notion document has what looks like counter measures for this.

Issue #333, opened 2023-01-11
access-api: email validation


natevw:
The `access-api` package's `/validate-email` endpoint does not provide sufficient authentication **or** authorization for its anticipated purpose.

Currently the flow works roughly as follows:

1. The client generates an identity and makes a request to the API to send a validation email
2. The API sends the email including a link to approve the association between the identity⇔email
3. When the link is opened, the API sends a message back to the client which allows that association

## The problem

The flow as implemented is **not** secure/trustworthy. There are two significant flaws, closely related but kept separate to highlight the need for improvements on both sides:

* The actual broadcast should **not** be triggered by a `GET` request — this breaks [HTTP expectations generally](https://softwareengineering.stackexchange.com/a/188861/347111) and will be particularly problematic if the user has any sort of link preview enabled or if the email goes through a provider that itself scans links (e.g. the [MS Outlook Safelink](https://stackoverflow.com/questions/62108517/clicking-outlook-safelink-protection-links-in-emails-seems-to-be-executing-the-c/65095191#65095191) protection service). This is the "authentication" side: just because a URL was fetched does not mean that the intended recipient was the one that actually fetched it.
* The general authorization itself is liable to end-user confusion and exploitation: if I've never heard of w3storage I may click the link just to see what it's about. or otoh if I use w3storage vociferously, I may click the link assuming it's my own request and not someone else trying to access my account maliciously! The "authorization" here is very weak — as the email recipient I need to be able to confirm **or deny** that the new identity is okay to associate.

In most cases where email login is used, whoever *opens the link* gains access to the underlying session/identity. But in this case opening the link *gives someone else* that access! A malicious actor Eve simply provides Alice/Bob's  email address — now Eve just needs to wait for Alice's computer to generate a page preview for the link, or for Bob to click the link, to gain access to their account.

## Proposed solution

This came out of a discussion we had at https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1673387411040389 around these concerns (which were also brought up in https://github.com/web3-storage/specs/issues/26 earlier as well). There needs to be some way to "close the loop" and ensure that maliciously **requested** privileges associated with an email address aren't accidentally **approved**.

I would propose the following improved email validation flow:

1. We still allow ± anyone to generate a new identity and request a verification email to an arbitrary address. [Business side will need to consider some sort of throttling/abuse prevention here but that's a general spam thing and doesn't effect this account security issue.]
[...]

PR #398, opened 2023-01-27
fix: avoid email delegation via GET request

natevw:
The email validation approval process is now split into two stages: a GET request with no side effects except to load a page, that then auto-submits a POST request to actually continue the flow.

## Summary of problem

This fixes the API so as to follow [proper HTTP semantics](https://github.com/web3-storage/w3protocol/issues/333#issuecomment-1380925535):

> The purpose of distinguishing between safe [i.e. like GET] and unsafe [like PUT/POST] methods is to allow automated retrieval processes (spiders) and cache performance optimization (pre-fetching) to work without fear of causing harm. In addition, it allows a user agent to apply appropriate constraints on the automated use of unsafe methods when processing potentially untrusted content.

That is, a `PUT` or `POST` (rather than a `GET`) **must** be the method used in order to do things like

* cause a message to be sent (forwarding a UCAN delegation via a separate connection's websocket)
* cause an untrusted keypair to be associated with a billable email address (which is the outcome of that forwarding, in practice!)

Fixing the HTTP semantics should address all of #348, and is the first step to addressing the security concerns in #333.

## Summary of solution

Clicking (or scanning/pre-fetching/previewing/etc.) the link in the email no longer finishes the validation process. Instead, it loads a (harmless to scan/pre-fetch/preview) landing page which simply says "Validating Email" while using JavaScript to auto-complete the process.

My preference would have been to move more of the approval process *out* of the email and *into* this landing page. (So rather than auto-approving, this landing page would contain details/context and force an informed clear "Yes, approve this new space" vs. "No, I didn't want this" decision.) However, the team preferred to keep\* all that in the initial email and requested that this fix be based on an auto-approval.

Given that preference, this patch is able to fix the core HTTP semantics very self-contained:

* no changes needed to the email templates (\* though still would be good to do)
* will not break any existing unexpired links at the moment it is deployed
* is essentially the exact same UX from a user's perspective (they might notice just a little extra blink)
* does degrade gracefully if user has JS disabled, and any non-browser clients could still trigger the `POST` ± just as easy as before
* no changes needed on the `w3ui` side for this part of the email validation improvements

### DIFF
--- a/packages/access-api/src/index.js
+++ b/packages/access-api/src/index.js
@@ -4,7 +4,7 @@ import { notFound } from '@web3-storage/worker-utils/response'
 import { Router } from '@web3-storage/worker-utils/router'
 import { postRaw } from './routes/raw.js'
 import { postRoot } from './routes/root.js'
-import { validateEmail } from './routes/validate-email.js'
+import { preValidateEmail, validateEmail } from './routes/validate-email.js'
 import { validateWS } from './routes/validate-ws.js'
 import { version } from './routes/version.js'
 import { getContext } from './utils/context.js'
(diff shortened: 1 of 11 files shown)


travis:
closing, this has moved to #430

Issue #402, opened 2023-01-27
Automatically push email templates into Postmark API?


natevw:
Via https://github.com/web3-storage/w3protocol/pull/355 we've moved a copy of the Postmark email templates into this repository for change tracking. But they still need to be manually maintained: when updated in repo, need to be pasted into Postmark; if changed directly in Postmark, need to remember to update repo.

As https://github.com/web3-storage/w3protocol/pull/355#pullrequestreview-1267602628 pointed out, maybe we could just use Postmark API to push the templates from the codebase automatically?

This maybe needs a bit more planning, e.g. does it always do that every time each server process starts or when? And would it also make sure any template "layout" stuff is also set up? etc. but maybe that doesn't need to all be solved just to iterate this another step improvement.

PR #403, opened 2023-01-27
fix: align postmark/welcome.txt with .html version

### LINKED ISSUE #356: text version of verification email needs to be updated
The email templates in Postmark have two versions - HTML and plaintext - and the plaintext version has some problems - I noticed a string of "f"s and a heading that reads "### Hugo" but there may be other issues as well and the whole thing should be copy edited by someone who's in the loop on what we want those emails to say.

I confirmed that the plaintext version goes out with the verification emails we send, but only people using email clients that don't support HTML will see them.

Here's the place in the Postmark UI that needs to be updated:

Screen Shot 2023-01-17 at 2 51 01 PM
Screen Shot 2023-01-17 at 2 51 06 PM

@hugomrdias said he'd dig into this so assigning to him

natevw:
Here's an update to the **plaintext** version of the email validation Welcome template. I've basically just made it match the HTML version ± verbatim *except* the wording around the link button itself which didn't make sense in a plaintext context.

This fixes #356 in the repo, but when this is merged **someone with Postmark account access will need to paste the new content there** manually before the situation is truly resolved. (See https://github.com/web3-storage/w3protocol/blob/953cbae/packages/access-api/postmark/README.md and xref #402.)


natevw:
@travis could you take a look and if it looks good, get this merged here and pasted into Postmark too?

travis:
closing, moved to #431

Issue #344, opened 2023-01-14
access-client Agent.url defaults to `HOST` even when custom connection is used


natevw:
I am having trouble connecting to a locally-hosted API from the web3ui `KeyringProvider`. I have been able to switch the URL so that the email validation HTTP `POST /` goes to my configured <http://localhost:8787> URL. But the `/validate-ws` WebSocket connection continues connecting to the default <https://access.web3.storage> where it doesn't receive the (local) successful delegation message.

I believe the code at https://github.com/web3-storage/w3protocol/blob/27f2f60dac7c95cb9efd42a28f5abfef8bdeb197/packages/access-client/src/agent.js#L82 might be incorrect?

Based on its usage in…

* https://github.com/web3-storage/w3ui/blob/c802e9909289a113bf646f22d25e5c6fbd1c1d3e/packages/react-keyring/src/providers/Keyring.tsx#L50 (react-keyring)
* https://github.com/web3-storage/w3ui/blob/c802e9909289a113bf646f22d25e5c6fbd1c1d3e/packages/solid-keyring/src/providers/Keyring.ts#L55 (solid-keyring)
* https://github.com/web3-storage/w3ui/blob/c802e9909289a113bf646f22d25e5c6fbd1c1d3e/packages/vue-keyring/src/providers/Keyring.ts#L62 (vue-keyring)
* (^^^ which all funnel through https://github.com/web3-storage/w3ui/blob/c802e9909289a113bf646f22d25e5c6fbd1c1d3e/packages/keyring-core/src/index.ts#L135)

…it seems that the assignment should probably fallback first to the *connection's* URL before jumping straight to using the default. I.e. something like:

```
this.url = options.connection?.url ?? options.url ?? new URL(HOST) 
```
so that if a connection is provided its URL is used, otherwise falling back to an auto-created connection. Does that match the intent of the `connection` vs. `url` options, or would there ever be a use case to have an `Agent.url` mismatched versus the `Agent.connection.url`?

---

Background to reproduce:

Using the tips from <https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1669658502951759?thread_ts=1669658366.189139&cid=C02BZPRS9HP>, I have added a file retaining the "staging-service.js" name for simplicity, although this is pointing at a **local** service instead:

```
import { connect } from '@ucanto/client'
import { CAR, CBOR, HTTP } from '@ucanto/transport'
import * as DID from '@ipld/dag-ucan/did'

[...]

PR #404, opened 2023-01-27
fix: inconsistent w3ui custom URL handling

natevw:
This is imo the cleanest way of resolving issue #344, where a custom server URL is used only for the individual HTTP requests and not for the corresponding WebSocket connection.

It took a while to decide the best place to fix this. For a little while, it seemed like the internals of `createAgent()` within  `@w3ui/keyring-core` needed to copy out the URL into its the top-level options. But ultimately I believe that would be an incorrect workaround!

The root cause is that `@web3-storage/access/agent` is pulling some tricks on its own that get it into trouble: instead of the `Agent` relying exclusive on its `this.connection` for all actual transport, it also cheats and creates its own `new Websocket(…)` on the side!

That extra websocket is the only reason the `Agent` instance keeps a copy of `this.url` at all — everything else goes through the proper `ConnectionView<Service>` abstraction. So I believe the `w3ui` code is fulfilling its side of the interface contract. Since it is the `Agent` that bypasses the abstraction, it is responsible to look for the raw URL inside of it.

This ends up being a relatively trivial code change, but did require some special (internal) extensions to the type definitions to support.

### DIFF
--- a/packages/access-client/src/agent.js
+++ b/packages/access-client/src/agent.js
@@ -79,7 +79,7 @@ export class Agent {
    * @param {import('./types').AgentOptions} [options]
    */
   constructor(data, options = {}) {
-    this.url = options.connection?.url ?? options.url ?? new URL(HOST)
+    this.url = options.connection?.channel.url ?? options.url ?? new URL(HOST)
     this.connection =
       options.connection ??
       connection({
--- a/packages/access-client/src/agent.js
+++ b/packages/access-client/src/agent.js
@@ -79,7 +79,7 @@ export class Agent {
    * @param {import('./types').AgentOptions} [options]
    */
   constructor(data, options = {}) {
-    this.url = options.url ?? new URL(HOST)
+    this.url = options.connection?.url ?? options.url ?? new URL(HOST)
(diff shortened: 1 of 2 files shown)


natevw:
Looks like @alanshaw and @hugomrdias might be good reviewers for this? (And if this PR is approved please do the merge too, since I do not have write access.)

hugomrdias:
the pr has multiple changes already merged to main or unrelated can you fix this please, it would help with the review

natevw:
@hugomrdias Rebased this against current main, so it now shows just the small patch needed.

hugomrdias:
i created this PR https://github.com/web3-storage/ucanto/pull/214 in ucanto to avoid the custom types here

alanshaw:
IMHO we should just lose the additional types here to get this mergable. We can do something like https://github.com/web3-storage/ucanto/pull/214#issuecomment-1437271520 if we really need to appease typescript.

travis:
Ah sweet - I like this approach @alanshaw - will update and get it moving

travis:
closing in favor of #440

Issue #414, opened 2023-02-01
export `access/delegate` capability parser from `@web3-storage/capabilities`


gobengo:
Motivation:
* part of https://github.com/web3-storage/w3protocol/issues/393
* unblocks further work using these capability parsers, e.g. with `Server.provide` in an access-api service method

gobengo:
it's out in `@web3-storage/capabilities@2.3.0`

PR #420, opened 2023-02-01
feat: add access/delegate capability parser exported from @web3-storage/capabilities

gobengo:
Motivation:
* https://github.com/web3-storage/w3protocol/issues/414

### DIFF
--- a/packages/capabilities/src/access.js
+++ b/packages/capabilities/src/access.js
@@ -121,6 +121,9 @@ export const delegate = base.derive({
     with: DID.match({ method: 'key' }),
     nb: {
       // keys SHOULD be CIDs, but we won't require it in the schema
+      /**
+       * @type {Schema.Schema<AccessDelegateDelegations>}
+       */
       delegations: Schema.dictionary({
         value: Schema.Link.match(),
       }),
@@ -141,11 +144,16 @@ export const delegate = base.derive({
 })
 
 /**
+ * @typedef {Schema.Dictionary<string, Types.Link<unknown, number, number, 0 | 1>>} AccessDelegateDelegations
+ */
+
+/**
  * Parsed Capability for access/delegate
  *
- * @template {Types.Ability} A
- * @template {Types.URI} R
- * @typedef {Types.ParsedCapability<A, R, { delegations?: Types.Failure | Schema.Dictionary<string, Types.Link<unknown, number, number, 0 | 1>> }>} ParsedAccessDelegate
+ * @typedef {object} ParsedAccessDelegate
+ * @property {string} can
+ * @property {object} nb
+ * @property {AccessDelegateDelegations} [nb.delegations]
  */
 
 /**
@@ -154,13 +162,11 @@ export const delegate = base.derive({
  * checks that the claimed delegation set is equal to or less than the proven delegation set.
  * usable with {import('@ucanto/interface').Derives}.
  *
- * @template {Types.Ability} A
- * @template {Types.URI} R
- * @param {ParsedAccessDelegate<A,R>} claim
- * @param {ParsedAccessDelegate<A,R>} proof
+ * @param {ParsedAccessDelegate} claim
+ * @param {ParsedAccessDelegate} proof
  */
 function subsetsNbDelegations(claim, proof) {
-  /** @param {ParsedAccessDelegate<A,R>} ucan */
+  /** @param {ParsedAccessDelegate} ucan */
   const nbDelegationsCids = (ucan) =>
     new Set(Object.values(ucan.nb.delegations || {}).map(String))
   const missingProofs = setDifference(
--- a/packages/capabilities/src/access.js
+++ b/packages/capabilities/src/access.js
@@ -166,12 +166,9 @@ export const delegate = base.derive({
  * @param {ParsedAccessDelegate} proof
  */
 function subsetsNbDelegations(claim, proof) {
-  /** @param {ParsedAccessDelegate} ucan */
-  const nbDelegationsCids = (ucan) =>
-    new Set(Object.values(ucan.nb.delegations || {}).map(String))
   const missingProofs = setDifference(
-    new Set(nbDelegationsCids(claim)),
-    new Set(nbDelegationsCids(proof))
+    delegatedCids(claim),
+    new Set(delegatedCids(proof))
   )
   if (missingProofs.size > 0) {
     return new Failure(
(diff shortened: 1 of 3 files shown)


packages/capabilities/src/access.js
```diff
@@ -108,3 +108,77 @@ export const claim = base.derive({
...
+    nb: {
+      // keys SHOULD be CIDs, but we won't require it in the schema
+      delegations: Schema.dictionary({
+        value: Schema.Link.match(),
+      }),
+    },
+    derives: (claim, proof) => {
+      return (
+        fail(equalWith(claim, proof)) ||
+        // @ts-ignore
```
gobengo, line 131: @Gozala I could use some help/pairing trying to not need this

packages/capabilities/src/access.js
```diff
@@ -108,3 +108,86 @@ export const claim = base.derive({
...
+    return fail(equalWith(claim, proof)) || true
+  },
+})
+
+/**
+ * Parsed Capability for access/delegate
+ *
+ * @template {Types.Ability} A
+ * @template {Types.URI} R
+ * @typedef {Types.ParsedCapability<A, R, { delegations?: Types.Failure | Schema.Dictionary<string, Types.Link<unknown, number, number, 0 | 1>> }>} ParsedAccessDelegate
```
gobengo, line 148: I use this below on lines 159, 160. @Gozala lmk if there is some better way of inferring this type. It's possible I can't just infer it from `delegate` because it's also used in the definition of `delegate` (derives) so there is a circularity?

Gozala: `delegations` should not be `Failure`,  ucanto will not create parsed capability like that. It will either parse successfully or just fail, it will never partially fail.

Gozala:
```suggestion
 * @typedef {Types.ParsedCapability<A, R, { delegations?: Schema.Dictionary<string, Types.Link<unknown, number, number, 0 | 1>> }>} ParsedAccessDelegate
```

Does this not work ?

gobengo: @Gozala and I met together, and realized I needed to [add this type assertion](https://github.com/web3-storage/w3protocol/pull/420/commits/4f1f2931cecff1d1d1d29e889c4fdfb63ff3b327#diff-e434cc6c1a699df311a0b2faed199a2ff6b6b291d30f95e20b2ea5abfa7da3d9R125)

gobengo: I proposed this in ucanto to make the type assertion unnecessary https://github.com/web3-storage/ucanto/pull/215

packages/capabilities/src/access.js
```diff
@@ -108,3 +108,90 @@ export const claim = base.derive({
...
+ * checks that the claimed delegation set is equal to or less than the proven delegation set.
+ * usable with {import('@ucanto/interface').Derives}.
+ *
+ * @param {ParsedAccessDelegate} claim
+ * @param {ParsedAccessDelegate} proof
+ */
+function subsetsNbDelegations(claim, proof) {
+  /** @param {ParsedAccessDelegate} ucan */
+  const nbDelegationsCids = (ucan) =>
+    new Set(Object.values(ucan.nb.delegations || {}).map(String))
```
Gozala, line 171: Nit: can you hoist this function, to the top level since it does not enclose anything

packages/capabilities/src/access.js
```diff
@@ -108,3 +108,90 @@ export const claim = base.derive({
...
+  const missingProofs = setDifference(
+    new Set(nbDelegationsCids(claim)),
```
Gozala, line 173: You create a set and then pass it into a set constructor, which seems unnecessary

packages/capabilities/test/capabilities/access.test.js
```diff
@@ -371,3 +371,256 @@ describe('access capabilities', function () {
...
+describe('access/delegate', () => {
+  it('can create valid delegations and authorize them', async () => {
+    const issuer = alice
+    const audience = service.withDID('did:web:web3.storage')
+    const bobCanStoreAllWithAlice = await delegate({
+      issuer: alice,
+      audience: bob,
+      capabilities: [{ can: 'store/*', with: alice.did() }],
+    })
+    const validInvocations = [
```
Gozala, line 384: This is really hard to follow and if one of the cases fails the whole test will fail as well. Could you generate tests instead meaning have loop that calls `it` ? It would be easier to follow and failure in one case will not fail the whole suite.

packages/capabilities/test/capabilities/access.test.js
```diff
@@ -371,3 +371,256 @@ describe('access capabilities', function () {
...
+          },
+        })
+        .delegate(),
+      // example from spec, but once each with several different, but all valid, property names to use in the `nb.delegations` dict
+      // https://github.com/web3-storage/specs/blob/7e662a2d9ada4e3fc22a7a68f84871bff0a5380c/w3-access.md?plain=1#L58
+      .../** @type {const} */ ([
+        // correct cid
+        [bobCanStoreAllWithAlice.cid.toString(), bobCanStoreAllWithAlice.cid],
+        // not a cid at all
+        ['thisIsNotACid', bobCanStoreAllWithAlice.cid],
```
Gozala, line 402: 🤔 I'm starting to wonder if the whole dict is a good idea here, but then if we chose list it also may not be ordered.

Gozala (changes requested):
Overall looks good, put bunch of comments but ones I'd like to see addressed before landing are:

1. Remove one layer of set wrapper `new Set(new Set(...`
2. Adjustment to how tests are generated so it's easier to follow what going on.

Thanks

packages/capabilities/src/access.js
gobengo: https://github.com/web3-storage/w3protocol/pull/420/commits/d03e081ef3bfda4e2aeaa01e09aded00b27c5efa

packages/capabilities/test/capabilities/access.test.js
gobengo: https://github.com/web3-storage/w3protocol/pull/420/commits/edeee6b9d3d619c8468b93c87d1276255effc68f

packages/capabilities/src/access.js
gobengo: hoisted in https://github.com/web3-storage/w3protocol/commit/d03e081ef3bfda4e2aeaa01e09aded00b27c5efa

Gozala (approved):
Suggested minor changes, feel free to land as is or change a bit and then land.

packages/capabilities/test/capabilities/access.test.js
```diff
@@ -371,3 +371,290 @@ describe('access capabilities', function () {
...
+function createSelfIssuedDelegateInvocation(alice, service) {
+  return Access.delegate
+    .invoke({
+      issuer: alice,
+      audience: service,
+      with: alice.did(),
+      nb: {
+        delegations: {},
+      },
+    })
```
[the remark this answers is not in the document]
gobengo: > This function is used just once and does not really improve readability of test.

I removed it https://github.com/web3-storage/w3protocol/pull/420/commits/57f51ea034b7134d7593481445a4f2b8650f3d31

> I would suggest changing it as follows and reusing it all the other tests

I'd prefer not to wrap use of `Access.delegate.invoke()` by other tests unnecessarily, because I find it easier to reason about the other tests if they just use the capability parser directly vs using it through a wrapper like `delegateAccess`

Issue #425, opened 2023-02-06
access-api can handle `access/delegate` invocations for `did:key`


gobengo:
Motivation:
* part of https://github.com/web3-storage/w3protocol/issues/393

* goal: unblock #395 sending `access/delegate` invocations and getting a non-error result
*  context:
    * #414 added a capabilityParser for `access/delegate` to `@web3-storage/capabilities` package
    * so the next thing someone might try to do with that is execute an invocation on a connection to access-api
    * but if they tried that now, they'd get some error
        * [x] test this in [w3protocol-test](https://github.com/gobengo/w3protocol-test) https://github.com/gobengo/w3protocol-test/pull/3/commits/0f5573adbe8c3250a5e68cf8b74e286ae627328c
        * [x] test access/delegate with a registered space on staging using w3protocol-test https://github.com/gobengo/w3protocol-test/issues/7
    * when #393 is done, there shouldn't be an error for valid access/delegate invocations, so the goal of this ticket to is accomplish/test that invariant

gobengo:
#status wip on #427. started getting into both sqlite/d1 storage and access/claim invocation handling so I can test a flow of `access/delegate` vs `access/claim`.

the capabilityParsers are now exported in `@web3-storage/capabilities@2.3.0` and sample code is here https://github.com/gobengo/w3protocol-test/pull/3

gobengo:
#status merged https://github.com/web3-storage/w3protocol/pull/427 and tested it on staging via https://github.com/gobengo/w3protocol-test/pull/3

I need to write a more advanced test that involves a 'registered' space to validate behavior beyond the 'InsufficientStorage' error that happens now. https://github.com/gobengo/w3protocol-test/pull/3/files#diff-698e92d7467a4a470689d4cb120272c6cac28864d4960ac12a3720ab7c35a15cR329

gobengo:
This hanlding is implemented in access-api w3protocol main branch.
And it's used by `@web3-storage/access` (access-client) as of https://github.com/web3-storage/w3protocol/issues/395

Issue #394, opened 2023-01-25
Implement `access/claim` in `access-api`


alanshaw:
https://github.com/web3-storage/specs/blob/feat/w3-authorization/w3-access.md#accessclaim

Tasks
* [x] https://github.com/web3-storage/w3protocol/issues/408
* [x] #393 - this is worth doing first as `access/claim` wont be useful for claiming anything until clients use `access/delegate` to write things that can be claimed [comment](https://github.com/web3-storage/w3protocol/issues/394#issuecomment-1412712124)
* [x] #455
* [x] #454
  * unblocks ["Access across multiple devices"](https://github.com/web3-storage/specs/blob/main/w3-access.md#access-across-multiple-devices)

gobengo:
* [ ] define `access/claim` ability schema in `@web3-storage/capabilities`

gobengo:
practically speaking, this can't be done until 

See [spec examples](https://github.com/web3-storage/specs/blob/main/w3-access.md#access-across-multiple-devices) for how `access/claim` is claiming delegations written by `access/delegate`.

So I added #393 as a subtask of this to reflect that dependency

gobengo:
Remember to try to include this optimization
* https://github.com/web3-storage/w3protocol/pull/427#issuecomment-1442458085
* e.g. `DelegationsStorage` can have a method to query delegations out as car bytes but without parsing, that way we don't have to re-encode them to car bytes in the `access/claim` handler

gobengo:
This is implemented in access-api main branch, and is depended on by `@web3-storage/access` as of https://github.com/web3-storage/w3protocol/issues/395

PR #427, opened 2023-02-07
feat: handle access/delegate invocations without error

gobengo:
Note:
* this is a superset of https://github.com/web3-storage/w3protocol/pull/420
  * that defines the `access/delegate` capability parsers that this PR adds handlers for in access-api 
  * my intention is to land that #420 before this

Motivation:
* part of https://github.com/web3-storage/w3protocol/issues/425

### DIFF
--- a/packages/access-api/src/bindings.d.ts
+++ b/packages/access-api/src/bindings.d.ts
@@ -39,6 +39,7 @@ export interface Env {
   PRIVATE_KEY: string
   SENTRY_DSN: string
   POSTMARK_TOKEN: string
+  POSTMARK_SENDER?: string
   LOGTAIL_TOKEN: string
   // bindings
   SPACES: KVNamespace
--- a/packages/access-api/src/bindings.d.ts
+++ b/packages/access-api/src/bindings.d.ts
@@ -11,6 +11,7 @@ import { Validations } from './models/validations.js'
 import { loadConfig } from './config.js'
 import { ConnectionView, Signer as EdSigner } from '@ucanto/principal/ed25519'
 import { Accounts } from './models/accounts.js'
+import { DelegationsStorage } from './types/delegations.js'
 
 export {}
 
@@ -59,6 +60,7 @@ export interface RouteContext {
     spaces: Spaces
     validations: Validations
     accounts: Accounts
+    delegations: DelegationsStorage
   }
   uploadApi: ConnectionView
 }
--- a/packages/access-api/src/bindings.d.ts
+++ b/packages/access-api/src/bindings.d.ts
@@ -9,7 +9,7 @@ import { Email } from './utils/email.js'
 import { Spaces } from './models/spaces.js'
 import { Validations } from './models/validations.js'
 import { loadConfig } from './config.js'
-import { Signer as EdSigner } from '@ucanto/principal/ed25519'
+import { ConnectionView, Signer as EdSigner } from '@ucanto/principal/ed25519'
 import { Accounts } from './models/accounts.js'
 
 export {}
@@ -35,7 +35,6 @@ export interface Env {
   DID: `did:web:${string}`
   // URLs to upload-api so we proxy invocations to it
   UPLOAD_API_URL: string
-  UPLOAD_API_URL_STAGING: string
   // secrets
   PRIVATE_KEY: string
   SENTRY_DSN: string
@@ -60,10 +59,7 @@ export interface RouteContext {
     validations: Validations
     accounts: Accounts
   }
-  uploadApi: {
-    production?: URL
-    staging?: URL
-  }
+  uploadApi: ConnectionView
 }
 
 export type Handler = _Handler<RouteContext>
--- a/packages/access-api/src/bindings.d.ts
+++ b/packages/access-api/src/bindings.d.ts
@@ -11,7 +11,7 @@ import { Validations } from './models/validations.js'
 import { loadConfig } from './config.js'
 import { ConnectionView, Signer as EdSigner } from '@ucanto/principal/ed25519'
 import { Accounts } from './models/accounts.js'
-import { DelegationsStorage } from './types/delegations.js'
+import { DelegationsStorage as Delegations } from './types/delegations.js'
 
 export {}
 
@@ -60,7 +60,7 @@ export interface RouteContext {
     spaces: Spaces
     validations: Validations
     accounts: Accounts
-    delegations: DelegationsStorage
+    delegations: Delegations
   }
   uploadApi: ConnectionView
 }
--- a/packages/access-api/src/models/delegations.js
+++ b/packages/access-api/src/models/delegations.js
@@ -1,5 +1,8 @@
 import * as Ucanto from '@ucanto/interface'
-import { delegationsToBytes } from '@web3-storage/access/encoding'
+import {
+  delegationsToBytes,
+  bytesToDelegations,
+} from '@web3-storage/access/encoding'
 
 /**
  * @typedef {import('@web3-storage/access/src/types').DelegationTable} DelegationRow
@@ -88,6 +91,42 @@ export class DbDelegationsStorage {
   [Symbol.iterator]() {
     throw new Error(`NotImplemented: D1DelegationsStorage#[Symbol.iterator]`)
   }
+
+  /**
+   * @returns {AsyncIterableIterator<Ucanto.Delegation>}
+   */
+  async *[Symbol.asyncIterator]() {
+    if (this.#db.canStream) {
+      for await (const row of this.#db
+        .selectFrom('delegations')
+        .select(['bytes'])
+        .stream()) {
+        yield rowToDelegation(row)
+      }
+    } else {
+      // this db doesn't support .stream() so do something worse
+      for (const row of await this.#db
+        .selectFrom('delegations')
+        .select(['bytes'])
+        .execute()) {
+        yield rowToDelegation(row)
+      }
+    }
+  }
+}
+
+/**
+ * @param {Pick<DelegationRow, 'bytes'>} row
+ * @returns {Ucanto.Delegation}
+ */
+function rowToDelegation(row) {
+  const delegations = bytesToDelegations(row.bytes)
+  if (delegations.length !== 1) {
+    throw new Error(
+      `unexpected number of delegations from bytes: ${delegations.length}`
+    )
+  }
+  return delegations[0]
 }
 
 /**
--- a/packages/access-api/src/models/delegations.js
+++ b/packages/access-api/src/models/delegations.js
@@ -36,7 +36,32 @@ export class DbDelegationsStorage {
 
   /** @type {import("../types/delegations").DelegationsStorage['length']} */
   get length() {
-    throw new Error(`NotImplemented: D1DelegationsStorage#length`)
+    return new Promise((resolve, reject) => {
+      this.#db
+        .selectFrom('delegations')
+        .select((e) => e.fn.count('cid').as('size'))
+        .executeTakeFirstOrThrow()
+        .then(({ size }) => {
+          if (typeof size === 'string') {
+            const sizeNumber = parseInt(size, 10)
+            if (isNaN(sizeNumber)) {
+              throw new TypeError(
+                `unable to determine size number of delegations table`
+              )
+            }
+            return sizeNumber
+          }
+          if (typeof size === 'bigint') {
+            if (size > Number.MAX_SAFE_INTEGER) {
+              throw new TypeError(`table size too big for js Number`)
+            }
+            return Number(size)
+          }
+          return size
+        })
+        .then(resolve)
+        .catch(reject)
+    })
   }
 
   /**
--- a/packages/access-api/src/models/delegations.js
+++ b/packages/access-api/src/models/delegations.js
@@ -70,14 +70,17 @@ export class DbDelegationsStorage {
    * store items
    *
    * @param  {Array<Ucanto.Delegation>} delegations
+   * @returns {Promise<void>}
    */
   push = async (...delegations) => {
+    if (delegations.length === 0) {
+      return
+    }
     const values = delegations.map((d) => createDelegationRowUpdate(d))
     await this.#db
       .insertInto('delegations')
       .values(values)
       .onConflict((oc) => oc.column('cid').doNothing())
-      .returningAll()
       .executeTakeFirst()
   }
 
--- a/packages/access-api/src/models/delegations.js
+++ b/packages/access-api/src/models/delegations.js
@@ -46,6 +46,15 @@ export class DbDelegationsStorage {
   }
 
   /**
+   * @param {import('../types/delegations').Query} query
+   */
+  async *find(query) {
+    for await (const row of await selectByAudience(this.#db, query.audience)) {
+      yield rowToDelegation(row)
+    }
+  }
+
+  /**
    * store items
    *
    * @param  {Array<Ucanto.Delegation>} delegations
@@ -111,4 +120,16 @@ function createDelegationRowUpdate(d) {
     issuer: d.issuer.did(),
     bytes: delegationsToBytes([d]),
   }
+}
+
+/**
+ * @param {DelegationsDatabase} db
+ * @param {Ucanto.DID<'key'>} audience
+ */
+async function selectByAudience(db, audience) {
+  return await db
+    .selectFrom('delegations')
+    .selectAll()
+    .where('delegations.audience', '=', audience)
+    .execute()
 }
--- a/packages/access-api/src/models/delegations.js
+++ b/packages/access-api/src/models/delegations.js
@@ -79,14 +79,6 @@ export class DbDelegationsStorage {
       .onConflict((oc) => oc.column('cid').doNothing())
       .returningAll()
       .executeTakeFirst()
-  };
-
-  // eslint-disable-next-line jsdoc/require-returns-check
-  /**
-   * @returns {IterableIterator<Ucanto.Delegation>}
-   */
-  [Symbol.iterator]() {
-    throw new Error(`NotImplemented: D1DelegationsStorage#[Symbol.iterator]`)
   }
 
   /**
--- a/packages/access-api/src/models/delegations.js
+++ b/packages/access-api/src/models/delegations.js
@@ -70,8 +70,11 @@ export class DbDelegationsStorage {
    */
   async *[Symbol.asyncIterator]() {
     if (!this.#db.canStream) {
-      throw new Error(
-        `cannot create asyncIterator because the underlying database does not support streaming`
+      throw Object.assign(
+        new Error(
+          `cannot create asyncIterator because the underlying database does not support streaming`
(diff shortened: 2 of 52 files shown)


packages/access-client/src/types.ts
```diff
@@ -90,6 +91,7 @@ export interface Service {
   access: {
     // returns a URL string for tests or nothing in other envs
     authorize: ServiceMethod<AccessAuthorize, string | undefined, Failure>
+    delegate: ServiceMethod<AccessDelegate, void, Failure>
```
gobengo, line 94: afaict there is no response type [in the spec](https://github.com/web3-storage/specs/blob/main/w3-access.md#accessdelegate).

It also looks like `void` is a fine `Success` placeholder for now, because I don't expect the invoker to need anything from  the `Success` other than knowing it wasn't a Failure. i.e. if `access/delegate` invocation was handled without error, then the delegations were persisted.

gobengo:
@Gozala I added [this](https://github.com/web3-storage/w3protocol/pull/427/commits/e0834964e6a2d0ab2f5bcdd012c45c9c02dda80a) to make it so this patch is safe to merge/release even while I still intend to add better validation/hardening before invoking `access/delegate` should actually store delegations in the db in prod, which I will probably need to iterate on a bit after I'm out next week.

I'm hoping with this change, I can get some early review feedback now, and if there's nothing too scary, we could even consider merging this partial progress into main, so others can build on it while I'm out, but there will be no risk of half-baked functionality running in production if main branch is released to production.

packages/access-api/src/service/access-claim.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * @param {object} options
+ * @param {import('../types/delegations').DelegationsStorage} options.delegations
+ * @returns {AccessClaimHandler}
+ */
+export function createAccessClaimHandler({ delegations }) {
+  /** @type {AccessClaimHandler} */
+  return async (invocation) => {
+    // @todo - this should filter based on invocation
```
gobengo, line 47: I will resolve this todo as part of https://github.com/web3-storage/w3protocol/issues/394
For now this implementation is sufficient for the most basic of 'test delegate, then claim' that I started.
But as that got larger, I realized I should just focus on `access/delegate` in this review round.

gobengo:
@alanshaw wdyt? It's a lot of new code and a MVP `access/delegate`, but has feature switches so the new functionality won't be enabled in production. I'd love some feedback. If it looks safe, perhaps we could merge it into main to unblock further test/dev against staging.

packages/access-api/src/models/spaces.js
```diff
@@ -21,7 +21,12 @@ export class Spaces {
   constructor(d1) {
     /** @type {GenericPlugin<SpaceRecord>} */
     const objectPlugin = new GenericPlugin({
-      metadata: (v) => JSON.parse(v),
+      metadata: (v) => {
+        if (v === 'EMPTY') {
```
alanshaw, line 25: Why "EMPTY" and not just ""?

packages/access-api/src/bindings.d.ts
```diff
@@ -59,6 +60,7 @@ export interface RouteContext {
     spaces: Spaces
     validations: Validations
     accounts: Accounts
+    delegations: DelegationsStorage
```
alanshaw, line 63: Please change all or be consistent...🙏

```suggestion
    delegations: Delegations
```

packages/access-api/src/models/delegations.js
```diff
@@ -0,0 +1,137 @@
...
+   */
+  constructor(db) {
+    this.#db = db
+    // eslint-disable-next-line no-void
+    void (
+      /** @type {import('../types/delegations').DelegationsStorage} */ (this)
+    )
+  }
+
+  get length() {
```
alanshaw, line 40: It's unusual (IMHO) for a property accessor to return a promise. Why not a function?

```suggestion
  async length() {
```

packages/access-api/src/models/delegations.js
```diff
@@ -0,0 +1,137 @@
...
+            if (size > Number.MAX_SAFE_INTEGER) {
+              throw new TypeError(`table size too big for js Number`)
+            }
+            return Number(size)
+          }
+          return size
+        })
+        .then(resolve)
+        .catch(reject)
+    })
```
alanshaw, line 66: Many questions:

* Why avoiding `await`?
* Why not return a `BigInt`?

```suggestion
    const data = await this.#db
        .selectFrom('delegations')
        .select((e) => e.fn.count('cid').as('size'))
        .executeTakeFirstOrThrow()
    return BigInt(data.size)
```

packages/access-api/src/models/delegations.js
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  /**
+   * store items
+   *
+   * @param  {Array<Ucanto.Delegation>} delegations
+   * @returns {Promise<void>}
+   */
+  push = async (...delegations) => {
```
alanshaw, line 75:
```suggestion
  async push (...delegations) {
```

packages/access-api/src/models/delegations.js
```diff
@@ -0,0 +1,137 @@
...
+      .executeTakeFirst()
+  }
+
+  /**
+   * iterate through all stored items
+   *
+   * @returns {AsyncIterableIterator<Ucanto.Delegation>}
+   */
+  async *[Symbol.asyncIterator]() {
+    if (this.#db.canStream) {
```
alanshaw, line 93: We should not allow this if the database cannot stream.

packages/access-api/src/service/access-delegate.js
```diff
@@ -0,0 +1,107 @@
...
+import { createDelegationsStorage } from './delegations.js'
+
+/**
+ * access/delegate failure due to the 'with' resource not having
+ * enough storage capacity to store the delegation.
+ * https://github.com/web3-storage/specs/blob/7e662a2d9ada4e3fc22a7a68f84871bff0a5380c/w3-access.md?plain=1#L94
+ *
+ * Semantics inspired by https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/507
+ *
+ * @typedef {import('@web3-storage/capabilities/types').InsufficientStorage} InsufficientStorage
```
alanshaw, line 13: Can we subclass Ucanto's [`Failure`](https://github.com/web3-storage/ucanto/blob/main/packages/validator/src/error.js#L8) instead?

packages/access-api/src/service/access-delegate.js
```diff
@@ -0,0 +1,107 @@
...
+    await delegations.push(...delegated)
+    return {}
+  }
+}
+
+/**
+ * @param {Ucanto.Invocation<import('@web3-storage/capabilities/types').AccessDelegate>} invocation
+ * @returns {Iterable<Ucanto.Delegation<Ucanto.Capabilities>>}
+ */
+function* extractProvenDelegations({ proofs, capabilities }) {
```
alanshaw, line 84: Maybe personal preference but I'd prefer collecting in an array to using a generator unless there's some kind of streaming benefit to be had.

packages/access-api/src/service/access-delegate.js
```diff
@@ -0,0 +1,107 @@
...
+  allowServiceWithoutStorageProvider = false,
+} = {}) {
+  return async (invocation) => {
+    const capabability = invocation.capabilities[0]
+    if (
+      !allowServiceWithoutStorageProvider &&
+      !(await hasStorageProvider(capabability.with))
+    ) {
+      return {
+        name: 'InsufficientStorage',
```
alanshaw, line 69: Is more `MissingProvider`...no?

packages/access-api/src/types/delegations.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface DelegationsStorage<
+  Cap extends Ucanto.Capability = Ucanto.Capability
+> {
+  /**
+   * push items into storage
+   *
+   * @param delegations - delegations to store
+   */
+  push: (
```
alanshaw, line 13: I'm not convinced this is a `push` - it's not ordered in any way - `put` may be more appropriate or `putMany`

packages/access-api/src/types/delegations.ts
```diff
@@ -0,0 +1,28 @@
+import * as Ucanto from '@ucanto/interface'
+
+export type Resolvable<T> = T | Promise<T>
```
alanshaw, line 3: Do we need this right now? Is `Promise<T>` ok? Commonly in nodejs you'd find `*Sync` suffixed versions.

alanshaw (changes requested):

packages/access-api/src/types/delegations.ts
Gozala: We have `Await` type in multiformats for this and I think it’s also used and reexported by ucanto

Gozala (commented):
I really quickly glanced at the pr and noticed that we’m be pulling delegation as bytes from db, decode them into delegations which then ucanto will take and encode back into bytes. Decode/encode there seems pretty wasteful & perhaps can be avoided by sending bytes and let client decode

packages/access-api/src/models/spaces.js
gobengo: https://github.com/web3-storage/w3protocol/issues/447

gobengo: I pushed a comment explaining the magic `EMPTY`

packages/access-api/src/bindings.d.ts
gobengo: I changed the import name so it looks like the others.

The thing is, this isn't the same as the others. This is an interface, whereas the other 'models' here point to classes.
@Gozala and I discussed in this work trying eager to apply the [dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) and depend on an interface here instead of an implementation/class.

So that's why the name is inconsistent, it's not doing the same thing as others.

packages/access-api/src/models/delegations.js
gobengo: originally this type started as a subset of `Array` and I did some TDD from there to implement the interface backed by sql, and that required affording for async. I figured it would make sense to adjust the names/types later in the context of a specific reviewer. Happy to change this to a function now. will do

gobengo: IMHO properties make sense for nouns like 'length' and functions make sense for methods. I'm gonna call the function count() (inspired by sql count(), mongo count(), etc). can change it again if you def want 'length()'

packages/access-api/src/types/delegations.ts
gobengo: We don't *need* it. But right now an `Array` satisfies the `DelegationsStorage` type and getting rid of it would remove that property which, for sure, we don't need. (I started with an `Array` and expanded the type from there to afford for async and adding `Resolvable` was an easy way of doing that)

packages/access-api/src/models/delegations.js
gobengo: * couldn't do `await` when `length` was a property. but now its not so i'll adjust.
* will also make it a bigint native now that I'm making DelegationsStorage not be met by an `Array`

packages/access-api/src/models/delegations.js
gobengo: renamed to `count(): Promise<bigint>` in https://github.com/web3-storage/w3protocol/pull/427/commits/022dc2167647dcbabb35b6ab0450dafd9c708a84

packages/access-api/src/models/delegations.js
gobengo: made those changes in https://github.com/web3-storage/w3protocol/pull/427/commits/022dc2167647dcbabb35b6ab0450dafd9c708a84

packages/access-api/src/models/delegations.js
gobengo: made this change in . https://github.com/web3-storage/w3protocol/pull/427/commits/022dc2167647dcbabb35b6ab0450dafd9c708a84

packages/access-api/src/models/delegations.js
gobengo: Why not? Would it be sufficient to e.g. only refuse to do that in production or if the overall db size is very large.

I removed the non-streaming case for now, but it was needed for a WIP version of the `testDelegateThenClaim` test. https://github.com/web3-storage/w3protocol/actions/runs/4256209423/jobs/7404788157#step:7:48
But for now I've removed it and will just disable tests that relied on it (since further work #394 will lead to a better impl that doesn't need asyncIteration).

packages/access-api/src/service/access-delegate.js
gobengo: Why is that preferable to defining a type like this? This type structurally matches to `Ucanto.Failure`, and if it deviates, the type-checker will catch that.

A reason I would avoid subclassing here is that subclassing would be a concrete implementation that depended via prototype chain on a specific instance/version of `Ucanto.Failure`. I intentionally want to avoid that dependency on the entirety of `Ucanto.Failure`. A lot of generic methods in ucanto ecosystem operate on type `{ error: true }` or `API.Failure` from `@ucanto/interface` but that specific `Failure` class you implement also has e.g. a `describe` method that I don't want/need here. I avoided subclassing in order to apply https://en.wikipedia.org/wiki/Interface_segregation_principle

packages/access-api/src/service/access-delegate.js
gobengo: defining this as a generator function yields delegations as we enounter them, which affords for streaming them out.

You can trivially collect them into an array like:
```
const collected = Array.from(extractProvenDelegations(invocation))
```

but if I authored this logic as a non-generator, it wouldn't be as useful to anyone that wanted to stream or e.g. only extract the first N proven delegation.

packages/access-api/src/service/access-delegate.js
gobengo: There is no notion in the protocols implemented today about a storage provider, so I want to avoid mentioning that concept and committing to it in the API indefinitely https://github.com/web3-storage/w3protocol/search?q=provider

But part of the issue here is that the [protocol spec](https://github.com/web3-storage/specs/blob/main/w3-access.md#accessdelegate) doesn't define the error types. I don't think we necessarily should add error names for all of the distinct things that could happen in our implementation here, so much as add just enough error types as are needed to make the protocol useful to a client regardless of implementation. I like `InsufficientStorage` because it's more than enough here to let the user know that retrying the request wont work, and it's also vague enough that we can use it for other situations. That was why I did it the way I did, but I could have painted the bikeshed another color. 

@alanshaw Can you document a use case for needing to distinguish `MissingProvider` from `InsufficientStorage`?

@Gozala I know sometimes you have success/failure types in mind in your specs, and sometimes even can point to earlier implementations where you specified some. Do you have any thoughts on what the `Failure` type for `access/delegate` should be?

We can also iterate on the details of this without blocking this PR, since this is all new functionality and the PR as written now disables the functionality in prod, so we can make backward-incompatible changes

packages/access-api/src/types/delegations.ts
gobengo: Removed `Resolvable` in https://github.com/web3-storage/w3protocol/pull/427/commits/022dc2167647dcbabb35b6ab0450dafd9c708a84#diff-0c692d0f8036f645c9fe7c647d216265dedfb7a84decb52490918e43750a9e9eL3

packages/access-api/src/types/delegations.ts
gobengo: renamed to `putMany` https://github.com/web3-storage/w3protocol/pull/427/commits/022dc2167647dcbabb35b6ab0450dafd9c708a84#diff-0c692d0f8036f645c9fe7c647d216265dedfb7a84decb52490918e43750a9e9eR11

gobengo:
> I really quickly glanced at the pr and noticed that we’ll be pulling delegation as bytes from db, decode them into delegations which then ucanto will take and encode back into bytes. Decode/encode there seems pretty wasteful & perhaps can be avoided by sending bytes and let client decode

This is a good point and something we can optimize later. I think this only matters in the context of `access/claim` which I stubbed out here but plan to flesh out after this in #394 . I added some WIP `access/claim`-related code in here, but none of it is reachable from the `access-api` worker because the underlying service only adds `access.delegate` and not `access.claim`, so I don't think merging this will lead to the performance concern *in the wild*, so I suggest that it is an acceptable risk to handle later as part of #394

packages/access-api/src/service/access-delegate.js
alanshaw: > Why is that preferable to defining a type like this?

There is more than one place where we'll want this error - I think it will be less code overall to just import and use.

> I intentionally want to avoid that dependency on the entirety of that class.

It is tiny, and re-exported from `@ucanto/server` anyway which you're already including here.

packages/access-api/src/service/access-delegate.js
alanshaw: I don't think I saw a significant amount of work that the function was doing that would benefit from it being done incrementally.

When this function is being used, you immediately spread the results into a function call, which is not using the generator in a streaming fashion anyway.

If in the future you need to consume part of the stream, or it is blocking the event loop or you want to do some work between recieving each proof then sure, make it a generator. Right now though, this function reads to _me_ as though it does some heavy lifting or as a consumer I'm going to want/need to process the data incrementally, which doesn't appear to be the case.

packages/access-api/src/service/access-delegate.js
alanshaw: No feels the same, happy to go with `InsufficientStorage` for now.

packages/access-api/src/service/access-delegate.js
alanshaw: This is not blocking!

packages/access-api/src/service/access-delegate.js
alanshaw: This is not blocking!

packages/access-api/src/models/delegations.js
alanshaw: Hmm so at the time I didn't know that this was going to become a filtered list - in the case of all data then there's no point in allowing buffered response as it'll immediately become unusable.

For a specific DID's delegations then sure buffered response is fine.

packages/access-api/src/bindings.d.ts
alanshaw: OK gotcha - next time just add that info to the PR description so I know it's intentional!

alanshaw (approved):

packages/access-api/src/service/access-delegate.js
gobengo: > There is more than one place where we'll want this error - I think it will be less code overall to just import and use.

to reuse, import the type from `import('@web3-storage/capabilities/types').InsufficientStorage}` as is done here

I don't need a subclass or really any class here, so I'm not going to add one, but if anyone else wants to, they can use that import with `@implements` to ensure it will jive with everything here.

PR #430, opened 2023-02-09
fix: avoid email delegation via GET request

travis:
Originally proposed in #398, moved here to make rebasing simpler and to fix tests.

Original PR message from @natevw:

The email validation approval process is now split into two stages: a GET request with no side effects except to load a page, that then auto-submits a POST request to actually continue the flow.

## Summary of problem

This fixes the API so as to follow [proper HTTP semantics](https://github.com/web3-storage/w3protocol/issues/333#issuecomment-1380925535):

> The purpose of distinguishing between safe [i.e. like GET] and unsafe [like PUT/POST] methods is to allow automated retrieval processes (spiders) and cache performance optimization (pre-fetching) to work without fear of causing harm. In addition, it allows a user agent to apply appropriate constraints on the automated use of unsafe methods when processing potentially untrusted content.

That is, a `PUT` or `POST` (rather than a `GET`) **must** be the method used in order to do things like

* cause a message to be sent (forwarding a UCAN delegation via a separate connection's websocket)
* cause an untrusted keypair to be associated with a billable email address (which is the outcome of that forwarding, in practice!)

Fixing the HTTP semantics should address all of #348, and is the first step to addressing the security concerns in #333.

## Summary of solution

Clicking (or scanning/pre-fetching/previewing/etc.) the link in the email no longer finishes the validation process. Instead, it loads a (harmless to scan/pre-fetch/preview) landing page which simply says "Validating Email" while using JavaScript to auto-complete the process.

My preference would have been to move more of the approval process *out* of the email and *into* this landing page. (So rather than auto-approving, this landing page would contain details/context and force an informed clear "Yes, approve this new space" vs. "No, I didn't want this" decision.) However, the team preferred to keep\* all that in the initial email and requested that this fix be based on an auto-approval.

Given that preference, this patch is able to fix the core HTTP semantics very self-contained:

* no changes needed to the email templates (\* though still would be good to do)
* will not break any existing unexpired links at the moment it is deployed
* is essentially the exact same UX from a user's perspective (they might notice just a little extra blink)
* does degrade gracefully if user has JS disabled, and any non-browser clients could still trigger the `POST` ± just as easy as before
* no changes needed on the `w3ui` side for this part of the email validation improvements

### DIFF
--- a/packages/access-api/test/access-authorize.test.js
+++ b/packages/access-api/test/access-authorize.test.js
@@ -88,7 +88,7 @@ describe('access/authorize', function () {
       /** @type {import('@web3-storage/access/types').EncodedDelegation<[import('@web3-storage/capabilities/types').AccessSession]>} */ (
         url.searchParams.get('ucan')
       )
-    const rsp = await mf.dispatchFetch(url)
+    const rsp = await mf.dispatchFetch(url, { method: 'POST' })
     const html = await rsp.text()
 
     assert(html.includes(encoded))
@@ -119,7 +119,7 @@ describe('access/authorize', function () {
 
     const url = new URL(inv)
     // click email url
-    await mf.dispatchFetch(url)
+    await mf.dispatchFetch(url, { method: 'POST' })
 
     // ws
     const res = await mf.dispatchFetch('http://localhost:8787/validate-ws', {
--- a/packages/access-api/src/index.js
+++ b/packages/access-api/src/index.js
@@ -4,7 +4,7 @@ import { notFound } from '@web3-storage/worker-utils/response'
 import { Router } from '@web3-storage/worker-utils/router'
 import { postRaw } from './routes/raw.js'
 import { postRoot } from './routes/root.js'
-import { validateEmail } from './routes/validate-email.js'
+import { preValidateEmail, validateEmail } from './routes/validate-email.js'
 import { validateWS } from './routes/validate-ws.js'
 import { version } from './routes/version.js'
 import { getContext } from './utils/context.js'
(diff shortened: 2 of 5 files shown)


vasco-santos (approved):
Code looks good to me. Thanks @travis !  It would be quite nice to test it and see if before shipping.

Sadly, we don't have per PR deployments in CF, so we should make sure we test this in staging once this gets merged and are happy with it

packages/access-api/test/access-authorize.test.js
```diff
@@ -119,7 +119,7 @@ describe('access/authorize', function () {
 
     const url = new URL(inv)
     // click email url
-    await mf.dispatchFetch(url)
+    await mf.dispatchFetch(url, { method: 'POST' })
```
gobengo, line 122: Why does the comment say 'click email url', which I assume would usually involve an HTTP GET, but this changes it to 'POST'?

travis: right so this change splits the approval process into 3 pieces (from the 2 that it is now):
1) an email with a link 
2) a page that gets served on GET that uses JavaScript to redirect users to
3) a page that gets served on POST that does the actual verification and triggers the websocket data flow

here's the page that gets served by the GET request:

https://github.com/web3-storage/w3protocol/pull/430/files#diff-1a9cb8eda7a1020a96a7a7c0db5406ee04e5f8953fb6c00ab35f3bb8e6fb2519R103

travis:
> Sadly, we don't have per PR deployments in CF, so we should make sure we test this in staging once this gets merged and are happy with it

sweet - agreed! fwiw I did test this locally and it looked basically indistinguishable from the current process because the redirect process happens so quickly. 

I haven't tested in our staging environment before, is there a persistent env for that or do I need to do something to get that deployed once this is merged? I'll merge now and we can get this tested in staging today!

packages/access-api/test/access-authorize.test.js
travis: and for clarity, in this framing our current process is:

1) an email with a link
2) a page that gets served on GET that does the verification and triggers the websocket data flow

PR #432, opened 2023-02-09
feat: verify legitimate claim↔︎redeem by assigning a random phrase during email validation

travis:
Originally PR #399, moving here to merge main and make it easier to move forward.

Original message:

This branch fulfills the key piece of issue #347, adding a "validation phrase" that serves as a human-friendly nonce — enabling **definitive** verification of the validity of an email request.

(The other request metadata mentioned can help savvy users investigate any abuse further, but the phrase links one specific request to one specific email, completely out of the control of an attacker.)

## Problem summary

The idea here is to avoid malicious account validation: currently, Eve can generate a new space keypair and request delegation against Alice's email address. Alice may not be able to tell if the request is legitimate (or may follow the activation link simply out of curiousity/confusion) and grant Eve access!

## Solution provided

1. The requesting UI and the resulting validation emails will both show a short nonce phrase like `carat tender electronic tee guessed`. ~~[At the moment can preview phrases via <http://localhost:8787/phrase-test?bits=50> but that should likely get dropped before this PR is merged….]~~
2. The email will prompt something like "does this phrase match the one displayed in the app you are registering?"
3. The user can then check to make sure that the email validation they are approving matches the request they made themselves.

The `Agent.registerSpace` and `Agent.recover` methods in this repo's `@web3-storage/access/agent` package now take an (optional but strongly encouraged!) callback via `opts.handlePhrase` which will be called when the API provides a `match_phrase` in its response to each method's respective invocation. **The `w3ui` components and any protocol documentation for third-party clients will need updating to support this on the requestor side!**

So this is a *somewhat* breaking change, but:

* until the email templates are updated, this won't lead to user confusion
* even then, anyone testing can just ignore the nonce phrase and say "yes it matches"
* this overall strategy (plaintext phrase) should also be compatible with any CLI or screen reader usage

Note also that the phrase generation is done at an [MVP](https://www.agilealliance.org/glossary/mvp/) or proof-of-concept level. I've generated a small word back and left the `DEFAULT_ENTROPY` for `generateNoncePhrase` correspondingly low as well. Next steps here would be:

* it might not be a bad idea to increase the number of words so that the phrase entropy can be increased. (I'm not terribly worried about this though since iiuc we're really just fighting the chance of very lucky random collisions on a few malicious attempts — if not already emailing should be generally rate-limited so presumably an attacker would only get a few tries per email address anyway just to prevent spam/abuse generally)
* at least review the existing word list and simply drop any unwanted entries (potentially offensive, overly negative, weird characters/punctuation)
* rather than just removing unwanted words, marketing could come up with a whole set of desirable "on-brand" words for enhanced onboarding vibes ✨ 
* or, could just find an existing npm word list package (or whole nonce phrase generator!) rather than maintaining a wordlist in this repo

## TODO

Here:

- [x] phrase generation
- [x] return phrase back to client
- [x] ^^^ check that above actually is working?
- [x] show `{{ match_phrase }}` in HTML+plain emails
- [x] update `access-client` here to parse out response
- [ ] update templates in postmark itself (also catch up plaintext after #403 merged)

Over in UI:

- [x] update `w3ui` low-level keyring stuff to handle phrase
- [x] update `w3ui` high-level examples/components to display phrase

This is tracked in https://github.com/web3-storage/w3ui/issues/307, and isn't a total blocker since you can either wait to update the templates in Postmark and/or just have the user click through even without being able to see the phrase.


travis:
w3ui changes now pushed to https://github.com/web3-storage/w3ui/pull/385

packages/access-api/src/utils/phrase.js
```diff
@@ -0,0 +1,48 @@
...
+/**
+ *
+ * @param {number} max
+ * @returns {number}
+ */
+function randomInt(max) {
+  // NOTE: does not support all calling patterns of the real one!
+  const min = 0
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random#getting_a_random_integer_between_two_values
+  return Math.floor(Math.random() * (max - min) + min)
```
alanshaw, line 25: Use [`crypto.getRandomValues()`](https://developers.cloudflare.com/workers/runtime-apis/web-crypto/#methods).

packages/access-client/src/agent.js
```diff
@@ -322,6 +323,10 @@ export class Agent {
       throw new Error('Recover validation failed', { cause: inv })
     }
 
+    if (inv && 'match_phrase' in inv) {
+      opts?.handlePhrase?.(inv.match_phrase)
```
alanshaw, line 327: I typically I'd call this `onPhrase` to signify that it is an event - you'd maybe call your event handler `handlePhrase` but the option should be the event name (prefixed with "on").

We have prior art in other modules e.g. https://github.com/web3-storage/w3protocol/blob/main/packages/upload-client/src/types.ts#L214 so this would be consistent with them.

packages/capabilities/src/types.ts
```diff
@@ -38,6 +38,11 @@ export type StoreList = InferInvokedCapability<typeof list>
 // Top
 export type Top = InferInvokedCapability<typeof top>
 
+export type NoncePhrase = string
+export interface ShouldShowValidationNonce {
```
alanshaw, line 42: How come `ShouldShow`?

```suggestion
export interface ValidationNonce {
```

packages/access-api/src/utils/email.js
```diff
@@ -32,6 +32,7 @@ export class Email {
           product_name: 'Web3 Storage',
           email: opts.to,
           action_url: opts.url,
+          match_phrase: opts.nonce,
```
alanshaw, line 35: I think it's fine to have `match_phrase` in snake case here but in all other places we should be using camelCase to be consistent...

packages/access-client/src/types.ts
```diff
@@ -94,7 +96,9 @@ export interface Service {
   voucher: {
     claim: ServiceMethod<
       VoucherClaim,
-      EncodedDelegation<[VoucherRedeem]> | undefined,
+      | ShouldShowValidationNonce
+      | EncodedDelegation<[VoucherRedeem]>
```
alanshaw, line 100: Please can we remove this, and cast it in the tests to appease typescript? I'm not super happy to have types used only for testing in our public facing API.

packages/access-client/src/types.ts
```diff
@@ -103,7 +107,7 @@ export interface Service {
     info: ServiceMethod<SpaceInfo, SpaceRecord, Failure | SpaceUnknown>
     'recover-validation': ServiceMethod<
       SpaceRecoverValidation,
-      EncodedDelegation<[SpaceRecover]> | undefined,
+      ShouldShowValidationNonce | EncodedDelegation<[SpaceRecover]> | undefined,
```
alanshaw, line 110: Please can we remove this, and cast it in the tests to appease typescript? I'm not super happy to have types used only for testing in our public facing API.

packages/access-client/src/types.ts
```diff
@@ -236,6 +240,8 @@ export type DelegationOptions = SetRequired<UCANBasicOptions, 'audience'> & {
   audienceMeta: AgentMeta
 }
 
+export type ValidationPhraseHandler = (please_show: NoncePhrase) => undefined
```
alanshaw, line 243:
```suggestion
export type ValidationPhraseHandler = (phrase: NoncePhrase) => void
```

packages/access-client/src/types.ts
alanshaw, line 110:
```suggestion
      ShouldShowValidationNonce | undefined,
```

alanshaw (changes requested):

packages/access-client/src/types.ts
travis: hm is this suggestion right or did you mean to suggest removing `ShouldShowValidationNonce` ?

packages/access-client/src/types.ts
travis: ok so - I removed the `ShouldShowValidationNonce` interface from these types and it lead to errors like this:

https://github.com/web3-storage/w3protocol/actions/runs/4301261826/jobs/7498281518#step:6:13

is there a better way to pass information back from a handler the way we're trying to do here? Happy to go read the ucanto docs in more depth if it's explained there but if you have ideas I'd be grateful for pointers.

packages/access-client/src/agent.js
```diff
@@ -417,6 +422,7 @@ export class Agent {
    * @param {string} email
    * @param {object} [opts]
    * @param {AbortSignal} [opts.signal]
+   * @param {import('./types').ValidationPhraseHandler} [opts.onPhrase]
```
gobengo, line 425: I can't quite tell how I would decide whether I should provide a value for this and, if so, what a good value would be. Can you include a comment in the jsdoc and/or a test that makes it more clear why/how it's useful?

packages/access-client/src/types.ts
gobengo: > is there a better way to pass information back from a handler the way we're trying to do here? 

That's a good way to pass data back. But what I don't get is why you need to pass this data back in the `voucher/redeem` result. It looks like the only reason is so it can be passed to `opts.onPhrase`, but *what is that for* (and is there a better way that avoids changing the response)?

Related: Why do we generate the phrase on the server side instead of on the client side? If on the client side 1) the end-user has more control and e.g. we have less ability to MITM and 2) there wouldn't be a need to pass it back like this because the issuer of `voucher/claim`

Lastly, the Space/recover-validation protocol isn't really documented anywhere right? So idk I consider it unstable regardless so I'm not that worried about that interface compared to `voucher/redeem`. Even `voucher/redeem` is pretty undocumented at this point, and I @Gozala and I have been hoping to remove it alltogether after https://github.com/web3-storage/w3protocol/issues/459
(and we're going to have to reconsider *all of these emails* as we implement https://github.com/web3-storage/specs/pull/50/files#diff-8b85a91a07ee20b795a9db66585675ca26dd7afd02d2c8066380ac1961497548R146 )

travis: So there's a lot more context for this in https://github.com/web3-storage/w3protocol/pull/399 but tldr - one of the primary benefits of this feature is that it helps _our service_ more definitively verify that the user clicking the link in the email is the user who initiated the space registration flow by creating a secret that is generated by our service and sent to both the invoker and the email. 

I thiiiink that letting the client generate the phrase opens up a path to an attacker obtaining the `voucher/claim` capability and sending the user an email containing the correct phrase, which they'd then be inclined to click because it has the right information. By keeping this a secret that's only sent in the `access-api` response and the email we send to the user I do think we get additional useful security properties.

TBH though I'm not sure this whole flow was fully thought-through before Nate and Hugo left, and I'm increasingly convinced this implementation should be a proof-of-concept to be integrated into the changes coming out of https://github.com/web3-storage/specs/pull/50/files#diff-8b85a91a07ee20b795a9db66585675ca26dd7afd02d2c8066380ac1961497548R146 - I do think there's a place for some flavor of this but, eg, we may want to go with a model where we ask the user to input the phrase they see and check that against something we've stored before signing off on the relationship between an email and a space.

Let's sync asap on this - have been meaning to pick your brain on how this relates to the work you're doing!

packages/access-client/src/agent.js
travis: yep good idea - fwiw here's an example:

https://github.com/web3-storage/w3ui/pull/385/files#diff-5ba6f9aada6d8184e7e8ee2fb08e2e6bfa67021b095c9a5603efd7b52b1dc3eeR120

packages/access-client/src/types.ts
travis: ok chatted out of sync I misunderstood and this is about `EncodedDelegation<[SpaceRecover]>` - I'm going to work on this in a separate PR

packages/access-client/src/types.ts
gobengo: > I thiiiink that letting the client generate the phrase opens up a path to an attacker obtaining the voucher/claim capability and sending the user an email containing the correct phrase, which they'd then be inclined to click because it has the right information. By keeping this a secret that's only sent in the access-api response and the email we send to the user I do think we get additional useful security properties.

this makes sense to me. thanks for explaining it to me :)

Issue #395, opened 2023-01-25
Implement w3access and w3session in `access-client`


alanshaw:
* https://github.com/web3-storage/specs/blob/feat/w3-authorization/w3-access.md
* https://github.com/web3-storage/specs/blob/feat/w3-authorization/w3-session.md

gobengo:
@alanshaw are you blocked on this until resolution of 
* https://github.com/web3-storage/w3protocol/issues/393
* https://github.com/web3-storage/w3protocol/issues/394

?

If so, is there one of those that would be most helpful to do focus on before the other?

gobengo:
ok. I'm still diving in, but I think you might want `access/delegate` implemented to fit into [`AccessClient#registerSpace(email)`](https://github.com/web3-storage/w3protocol/blob/64da6e50c4d0d0ef3b78c00298769665463e421d/packages/access-client/src/agent.js#L429) if we remove the `voucher/claim` handlers that it currently uses, which @hugomrdias and I briefly discussed doing today, but obv we won't do any sooner than until #393 is resolved

gobengo:
started PR over here that would unblock sending `access/delegate` invocation without error https://github.com/web3-storage/w3protocol/pull/427

gobengo:
use case: https://github.com/web3-storage/w3cli/issues/46

PR #433, opened 2023-02-09
feat!: implement new account-based multi-device flow

alanshaw:
With this PR we're able to use two different devices on behalf of a single account identified by an email address.

An agent (ie, a device like w3console or w3cli) can now:

1) use `access/authorize` to trigger an email verification flow that will give them delegations to act on behalf of an account 
2) create a space locally
3) add a storage provider to that space with `provider/add`
4) delegate capabilities to the account they are authorized as that permit the account to delegate all capabilities on those spaces to other agents - in other words, create spaces and assign all "permissions" on those spaces to their account
5) upload data to the space

A second agent (ie, another device) can then:
1) use `access/authorize` to trigger an email verification flow that will give them delegations to act on behalf of the same account
2) get a list of spaces they can store data in, which includes the space created on the first device
3) upload data to the space

This PR also contains various refactoring of the `Agent` class to minimize its responsibilities and move in the direction of letting user agents take responsibility for state storage.

refs https://github.com/web3-storage/w3protocol/issues/395

* [x] setup tests for access-client agent + access-api
* [x] simple test agent createSpace
* [x] @gobengo test agent authorize happy path https://github.com/web3-storage/w3protocol/pull/535
* [x] @gobengo upgrade to ucanto 6.2 https://github.com/web3-storage/w3protocol/pull/541
* [x] @travis ensure what's proposed here can work in w3up-client, w3ui, w3console
* [x] upgrade this branch to `@ucanto/transport@5.1.1` after https://github.com/web3-storage/ucanto/issues/261
* [x] minimize new public api surface area on access-client Agent
  * [x] (e.g. `sessionProof`) https://github.com/web3-storage/w3protocol/pull/545/files
  * [x] `sessionPrincipal` https://github.com/web3-storage/w3protocol/pull/546
* [x] review comments
* [x] `authorize` should access/claim `with=did:mailto:...` https://github.com/web3-storage/w3protocol/pull/556/files#


packages/access-client/src/agent.js
```diff
@@ -404,6 +421,43 @@ export class Agent {
...
+    if (res?.error) {
+      throw new Error('failed to authorize session', { cause: res })
+    }
+
+    const sessionDelegation =
+      /** @type {Ucanto.Delegation<[import('./types').AccessSession]>} */
+      (await this.#waitForDelegation(opts))
+
+    const cap = sessionDelegation.capabilities.find(c => c.can === Access.session.can && c.nb.key === this.issuer.did())
+    if (!cap && isExpired(sessionDelegation)) {
```
gobengo, line 452:
```suggestion
    if (!cap || isExpired(sessionDelegation)) {
```

packages/access-client/src/agent.js
```diff
@@ -404,6 +421,43 @@ export class Agent {
...
+      throw new Error('received invalid delegation')
+    }
+
+    await this.addProof(sessionDelegation)
+
+    // TODO: claim delegations
```
gobengo, line 458: * [ ] https://github.com/gobengo/w3protocol-test/pull/4

packages/access-client/src/agent.js
```diff
@@ -409,6 +480,115 @@ export class Agent {
...
+    const res = await this.invokeAndExecute(Access.claim, {
+      audience: this.connection.id,
+      with: this.issuer.did(),
+    })
+
+    if (res.error) {
+      throw new Error('error claiming delegations')
+    }
+    const delegations = Object.values(res.delegations).flatMap((bytes) =>
+      bytesToDelegations(bytes)
```
gobengo, line 532: I think you can import this from `@web3-storage/access/encoding` (i.e. this package)

gobengo: (I couldn't in observable due to import error)

packages/access-client/src/agent-data.js
```diff
@@ -151,11 +151,10 @@ export class AgentData {
    */
   sessionProof() {
     for (const { delegation } of this.delegations.values()) {
-      // @ts-expect-error "key" does not exist in object, unless it's a session capability
       const cap = delegation.capabilities.find(
         (c) =>
           // @ts-expect-error "key" does not exist in object, unless it's a session capability
-          c.can === Access.session.can && c.nb?.key === this.principal.did()
+          c.can === Access.session.can && c?.nb?.key === this.principal.did()
```
travis, line 157: if `c` is undefined I think it will already blow up earlier here right (in the c.can === Access.session.can)

packages/access-client/src/agent-data.js
```diff
@@ -143,4 +145,18 @@ export class AgentData {
     this.delegations.delete(cid.toString())
     await this.#save(this.export())
   }
+
+  /**
+   * The current session proof.
+   */
+  sessionProof() {
```
gobengo, line 152: This is written in a way that seems to imply there will be a zero or one 'session proof' that can be determine at all times, which I think may be an invalid assumption. The session proof you need is a function of the action your atrying to take.

packages/access-client/src/agent-data.js
gobengo: I removed the `?` I added.

packages/access-client/src/agent.js
```diff
@@ -520,11 +520,19 @@ export class Agent {
   }
 
   async claimDelegations() {
-    const res = await this.invokeAndExecute(Access.claim, {
-      audience: this.connection.id,
-      with: this.issuer.did(),
-    })
-
+    const [res] = await this.connection.execute(
```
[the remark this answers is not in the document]
gobengo: ok https://github.com/web3-storage/w3protocol/pull/433/commits/75e243fbd49e26671b3606ac148f09cb96ebc410#diff-225f337b92bbab33fa4af35cf1e7543a70b90531bf36c394e816730282058babR339

packages/access-client/src/agent-data.js
```diff
@@ -152,9 +168,7 @@ export class AgentData {
   sessionProof() {
     for (const { delegation } of this.delegations.values()) {
       const cap = delegation.capabilities.find(
-        (c) =>
-          // @ts-expect-error "key" does not exist in object, unless it's a session capability
-          c.can === Access.session.can && c.nb?.key === this.principal.did()
+        (c) => c.can === Access.session.can // TODO we should make sure this is the current session proof - we were checking nb.key but that doesn't seem to exist in the staging ucan/attest at the moment
```
gobengo, line 171: it's also possible to do something like `Access.session.match(...).is(...)` and it's a type guard in a nice way, and it might be useful here (or not)

packages/capabilities/test/capabilities/provider.test.js
```diff
@@ -18,7 +18,7 @@ describe('provider/add', function () {
       audience: service,
       with: account,
       nb: {
-        provider: service.did(),
+        provider: 'did:web:test.web3.storage',
```
Gozala, line 21: I'm not sure I understand rational here but I have to say I liked the version before.

packages/access-client/src/agent-data.js
Gozala: Put it differently you'll have a set of attestations (sessions) that which have `capabilities[0].nb.proof` CID corresponding to specific delegation. Idea is to send delegation and a session together, so perhaps better api would be to a function that takes `delegation` and then finds `session` where `session.capabilities[0].proof.toString() === delegation.cid.toString()`

packages/access-client/src/agent.js
```diff
@@ -28,6 +32,59 @@ export { AgentData }
...
 const PRINCIPAL = DID.parse('did:web:web3.storage')
 
+/**
+ *
+ * @param {string} email
+ * @returns {Ucanto.Principal<Ucanto.DID<'mailto'>>}
+ */
+function emailToSessionPrincipal(email) {
+  const parts = email.split('@').map((s) => encodeURIComponent(s))
+  return DID.parse(`did:mailto:${parts[1]}:${parts[0]}`)
```
Gozala, line 42: Might be a good idea to verify that you have two parts and non are empty. I remember seeing cases where that was not the case.

packages/access-client/src/agent.js
```diff
@@ -28,6 +32,59 @@ export { AgentData }
...
+/**
+ * @param {Ucanto.Signer<Ucanto.DID<'key'>>} space
+ * @param {Ucanto.Principal<Ucanto.DID<'mailto'>>} account
+ * @returns
+ */
+async function createSpaceSaysAccountCanAdminSpace(space, account) {
+  return ucanto.delegate({
+    issuer: space,
+    audience: account,
+    capabilities: [
```
Gozala, line 54: I think it's best to delegate `{ with: "ucan:*", can: "*" }` to an account so that account has complete authority over the space not just over the capabilities that we use today.

Gozala (commented):
I have stopped reviewing this because I think it's out of date on things

packages/capabilities/test/capabilities/provider.test.js
gobengo: `service` can be `did:key` so didn't match the type.
Do you want me to allow `did:key` for providers?

packages/access-client/src/delegations.js
```diff
@@ -62,12 +65,16 @@ export function validate(delegation, opts) {
  * @param {import('@ucanto/interface').Capability} child
  */
 export function canDelegateCapability(delegation, child) {
-  for (const parent of delegation.capabilities) {
-    if (
-      parent.with === child.with &&
-      canDelegateAbility(parent.can, child.can)
-    ) {
-      return true
+  // @ts-ignore ignore this until we bring in ucanto as a normal dependency rather than linking to a local version
```
gobengo: https://github.com/web3-storage/ucanto/pull/260

packages/access-client/src/agent-data.js
gobengo: Maybe we could make this `sessionProof` `protected` for now and that way we can merge this and iterate on the api. @travis do you need this to be `public`? i.e. do you call it directly from w3up-client et al?

packages/access-client/src/agent-data.js
```diff
@@ -74,6 +79,10 @@ export class AgentData {
         meta: raw.meta,
         // @ts-expect-error for some reason TS thinks this is a EdSigner
         principal: Signer.from(raw.principal),
+        // @ts-expect-error TODO figure out the types for this too
```
gobengo, line 82: TODO

packages/access-client/src/agent-data.js
travis: don't need to call it from w3up-client nope, let's make it `protected`! I think this was just something @alanshaw threw in provisionally so definitely feel free to change it

packages/access-client/src/agent-data.js
```diff
@@ -124,6 +134,14 @@ export class AgentData {
     await this.#save(this.export())
   }
 
+  /**
```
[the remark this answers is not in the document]
gobengo: yeah I'm worried about committing to a public setter like this that seems to imply there is only one session principal at a time.

Can this be non-public method @travis ?

packages/access-client/src/agent.js
gobengo: idea: we should have a `@web3-storage/did-mailto` package in this monorepo with e.g. canonical type for this and maybe also a ucanto validator schema that matches the type.

packages/access-client/src/agent.js
gobengo: seems like a good idea to me. @travis wdyt?

packages/access-client/src/agent.js
```diff
@@ -191,19 +253,36 @@ export class Agent {
...
    *
-   * Proofs are delegations with an audience matching agent DID.
+   * Proof of session will also be included in the returned proofs if any
+   * proofs matching the passed capabilities require it.
    *
    * @param {import('@ucanto/interface').Capability[]} [caps] - Capabilities to filter by. Empty or undefined caps with return all the proofs.
    */
   proofs(caps) {
     const arr = []
+    const session = this.#data.sessionProof()
```
gobengo, line 268: seems like we could accomodate multiple sessionProofs here. We'd just need to index them by `.audience.did()`.
Might also be good to pass `caps` to sessionProof to it can omit any that aren't relevant.

packages/capabilities/test/capabilities/provider.test.js
Gozala: I think service should not be a `did:key` and we probably should fix it in cases where it is ? That said I'm ok with things as is

gobengo: ok in another PR maybe I will change the type of the `service` fixture to always only be `did:web`. I avoided that here because I didn't want to revisit/update all the tests that rely on that fixture.

packages/access-client/src/agent.js
gobengo: I removed `sessionProof` from the public api so idc what we do here for now

packages/access-client/src/agent.js
travis: sounds right to me!

packages/access-client/src/agent-data.js
```diff
@@ -17,6 +20,7 @@ export class AgentData {
   constructor(data, options = {}) {
     this.meta = data.meta
     this.principal = data.principal
+    this.sessionPrincipal = data.sessionPrincipal
```
gobengo, line 23: I think we should avoid storing a single `sessionPrincipal` on any of this agent stuff.
A single sessionPrincipal may make sense in a specific application context, but that application should keep track of it itself and e.g. pass it in to things like `addProvider`

packages/access-client/src/agent.js
gobengo: I'm worried if I commit this, it will lead to new errors when claiming expired delegations.
might wanna do this first https://github.com/web3-storage/w3protocol/issues/526

packages/access-client/src/agent-data.js
gobengo: removed `sessionProof` in https://github.com/web3-storage/w3protocol/pull/545/files

packages/access-client/src/agent-data.js
gobengo: https://github.com/web3-storage/w3protocol/pull/546

packages/access-client/src/agent-data.js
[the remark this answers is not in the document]
gobengo: I removed this method

packages/access-client/src/agent.js
gobengo: https://github.com/web3-storage/w3protocol/pull/433/commits/0cc4e47fd8383927393c32e7d06fc5e56f0fe10a

packages/access-client/src/agent.js
gobengo: This function isn't the right place to do `ucan:*`, but I did change this to
* allow `capabilities` to be passed as param
* default them to `can="*"` instead of the specific cans used before

https://github.com/web3-storage/w3protocol/pull/433/commits/1ea805e12b5f9ffcedca6e927a41ff58b7c85f28

(I don't disagree that it prob makes sense to do `{ with: "ucan:*", can: "*" }` in delegation from space->account more generally, but this function is named in a way that makes caller think `with=space`

packages/access-client/src/agent.js
```diff
@@ -410,78 +482,191 @@ export class Agent {
...
+    }
+    const delegations = Object.values(res.delegations).flatMap((bytes) =>
+      bytesToDelegations(bytes)
+    )
+    for (const delegation of delegations) {
+      this.addProof(delegation)
+
+      // if we can find a store/* capability in this delegation, look in the proofs
+      // for the concrete capabilities where space DIDs will be specified
+      // TODO: this was my first attempt at inferring spaces from claimed delegations, but I think it needs work - tv
```
gobengo, line 545: resolve or remove TODO

packages/access-client/src/agent.js
```diff
@@ -410,78 +482,191 @@ export class Agent {
...
-      )
-    await this.addProof(voucherRedeem)
-    const delegationToService = await this.delegate({
-      abilities: ['*'],
-      audience: service,
-      expiration: Infinity,
-      audienceMeta: {
-        name: 'w3access',
-        type: 'service',
+    // TODO: should we be inferring which spaces we have access to here and updating local space state?
```
gobengo, line 563: resolve or remove TODO

packages/access-client/src/agent.js
```diff
@@ -669,6 +670,87 @@ export class Agent {
...
+          agent: this.meta,
+        },
+      ],
+    })
+
+    if (accInv && accInv.error) {
+      throw new Error('Space registration failed', { cause: accInv })
+    }
+
+    spaceMeta.isRegistered = true
```
gobengo: note that this being absent or falsy doesn't mean that it's not registered.
You can know for sure by attempting to invoke `space/info`, which will error if the space has no storage provider added.

packages/access-client/src/agent.js
travis: need to resolve

packages/access-client/src/agent-data.js
```diff
@@ -144,3 +147,18 @@ export class AgentData {
...
+/**
+ * get session proof
+ *
+ * @param {AgentData} data
+ * @returns {Ucanto.Delegation | undefined}
+ */
+export function getSessionProof(data) {
+  for (const { delegation } of data.delegations.values()) {
+    const cap = delegation.capabilities.find(
+      (c) => c.can === Access.session.can // TODO we should make sure this is the current session proof - we were checking nb.key but that doesn't seem to exist in the staging ucan/attest at the moment
```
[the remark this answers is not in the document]
travis: hm up to you - I don't know what the semantics of this sessionProof stuff should be

packages/access-client/src/agent.js
travis: need to resolve

packages/access-client/src/agent-data.js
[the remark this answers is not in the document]
gobengo: I’ll take a look

packages/access-client/src/agent.js
```diff
@@ -409,6 +482,211 @@ export class Agent {
...
+   * Invokes voucher/redeem for the free tier, wait on the websocket for the voucher/claim and invokes it
+   *
+   * It also adds a full space delegation to the service in the voucher/claim invocation to allow for recovery
+   *
+   * @param {string} email
+   * @param {object} [opts]
+   * @param {AbortSignal} [opts.signal]
+   * @param {Ucanto.DID<'web'>} [opts.provider] - provider to register - defaults to this.connection.id
+   */
+  async #newRegisterSpace(email, opts) {
```
gobengo, line 648: we should make this `private`

gobengo: oh it is :P `#`

packages/access-client/src/agent-data.js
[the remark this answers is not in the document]
gobengo: travis fixed it!

packages/access-client/src/agent-use-cases.js
```diff
@@ -0,0 +1,157 @@
+import { Agent as AccessAgent, agentToData } from './agent.js'
```
gobengo, line 1: The inspiration for calling these 'use cases' is https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html#use-cases

packages/access-client/src/agent.js
```diff
@@ -352,7 +431,7 @@ export class Agent {
    *
    * Other methods will default to use the current space if no resource is defined
    *
-   * @param {Ucanto.DID} space
+   * @param {Ucanto.DID<'key'>} space
    */
   async setCurrentSpace(space) {
```
gobengo, line 436: I want to mark this private and/or deprecate it

packages/access-client/src/agent.js
```diff
@@ -409,6 +487,185 @@ export class Agent {
...
+  /**
+   * Request authorization of a session allowing this agent to issue UCANs
+   * signed by the passed email address.
+   *
+   * @param {`${string}@${string}`} email
+   * @param {object} [opts]
+   * @param {AbortSignal} [opts.signal]
+   * @param {Iterable<{ can: Ucanto.Ability }>} [opts.capabilities]
+   */
+  async authorize(email, opts) {
```
gobengo, line 499: We should make this a use-case fn that operates on agent.

We can do it if we do the same for `this.#waitForDelegation`

packages/access-client/src/agent.js
```diff
@@ -417,9 +674,16 @@ export class Agent {
...
    * @param {AbortSignal} [opts.signal]
+   * @param {Ucanto.DID<'key'>} [opts.space] - space to register
+   * @param {Ucanto.DID<'web'>} [opts.provider] - provider to register - defaults to this.connection.id
    */
-  async registerSpace(email, opts) {
-    const space = this.currentSpace()
+  async registerSpace(email, opts = {}) {
+    // if the client passes `provider` use the new space registration flow
+    if (opts.provider) {
+      return this.#newRegisterSpace(email, opts)
```
gobengo, line 683: Irakli had proposed not doing new functionality here via option.
I'd be happy if we just made the new stuff usable via 'use case' fn(agent, ...)

packages/access-client/src/agent.js
```diff
@@ -491,7 +755,7 @@ export class Agent {
    */
   async #waitForDelegation(opts) {
```
gobengo, line 756: We can make this a function(agent) since it doesn't rely on any private api afaict

packages/access-client/src/agent.js
```diff
@@ -642,23 +907,28 @@ export class Agent {
   async invoke(cap, options) {
     const space = options.with || this.currentSpace()
```
gobengo, line 908: long term it feels weird that a generic `invoke` here would try to be aware of currentSpace (because it could be invoking something unrelated to spaces)

gobengo: Sending extra proofs related to spaces could leak information about the end-user they don't expect

packages/access-client/src/agent.js
gobengo: https://github.com/web3-storage/w3protocol/pull/563

packages/access-client/src/agent.js
gobengo: https://github.com/web3-storage/w3protocol/pull/563

packages/access-client/src/agent.js
gobengo: not in this pr. I think we can do this PR as a non breaking change

packages/access-client/src/agent.js
```diff
@@ -409,6 +481,101 @@ export class Agent {
...
+        delegations: {
+          [issuerSaysAccountCanAdminSpace.cid.toString()]:
+            issuerSaysAccountCanAdminSpace.cid,
+        },
+      },
+      proofs: [
+        await createIssuerSaysDeviceCanAccessDelegateWithSpace(
+          this.issuer,
+          space,
+          this.issuer
```
Gozala, line 521: This delegation makes no sense. Issuer delegates access to space to itself. If it already has access it does not need this delegation. If it does not have access than it should have no authority over space to be able to delegate it in first place.

I suspect all of this only because the space and the issuer are the same so this proof is never utilized.

packages/access-client/src/agent.js
```diff
@@ -409,6 +481,101 @@ export class Agent {
...
+   *
+   * It also adds a full space delegation to the service in the voucher/claim invocation to allow for recovery
+   *
+   * @param {string} email
+   * @param {object} [opts]
+   * @param {AbortSignal} [opts.signal]
+   * @param {Ucanto.DID<'key'>} [opts.space]
+   * @param {Ucanto.DID<'web'>} [opts.provider] - provider to register - defaults to this.connection.id
+   */
+  async #newRegisterSpace(email, opts) {
```
Gozala, line 540: > Opinion alert!

Nit: I think private methods make no sense they could be separate static methods instead so not part of the interface yet usable in tests etc..

packages/access-client/src/agent.js
```diff
@@ -409,6 +481,101 @@ export class Agent {
...
+          // connection.id did is a valid provider value. Try using that.
+          return service
+        }
+        throw new Error(
+          `unable to determine provider to use to register space. Pass opts.provider`
+        )
+      })()
+
+    if (!space || !spaceMeta) {
+      throw new Error('No space selected')
```
Gozala, line 557: Nit: this is misleading. I pass `space` but you could not find a `spaceMeta` it would tell me no space selected, instead it should probably say it has no data for that space ? Not sure what is in the data so it's hard for me to tell what exactly that means.

In other words it would be good to have an error that gives some actionable feedback on what user can do.

packages/access-client/src/agent.js
```diff
@@ -409,6 +481,101 @@ export class Agent {
...
+    const provider =
+      opts?.provider ||
+      (() => {
+        const service = this.connection.id.did()
+        if (DIDValidator.match({ method: 'web' }).is(service)) {
+          // connection.id did is a valid provider value. Try using that.
+          return service
+        }
+        throw new Error(
+          `unable to determine provider to use to register space. Pass opts.provider`
```
Gozala, line 552: I think it would be a good idea to be more explicit and say that it tried to derive provider from the connection but it isn't did:web so perhaps there something wrong with connection or user could specify provider

packages/access-client/src/agent.js
```diff
@@ -642,23 +763,28 @@ export class Agent {
...
     const space = options.with || this.currentSpace()
     if (!space) {
-      throw new Error('No space selected, you need pass a resource.')
+      throw new Error(
+        'No space or resource selected, you need pass a resource.'
+      )
     }
 
+    const extraProofs = options.proofs || []
     const proofs = this.proofs([
```
Gozala, line 777: Nit: Maybe add a comment that it select matching stored proofs

packages/access-client/src/agent.js
```diff
@@ -642,23 +763,28 @@ export class Agent {
   async invoke(cap, options) {
     const space = options.with || this.currentSpace()
```
Gozala, line 769: This probably should be `cap.with` instead

Gozala: Long term we want to do something like this

```js
const { delegation } = await Validator.claim(cap, proofs, { ... })
```

packages/access-client/src/agent.js
```diff
@@ -642,23 +763,28 @@ export class Agent {
...
+      )
     }
 
+    const extraProofs = options.proofs || []
     const proofs = this.proofs([
       {
         with: space,
         can: cap.can,
       },
     ])
```
Gozala, line 782:
```suggestion
    const proofs = [
      ...(options.proofs || []),
      ...this.proofs([
      {
        with: space,
        can: cap.can,
      },
    ])
 ]
```
Might be easier to read

packages/access-client/src/agent.js
Gozala, line 1: We should error on sending more proofs because chances are we will have an invalid proof chain which will prevent using the client.

packages/access-client/src/utils/did-mailto.js
```diff
@@ -0,0 +1,15 @@
...
+ * @returns {`did:mailto:${string}:${string}`}
+ */
+export function createDidMailtoFromEmail(email) {
+  const emailParts = email.split('@')
+  if (emailParts.length !== 2) {
+    throw new Error(`unexpected email ${email}`)
+  }
+  const [local, domain] = emailParts
+  const did = /** @type {const} */ (
+    `did:mailto:${encodeURIComponent(domain)}:${encodeURIComponent(local)}`
```
Gozala, line 12: Nit: We should check that neither part is empty

Gozala (approved):

packages/access-api/src/routes/validate-email.js
```diff
@@ -152,7 +152,7 @@ async function authorize(req, env) {
 
     if (confirmation.error) {
       throw new Error(`unable to validate access session: ${confirmation}`, {
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-api/src/routes/validate-email.js
```diff
@@ -168,12 +168,12 @@ async function authorize(req, env) {
       }
     )
     const confirmResult = await confirm(request, {
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-api/src/routes/validate-email.js
```diff
@@ -193,6 +193,8 @@ async function authorize(req, env) {
       )
     )
   } catch (error) {
+    // eslint-disable-next-line no-console
+    console.warn('error in validate-email', error)
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-api/src/service/access-authorize.js
```diff
@@ -25,7 +24,7 @@ export function accessAuthorizeProvider(ctx) {
     const confirmation = await Access.confirm
       .invoke({
         issuer: ctx.signer,
-        audience: DID.parse(capability.nb.iss),
+        audience: ctx.signer,
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-api/src/service/access-authorize.js
```diff
@@ -35,8 +34,9 @@ export function accessAuthorizeProvider(ctx) {
         nb: {
           // we copy request details and set the `aud` field to the agent DID
           // that requested the authorization.
-          ...capability.nb,
+          iss: capability.nb.iss,
           aud: capability.with,
+          att: capability.nb.att,
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-api/src/service/access-confirm.js
```diff
@@ -18,7 +19,9 @@ export function parse(invocation) {
   const capability = invocation.capabilities[0]
   // Create a absentee signer for the account that authorized the delegation
   const account = Absentee.from({ id: capability.nb.iss })
-  const agent = Verifier.parse(capability.nb.aud)
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-api/src/service/access-confirm.js
```diff
@@ -89,3 +82,40 @@ export async function handleAccessConfirm(invocation, ctx) {
...
+/**
+ * @param {Ucanto.Signer} service
+ * @param {Ucanto.Principal<Ucanto.DID<'mailto'>>} account
+ * @param {Ucanto.Principal<Ucanto.DID<'key'>>} agent
+ * @param {Ucanto.Capabilities} capabilities
+ * @param {AsyncIterable<Ucanto.Delegation>} delegationProofs
+ * @param {number} expiration
+ * @returns {Promise<[delegation: Ucanto.Delegation, attestation: Ucanto.Delegation]>}
+ */
+export async function createSessionProofs(
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-api/test/access-client-agent.test.js
```diff
@@ -0,0 +1,581 @@
...
+          registerSpaces: [spaceWithStorageProvider],
+        }),
+      }
+    })(),
+  },
+])) {
+  describe(`access-client-agent ${accessApiVariant.name}`, () => {
+    it('can createSpace', async () => {
+      const { connection } = await accessApiVariant.create()
+      const accessAgent = await AccessAgent.create(undefined, {
```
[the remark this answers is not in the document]
gobengo: I agree but didn't want to change API here

packages/access-client/src/agent-data.js
```diff
@@ -144,3 +149,43 @@ export class AgentData {
...
+const isSessionCapability = (cap) => cap.can === Access.session.can
+
+/**
+ * Is the given delegation a session proof?
+ *
+ * @param {Ucanto.Delegation} delegation
+ * @returns {delegation is Ucanto.Delegation<[import('./types').AccessSession]>}
+ */
+export const isSessionProof = (delegation) =>
+  delegation.capabilities.some((cap) => isSessionCapability(cap))
```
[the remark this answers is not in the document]
gobengo: there is a lint rule that requires not doing that (which we will remove later #479 )

packages/access-client/src/agent-use-cases.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Request authorization of a session allowing this agent to issue UCANs
+ * signed by the passed email address.
+ *
+ * @param {AccessAgent} access
+ * @param {Ucanto.Principal<Ucanto.DID<'mailto'>>} account
+ * @param {Iterable<{ can: Ucanto.Ability }>} capabilities
+ */
+export async function requestAuthorization(access, account, capabilities) {
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-client/src/agent-use-cases.js
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * claim delegations delegated to an audience
+ *
+ * @param {AccessAgent} access
+ * @param {Ucanto.DID} [delegee] - audience of claimed delegations. defaults to access.connection.id.did()
+ * @param {object} options
+ * @param {boolean} [options.addProofs] - whether to addProof to access agent
+ * @returns
+ */
+export async function claimDelegations(
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-client/src/agent-use-cases.js
```diff
@@ -0,0 +1,157 @@
...
+  access,
+  delegee = access.connection.id.did(),
```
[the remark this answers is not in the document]
gobengo: I avoid `agent` beacuse it's not very descriptive

packages/access-client/src/agent-use-cases.js
```diff
@@ -0,0 +1,157 @@
...
+  { addProofs = false } = {}
+) {
+  const res = await access.invokeAndExecute(Access.claim, {
+    audience: access.connection.id,
+    with: delegee,
+  })
+  if (res.error) {
+    throw new Error('error claiming delegations')
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-client/src/agent-use-cases.js
```diff
@@ -0,0 +1,157 @@
...
+export async function addSpacesFromDelegations(access, delegations) {
+  const data = agentToData.get(access)
+  if (!data) {
+    throw Object.assign(new Error(`cannot determine AgentData for Agent`), {
+      agent: access,
+    })
+  }
+  if (delegations.length > 0) {
+    const allows = Delegation.allows(delegations[0], ...delegations.slice(1))
+    for (const [did, value] of Object.entries(allows)) {
```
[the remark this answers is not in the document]
gobengo: @travis

packages/access-client/src/agent-use-cases.js
```diff
@@ -0,0 +1,157 @@
...
+
+  if (addProofs) {
+    await addSpacesFromDelegations(access, delegations)
+  }
+
+  return delegations
+}
+
+/**
+ * @private
```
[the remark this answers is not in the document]
gobengo: I will delete export
@gobengo

packages/access-client/src/agent.js
```diff
@@ -21,13 +23,50 @@ import {
...
 
-export { AgentData }
+export { AgentData, createDidMailtoFromEmail }
+export * from './agent-use-cases.js'
 
 const HOST = 'https://access.web3.storage'
 const PRINCIPAL = DID.parse('did:web:web3.storage')
 
+/** @type {WeakMap<Agent, AgentData>} */
+export const agentToData = new WeakMap()
```
gobengo, line 37: Explain why to use this.
Explain long term plan.
Avoid exporting it if possible.

packages/access-client/src/agent-use-cases.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * @param {AccessAgent} access
+ * @param {Ucanto.DID} delegee
+ * @param {object} [options]
+ * @param {number} [options.interval]
+ * @param {AbortSignal} [options.abort]
+ * @returns {Promise<Iterable<Ucanto.Delegation>>}
+ */
+export async function expectNewClaimableDelegations(access, delegee, options) {
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-client/src/agent.js
gobengo: @travis

packages/access-client/src/agent.js
```diff
@@ -409,6 +481,101 @@ export class Agent {
     }
   }
 
+  /**
+   * @param {Ucanto.DID<'key'>} space
+   * @param {Ucanto.Principal<Ucanto.DID<'mailto'>>} account
+   * @param {Ucanto.DID<'web'>} provider - e.g. 'did:web:staging.web3.storage'
+   */
+  async #addProvider(space, account, provider) {
```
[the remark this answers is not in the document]
gobengo: @gobengo

packages/access-client/src/agent.js
gobengo: @gobengo

packages/access-client/src/agent.js
```diff
@@ -409,6 +459,96 @@ export class Agent {
...
+    await this.#addProvider(space, account, provider)
+    const delegateSpaceAccessResult = await this.#delegateSpaceAccessToAccount(
+      space,
+      account
+    )
+    if (delegateSpaceAccessResult.error) {
+      // @ts-ignore it's very weird that this is throwing an error but line 692 above does not - ignore for now
+      throw new Error(delegateSpaceAccessResult.message, {
+        cause: delegateSpaceAccessResult,
+      })
```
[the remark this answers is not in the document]
gobengo: I'll verify but I think the linter wants it to be instanceof Error

@gobengo

packages/access-client/src/agent.js
gobengo: @travis

packages/access-client/src/agent.js
```diff
@@ -642,23 +768,28 @@ export class Agent {
...
         with: space,
         can: cap.can,
       },
     ])
 
-    if (proofs.length === 0 && options.with !== this.did()) {
+    if (
+      proofs.length === 0 &&
+      options.with !== this.did() &&
+      extraProofs.length === 0
```
gobengo, line 787: remove this if possible

packages/access-client/src/utils/ws.js
```diff
@@ -59,6 +63,10 @@ export class Websocket {
     this.ws = ws
 
     ws.addEventListener('close', (event) => {
+      if (event.code === 1006) {
```
gobengo, line 66: comment or remove

packages/access-api/src/routes/validate-email.js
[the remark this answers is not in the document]
gobengo: I did this to help with debugging failed tests. When this runs in miniflare, it doesn't log the errors on stdout.
But later I can make it so `env.log`, when in tests, does the `console.warn` so we dont have to do it here

packages/access-api/src/service/access-authorize.js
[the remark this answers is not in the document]
gobengo: I changed it back.
At the time I was tracking down a bug and was worried it was applying more than it should.

packages/access-api/src/service/access-confirm.js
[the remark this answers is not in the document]
gobengo: It was to appease the fact that I wanted to make `opts.agent` on createSessionProofs of type `DID<'key'>` since that's all I ever expect it to be. I relaxed that to `DID` and so changed this back to what it was too.

packages/access-client/src/utils/did-mailto.js
gobengo: If the did-mailto syntax banned empty segments, I would add that, but I think it is relatively undefined.
I would be in favor of `email` opt being an already-validated `Email` obj, and that may make sense as a check when parsing a string into an email.

packages/access-client/src/agent.js
```diff
@@ -141,7 +181,6 @@ export class Agent {
    */
   async addProof(delegation) {
     validate(delegation, {
-      checkAudience: this.issuer,
```
travis, line 144: I don't think removing this check should be necessary - added it back and tested the whole flow manually and it worked. I tried to add it back but there's a an access-api test that's failing, @gobengo is planning on looking into why there's a discrepancy between the test and the observed behavior

packages/access-client/src/agent-use-cases.js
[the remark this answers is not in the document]
gobengo: (sometimes `delegee` is passed as a `did:mailto` but usually we use `agent` to refer to a `did:key`. also sometimes we make functions that take two different agents, so it's nice to describe the role they play not what they are)

packages/access-client/src/agent-use-cases.js
```diff
@@ -0,0 +1,157 @@
...
+  const res = await access.invokeAndExecute(Access.claim, {
+    audience: access.connection.id,
+    with: delegee,
+  })
+  if (res.error) {
+    throw new Error('error claiming delegations')
+  }
+  const delegations = Object.values(res.delegations).flatMap((bytes) =>
+    bytesToDelegations(bytes)
+  )
```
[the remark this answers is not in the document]
gobengo: if that's the case something is wrong and I want this to error bigly, which I think it would.

What would you expect to happen?

packages/access-client/src/agent-use-cases.js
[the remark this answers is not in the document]
gobengo: > Is this replacing a web socket ?

no, it's just providing an alternative

packages/access-client/src/agent.js
```diff
@@ -21,13 +24,74 @@ import {
...
+      with: space,
+    },
+  ],
+  proofs = []
+) {
+  return ucanto.delegate({
+    issuer,
+    audience: account,
+    capabilities,
+    proofs,
```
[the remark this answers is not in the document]
gobengo: done

packages/access-client/src/agent.js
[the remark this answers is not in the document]
gobengo: removed it

packages/access-client/src/agent.js
travis: filed https://github.com/web3-storage/w3protocol/issues/566 for this

packages/access-client/src/agent.js
gobengo: moved to static `addProviderAndDelegateToAccount`

packages/access-client/src/agent.js
gobengo: together we realized @travis is right. It was only necessary because my test was explicitly trying to claim delegations with=account, which I removed.

gobengo: https://github.com/web3-storage/w3protocol/pull/433/commits/70b663d534edf96959d86fd98f372fcbfc2c4fdc

gobengo (approved):
good idea

travis (approved):

PR #437, opened 2023-02-14
feat: add startCursor and endCursor to ListResponse

travis:
per the change in https://github.com/web3-storage/w3infra/pull/139, add startCursor and endCursor and deprecate cursor


vasco-santos (approved):
LGTM. Let's wait on the `upload-api` before merging

PR #449, opened 2023-02-24
feat: move validation flow to a Durable Object to make it ⏩ fast ⏩ fast ⏩ fast ⏩ 

travis:
Introduce a [Cloudflare Durable Objects](https://developers.cloudflare.com/workers/runtime-apis/durable-objects/)-based space verification workflow.

Rather than stashing a UCAN in KV and polling inside the Websocket handler until eventual consistency makes it available to the registration process in the CLI or w3console, we use a Durable Object to create a direct connection between the Websocket and the HTTP POST that provides the delegation.

It feels most natural to create a Durable Object per space and forward the Websocket request on to the Durable Object and let it handle it entirely. This necessitates adding the space DID information to the HTTP path, which means adding a new endpoint and will require clients to be updated in the wild before we get rid of the KV flow entirely. We should remove KV-based verification once this has rolled out to all of our users.

TODO
- [x] write tests
- [ ] make sure error handling makes sense


alanshaw (changes requested):

packages/access-api/src/routes/validate-ws-did.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * @param {import('@web3-storage/worker-utils/router').ParsedRequest} req
+ * @param {import('../bindings.js').RouteContext} env
+ */
+export async function validateWSDID(req, env) {
+  const durableObjectID = env.spaceVerifiers.idFromName(req.params.did)
+  const durableObject = env.spaceVerifiers.get(durableObjectID)
+  const response = await durableObject.fetch(req)
+  // wrap the response because it's not possible to set headers on the response we get back from the durable object
```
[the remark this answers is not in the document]
travis: ah so in the outer worker handler function we use `corsHeaders` and `console.log.end` which both set headers on the response:

https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/src/index.js#L30

Without this wrapping this was failing with:

```
[2023-02-24T15:37:32.240Z]  TypeError: immutable
    at Headers.set (/Users/travis/dev/pl/w3protocol/node_modules/.pnpm/undici@5.9.1/node_modules/undici/lib/fetch/headers.js:393:13)
    at corsHeaders (/Users/travis/dev/pl/w3protocol/node_modules/.pnpm/@web3-storage+worker-utils@0.4.3-dev/node_modules/@web3-storage/worker-utils/src/cors.js:15:17)
    at Object.fetch (/Users/travis/dev/pl/w3protocol/packages/access-api/src/index.js:32:30)
    at ServiceWorkerGlobalScope.[kDispatchFetch] (/Users/travis/dev/pl/w3protocol/node_modules/.pnpm/@miniflare+core@2.11.0/node_modules/@miniflare/core/src/standards/event.ts:385:13)
    at /Users/travis/dev/pl/w3protocol/node_modules/.pnpm/@miniflare+http-server@2.11.0/node_modules/@miniflare/http-server/src/index.ts:298:20
    at Server.<anonymous> (/Users/travis/dev/pl/w3protocol/node_modules/.pnpm/@miniflare+http-server@2.11.0/node_modules/@miniflare/http-server/src/index.ts:418:24)
```

which from what I can tell (eg, https://github.com/cloudflare/miniflare/issues/365) seems like it does mirror the behavior of Cloudflare production workers?

packages/access-api/src/durable-objects/space-verifier.js
```diff
@@ -0,0 +1,89 @@
...
+        this.sendDelegation()
+      }
+      return new Response(undefined, {
+        status: 101,
+        webSocket: client,
+      })
+    } else if (req.method === 'PUT' && path === '/delegation') {
+      const ucan = await req.text()
+      await this.state.storage.put('ucan', ucan)
+      this.ucan = ucan
```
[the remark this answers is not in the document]
travis: ah good point! will change that

travis:
OK this is good to go! caught up with @gobengo about how this intersects with the work he and @Gozala are going and his recommendation was to get this shipped. I do think this whole flow will be deprecated soon so I haven't bothered with extensive testing or refactoring tests, but did add one test similar to the `access/authorize` test that makes sure the happy path (ie, user opens websocket, clicks email, delegation is sent through websocket) works.

My understanding is that with the intention for the `access/authorize` based work that's in progress is to make that transport agnostic (ie, purely based on ucanto RPC), which will make it slightly trickier to do fancy transport stuff like we're doing with this websocket. After chatting with @gobengo I'm going to do a little research and see what it might take to implement ucanto channel that uses a websocket - that would be the ideal followup to this work and would let us get this same great user experience with that work.

tldr, ready for re-review! this should all be backwards compatible with existing clients so I think we can ship it at will.

packages/access-api/src/durable-objects/space-verifier.js
```diff
@@ -0,0 +1,131 @@
...
+
+  async cleanupUCAN() {
+    this.ucan = undefined
+    await this.state.storage.put('ucan', '')
+  }
+
+  /**
+   * @param {Request} req
+   */
+  async fetch(req) {
```
gobengo, line 70: I would expect to see an authorization check somewhere to make we don't give the ucan to someone who shouldn't have it?

If someone else tries to connects first before the client we expect, what would happen? (It's been a long day, but I read it like that would deny service to the desired user)

I wonder if (long term) maybe this should be like `AgentVerifier` and be keyed off of the agent did, not space did.
(or `agent did` + `cid(authorizationRequest)` even).

It seems like keying this off of space and only allowing one server/ucan at a time will:
* prevent more than one of these flows happening at a time for a space - not really needed for `voucher/redeem` but it makes this less useful with new `authorize` flow

travis: > I would expect to see an authorization check somewhere to make we don't give the ucan to someone who shouldn't have it?

hm, I would probably handle this further up the stack, but also I'm not sure we do this with the existing implementation and if we do I think we're still doing that for this flow? 

tbh I'm not entirely sure whether we need to do additional auth here or not, but I'm not sure what that would look like concretely - do you have some sort of signature verification in mind?

travis: > I wonder if (long term) maybe this should be like AgentVerifier and be keyed off of the agent did, not space did.
> (or agent did + cid(authorizationRequest) even).

I think this is exactly the direction we should go with the new stuff coming in yep

travis: > It seems like keying this off of space and only allowing one server/ucan at a time will prevent more than one of these flows happening at a time for a space - not really a problem for voucher but it makes this less useful with new access/authorize + access/claim flow which is not 1:1 with a single space.

yep - keying off Agent will be the way to go for the path that handles the `access/authorize` flow

gobengo (approved):
I don't think this introduces any new issues, but it does definitely improve the perf of the voucher websocket

packages/access-api/src/durable-objects/space-verifier.js
```diff
@@ -0,0 +1,131 @@
...
+  spaceVerifiers,
+  space,
+  ucan
+) {
+  const durableObjectID = spaceVerifiers.idFromName(space)
+  const durableObject = spaceVerifiers.get(durableObjectID)
+  // hostname is totally ignored by the durable object but must be set so set it to example.com
+  const response = await durableObject.fetch('https://example.com/delegation', {
+    method: 'PUT',
+    body: ucan,
```
Gozala, line 20: Does body need to be a string or can we pass binary data ? If we could pass binary it might be a better option.

packages/access-api/src/durable-objects/space-verifier.js
Gozala: I think invocation CID might be better yet as agent in theory might start multiple authorization flows or maybe we'll want to deliver other messages over websocket.

packages/access-api/src/service/voucher-claim.js
```diff
@@ -35,16 +35,16 @@ export function voucherClaimProvider(ctx) {
...
-
     await ctx.email.sendValidation({
       to: capability.nb.identity.replace('mailto:', ''),
       url,
     })
+
+    // For testing
+    if (ctx.config.ENV === 'test') {
+      return encoded
+    }
```
Gozala, line 48: Nit: Is this really necessary now hat we can provide debug `email` thing for tests ? Because ucan is already in the query param. I have changed it in other placed, must have missed it here.

Gozala (approved):
Thanks for improving things. I think it's good as is, yet I would really appreciate some code comments because without been familiar with these code paths I struggled to follow along. Maybe some comments could be added in a followup change ?

packages/access-api/src/durable-objects/space-verifier.js
travis: ah good question, I suspect binary would work? noted and a good change for the next version of this I think

packages/access-api/src/durable-objects/space-verifier.js
travis: hm I don't totally follow, but seems reasonable as long as both sides of the equation (ie, the code calling registerSpace and the code handling the request from the email link) can both calculate the invocation CID

packages/access-api/src/service/voucher-claim.js
travis: ah good question and I think no - I had originally been planning to rework the tests to get rid of this, but paused on that because we'll probably end up porting all of this over to the new session stuff, and it will make sense to rework tests and generally clean all of this up at that point.

travis:
> Thanks for improving things. I think it's good as is, yet I would really appreciate some code comments because without been familiar with these code paths I struggled to follow along. Maybe some comments could be added in a followup change ?

thank youuuuu - yes, I think this all needs a bunch of cleanup, and more than that needs to be ported over to the new session based stuff - I think most of your comments make sense to implement as part of that next tranche of work, but that may be coming very soon as I'm quite annoyed by having to wait for my account authorization in the new code paths ;-p

alanshaw (approved):

Issue #457, opened 2023-02-27
Succesfull access/authorize should delegate capabilities to the agent so it can claim them


Gozala:
When user issues `access/authorize` with some agent it should send invocation like shown below which in English translates to: "Agent `did:key:z6Mkk89bC3JrVqKie71YEcc5M1SMVxuCgNx6zLZ8SYJsxALi` is requesting all of the capabilities (`can: *`) to be delegated to it by `did:mailto:web.mail:alice` 

```json
{
  "v": "0.9.1",
  "iss": "did:key:z6Mkk89bC3JrVqKie71YEcc5M1SMVxuCgNx6zLZ8SYJsxALi",
  "aud": "did:web:web3.storage",
  "att": [
    {
      "with": "did:key:z6Mkk89bC3JrVqKie71YEcc5M1SMVxuCgNx6zLZ8SYJsxALi",
      "can": "access/authorize",
      "nb": {
        "iss": "did:mailto:web.mail:alice",
        "att": [{"can":"*"}]
       }
    }
  ],
  "prf": [],
  "exp": 1685602800,
  "s": {
    "/": {
      "bytes": "7aEDQJbJqmyMxTxcK05XQKWfvxG+Tv+LWCJeE18RSMnciCZ/RQ21U75LA0uFSvIjdqnF5RaauZTE8mh2ZYMBBejdJQ4"
    }
  }
}
```

Capability provider is expected to perform interactive authorization with an email holder using email flow. If and when user grants requested capabilities `access/authorize` provider is expected to delegate two capabilities to `did:key:z6Mkk89bC3JrVqKie71YEcc5M1SMVxuCgNx6zLZ8SYJsxALi` (probably through `access/delegate` invocation)

1. Delegation from `did:mailto:web.mail:alice` to `did:key:z6Mkk89bC3JrVqKie71YEcc5M1SMVxuCgNx6zLZ8SYJsxALi` that is [signed a special signature type](https://github.com/web3-storage/specs/blob/feat/auth%2Baccount/w3-account.md#authorization-session-signature-format) (that is basically not signed).

```json
{
  "v": "0.9.1",
  "iss": "did:mailto:web.mail:alice",
  "aud": "did:key:z6Mkk89bC3JrVqKie71YEcc5M1SMVxuCgNx6zLZ8SYJsxALi",
  "att": [
    {
        "with": "ucan:*",
        "can": "*"
    }
  ],
  "prf": [],
  "exp": null,
  "s": {
[...]

Issue #455, opened 2023-02-25
I can access/claim as `did:key:{}`


gobengo:
Part of:
* https://github.com/web3-storage/w3protocol/issues/394

Unblocks:
* https://github.com/gobengo/w3protocol-test/pull/4
* #454 

This is going to be very quick compared to #454 

* [x] add `claim` to the access-api service.access obj https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/src/service/index.js#L28
* [x] add at least a cursory test that access-api handles invocations of access/claim issued by did:key

PR #460, opened 2023-02-28
feat: access/authorize confirmation email click results in a delegation back to the issuer did:key so that access/claim works

gobengo:
Motivation:
* #455 
* This implements the second delegation described in https://github.com/web3-storage/w3protocol/issues/457
* make this test pass once we deploy this to staging https://github.com/gobengo/w3protocol-test/pull/6

Test Case:
* https://github.com/gobengo/w3protocol-test/pull/6/files#diff-698e92d7467a4a470689d4cb120272c6cac28864d4960ac12a3720ab7c35a15cR364

### DIFF
--- a/packages/access-api/src/routes/validate-email.js
+++ b/packages/access-api/src/routes/validate-email.js
@@ -160,6 +160,34 @@ async function session(req, env) {
 
     const agentPubkey = accessSessionResult.capability.nb.key
 
+    const update = await ucanto.delegate({
+      issuer: env.signer,
+      audience: account,
+      capabilities: [
+        {
+          with: env.signer.did(),
+          can: './update',
+          nb: {
+            key: agentPubkey,
+          },
+        },
+      ],
+    })
+    const attestKeyAuthorizesAccount = await ucanto.delegate({
+      issuer: env.signer,
+      audience: { did: () => agentPubkey },
+      capabilities: [
+        {
+          with: env.signer.did(),
+          can: 'ucan/attest',
+          nb: {
+            proof: update.cid,
+          },
+        },
+      ],
+      proofs: [update],
+    })
+
     // create delegations that should be claimable
     const delegationAccountToKey = await ucanto.delegate({
       issuer: account,
@@ -182,10 +210,10 @@ async function session(req, env) {
       issuer: env.signer,
       audience: {
         did() {
-          return accessSessionResult.capability.nb.key
+          return agentPubkey
         },
       },
-      proofs: [delegation],
+      proofs: [delegationAccountToKey],
       capabilities: [
         {
           can: 'ucan/attest',
@@ -196,7 +224,10 @@ async function session(req, env) {
         },
       ],
     })
-    await env.models.delegations.putMany(delegateToKey)
+    await env.models.delegations.putMany(
+      delegateToKey,
+      attestKeyAuthorizesAccount
+    )
   }
 
   try {
--- a/packages/access-api/src/routes/validate-email.js
+++ b/packages/access-api/src/routes/validate-email.js
@@ -1,5 +1,8 @@
 /* eslint-disable no-unused-vars */
-import { stringToDelegation } from '@web3-storage/access/encoding'
+import {
+  delegationsToBytes,
+  stringToDelegation,
+} from '@web3-storage/access/encoding'
 import * as Access from '@web3-storage/capabilities/access'
 import QRCode from 'qrcode'
 import { toEmail } from '../utils/did-mailto.js'
@@ -12,6 +15,7 @@ import {
 import * as ucanto from '@ucanto/core'
 import * as validator from '@ucanto/validator'
 import { Verifier } from '@ucanto/principal/ed25519'
+import { ed25519 } from '@ucanto/principal'
 
 /**
  * @param {import('@web3-storage/worker-utils/router').ParsedRequest} req
@@ -149,6 +153,28 @@ async function session(req, env) {
         `unable to validate access session: ${accessSessionResult.error}`
       )
     }
+    const accountDID = accessSessionResult.audience.did()
+    // @todo: use new ucanto `Account` instead
+    const accountOneOffKey = await ed25519.generate()
+    const account = accountOneOffKey.withDID(accountDID)
+
+    const agentPubkey = accessSessionResult.capability.nb.key
+
+    // create delegations that should be claimable
+    const delegationAccountToKey = await ucanto.delegate({
+      issuer: account,
+      audience: {
+        did() {
+          return agentPubkey
+        },
+      },
+      capabilities: [
+        {
+          with: 'ucan:*',
+          can: '*',
+        },
+      ],
+    })
     // generate a delegation to the key that we can save in
     // models.delegations to be found by subsequent access/claim
     // invocations invoked by the did:key
@@ -162,9 +188,11 @@ async function session(req, env) {
       proofs: [delegation],
       capabilities: [
         {
-          can: './update',
+          can: 'ucan/attest',
           with: env.signer.did(),
-          nb: accessSessionResult.capability.nb,
+          nb: {
+            proof: delegationsToBytes([delegationAccountToKey]),
+          },
         },
       ],
     })
--- a/packages/access-api/src/routes/validate-email.js
+++ b/packages/access-api/src/routes/validate-email.js
@@ -149,7 +149,7 @@ async function session(req, env) {
         `unable to validate access session: ${accessSessionResult.error}`
       )
     }
-    const account = { did: () => accessSessionResult.audience.did() }
+    const account = accessSessionResult.audience
     const agentPubkey = accessSessionResult.capability.nb.key
     const wrappedKeyCanAsignForAccount = await ucanto.delegate({
       issuer: env.signer,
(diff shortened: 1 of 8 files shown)


packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +138,38 @@ async function session(req, env) {
...
+      throw new Error(
+        `unable to validate access session: ${accessSessionResult.error}`
+      )
+    }
+    // generate a delegation to the key that we can save in
+    // models.delegations to be found by subsequent access/claim
+    // invocations invoked by the did:key
+    const delegateToKey = await ucanto.delegate({
+      issuer: env.signer,
+      audience: {
```
Gozala, line 157: Audience of `./update` capabilities is meant to be non `did:key` principal, because those are only times ucanto would look at `./update` capabilities. Delegation to the agent key is not going to get you anything other than been able to claim that delegation from that agent afterwards.

Gozala (changes requested):
It's cool that more tests are added, but I'm afraid it's not going to do what you're aiming for.

gobengo:
@Gozala after your last review e.g. [this comment](https://github.com/web3-storage/w3protocol/pull/460#discussion_r1119598412), I [updated](https://github.com/web3-storage/w3protocol/pull/460/commits/1be596f151a72a1d8ad793882d59d7cf0e4e6e29) this to have clicking the confirmation email generate a `ucan/attest` as described in #457 .

wdyt?

packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +139,93 @@ async function session(req, env) {
...
+      )
+    }
+    const accountDID = accessSessionResult.audience.did()
+    // @todo: use new ucanto `Account` instead
+    const accountOneOffKey = await ed25519.generate()
+    const account = accountOneOffKey.withDID(accountDID)
+
+    const agentPubkey = accessSessionResult.capability.nb.key
+
+    const update = await ucanto.delegate({
```
Gozala, line 160: It would really help to add comment above explaining what this delegation intends to do. In the version that I think this implements this delegation is attestation from us that account has authorized agent key to be able to sign ucans where issuer is account did.

packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +139,93 @@ async function session(req, env) {
...
+        {
+          with: env.signer.did(),
+          can: './update',
+          nb: {
+            key: agentPubkey,
+          },
+        },
+      ],
+    })
+    const attestKeyAuthorizesAccount = await ucanto.delegate({
```
Gozala, line 173: Comment here describing what it's meant to do would also help, because it does not do what spec or issue suggests and it's not clear to me what the intentions are.

packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +139,93 @@ async function session(req, env) {
...
+      issuer: env.signer,
+      audience: { did: () => agentPubkey },
+      capabilities: [
+        {
+          with: env.signer.did(),
+          can: 'ucan/attest',
+          nb: {
+            proof: update.cid,
```
Gozala, line 181: This should be the CID for the (unsigned) delegation from an account to the agent. This delegation should be replacement for the `update` above. More concretely if `update` was attesting to the authorization of the key, this attestation should be for the authorization of the specific delegation.

That is the core change in the spec, if previously we were attesting to key authorization which implied sudo privileges, in a new version we attest to the specific delegation authorization. Which means it could be sudo privileges or a very specific capabilities.

packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +139,93 @@ async function session(req, env) {
...
+          },
+        },
+      ],
+      proofs: [update],
+    })
+
+    // create delegations that should be claimable
+    const delegationAccountToKey = await ucanto.delegate({
```
Gozala, line 189: This is somewhat correct, except it will have a signature from `accountOneOffKey` key as opposed to having an "absent" signature. Which may be ok temporarily, but not in MVP because those could fail to validate.

packages/access-api/src/routes/validate-email.js
Gozala: Reading further down I think you could just remove this delegation.

packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +139,93 @@ async function session(req, env) {
...
+        {
+          with: 'ucan:*',
+          can: '*',
+        },
+      ],
+    })
+    // generate a delegation to the key that we can save in
+    // models.delegations to be found by subsequent access/claim
+    // invocations invoked by the did:key
+    const delegateToKey = await ucanto.delegate({
```
Gozala, line 206: Probably should be renamed and update comments as it does a different thing now.

packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +139,93 @@ async function session(req, env) {
...
+      issuer: env.signer,
+      audience: {
+        did() {
+          return agentPubkey
+        },
+      },
+      proofs: [delegationAccountToKey],
```
Gozala, line 213: I'm not sure what is this for ? This could either contain delegation from `did:web:web3.storage` to a service key, or if they are the same no proof is required.

packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +139,93 @@ async function session(req, env) {
...
+          with: env.signer.did(),
+          nb: {
+            proof: delegationAccountToKey.cid,
+          },
+        },
+      ],
+    })
+    await env.models.delegations.putMany(
+      delegateToKey,
+      attestKeyAuthorizesAccount
```
Gozala, line 226: I think you need to put `delegationAccountToKey` and `delegateToKey` instead.

Gozala (commented):
I think it's trying to do both old auth and new auth together but mixes two in a weird way.

packages/access-api/src/routes/validate-email.js
gobengo: this is what's needed to make this work https://github.com/web3-storage/w3protocol/pull/460/files#diff-0cebcdde20e55de2911e90a385f368240bd9109fc54b074b89759628b695e88bR202

packages/access-api/test/access-authorize.test.js
```diff
@@ -95,6 +99,129 @@ describe('access/authorize', function () {
...
+    const claimedNb = attest.capabilities[0].nb
+    assert.ok(
+      claimedNb && typeof claimedNb === 'object' && 'proof' in claimedNb,
+      'should have nb.proof'
+    )
+    const expectAccountToKey = proofDelegation(attest.proofs[0])
+    assert.ok(expectAccountToKey, 'expect proofs to contain delegation')
+    assert.deepEqual(expectAccountToKey.issuer.did(), accountDID)
+    assert.deepEqual(expectAccountToKey.audience.did(), issuer.did())
+    assert.deepEqual(expectAccountToKey.capabilities, [
```
gobengo, line 189: remove this whole delegation if it's not needed for this test

gobengo: done in https://github.com/web3-storage/w3protocol/pull/460/commits/685af659cfe1c7f6ab09e79a36cce36df3aa0ecb

packages/access-api/src/routes/validate-email.js
gobengo: renamed/recommented in https://github.com/web3-storage/w3protocol/pull/460/commits/685af659cfe1c7f6ab09e79a36cce36df3aa0ecb

gobengo:
> I think it's trying to do both old auth and new auth together but mixes two in a weird way.

I removed thew new

packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +138,46 @@ async function session(req, env) {
...
+      capability: Access.session,
+      principal: Verifier,
+      authority: env.signer,
+    })
+    if (accessSessionResult.error) {
+      throw new Error(
+        `unable to validate access session: ${accessSessionResult.error}`
+      )
+    }
+    const account = { did: () => accessSessionResult.audience.did() }
```
Gozala, line 152: nit: You don't need to wrap you could just

```suggestion
    const account = accessSessionResult.audience
```

packages/access-api/src/routes/validate-email.js
```diff
@@ -134,6 +138,46 @@ async function session(req, env) {
...
+    const agentPubkey = accessSessionResult.capability.nb.key
+    const wrappedKeyCanAsignForAccount = await ucanto.delegate({
+      issuer: env.signer,
+      audience: { did: () => agentPubkey },
+      capabilities: [
+        {
+          with: env.signer.did(),
+          can: 'access-api/delegation',
```
Gozala, line 160:
```suggestion
          can: 'access/delegation'
```

Nit: It just occurred to me that you can use `access/delegate` here as spec-ed except with `audience` being  an `agent` as opposed to service.

Gozala (approved):

packages/access-api/src/routes/validate-email.js
gobengo: I'm down to change to that post-merge but otherwise I'd prefer to make this a meaningless `can`

gobengo:
when testing this on staging, I ran into an error due to https://github.com/web3-storage/w3protocol/issues/464

Issue #459, opened 2023-02-28
story: I can register a storage provider with a space using access/authorize + provider/add


gobengo:
Motivation:
* enable demonstrating how to register a space without using any `voucher/` invocations as described by @Gozala here https://github.com/gobengo/w3protocol-test/pull/5#issuecomment-1447342066
  * this is useful because access/delegate is only allowed on a registered space
* will probably be used by https://github.com/web3-storage/w3protocol/issues/459

Inputs:
* https://github.com/web3-storage/specs/blob/main/w3-provider.md#provideradd
> A more simplified flow exists for cases when provider is needed for a specific (space) consumer. In those cases a provider/add capability can be invoked, which is equivalent of [provider/get](https://github.com/web3-storage/specs/blob/main/w3-provider.md#providerget), except [nb.consumer](https://github.com/web3-storage/specs/blob/main/w3-provider.md#get-nbconsumer) MUST be a concrete DID. On successful invocation, the provider takes care of invoking [consumer/add](https://github.com/web3-storage/specs/blob/main/w3-provider.md#consumeradd) instead of delegating it back to an agent, which removes the need for an extra roundtrip.
* https://github.com/web3-storage/w3protocol/pull/410/files

gobengo:
#status implementing at https://github.com/web3-storage/w3protocol/pull/462

PR #462, opened 2023-02-28
feat: access-api handles provider/add invocations

gobengo:
Motivation:
* implement https://github.com/web3-storage/w3protocol/issues/459
* handle `provider/add` as a way of registering space

Steps
* [x] scaffold with in-memory data store
* [x] make it work with sqlite data store


packages/access-api/src/utils/context.js
```diff
@@ -60,6 +61,7 @@ export function getContext(request, env, ctx) {
       spaces: new Spaces(config.DB),
       validations: new Validations(config.VALIDATIONS),
       accounts: new Accounts(config.DB),
+      storageProvisions: createStorageProvisions([]),
```
gobengo, line 64: @todo build instance backed by `config.DB`

packages/access-api/migrations/0006_add_storage_provider_consumer_schema.sql
```diff
@@ -0,0 +1,21 @@
...
+/*
+goal: add a table to keep track of the storage provider for a space.
+to enable https://github.com/web3-storage/w3protocol/issues/459
+*/
+
+CREATE TABLE
+    -- provision: the action of providing or supplying something for use
+    -- use case: representing the registration of a storage provider to a space
+    IF NOT EXISTS provisions (
+        id INTEGER NOT NULL PRIMARY KEY,
```
Gozala, line 12: Mind adding comment also what is the `id` here ? I'll probably discover as I read more of the pr, but comment here would be helpful

packages/access-api/src/service/provider-add.js
```diff
@@ -0,0 +1,58 @@
...
+ * @param {Ucanto.Invocation<import('@web3-storage/capabilities/types').ProviderAdd>} invocation
+ * @returns {Promise<Ucanto.Result<ProviderAddSuccess, ProviderAddFailure>>}
+ */
+
+/**
+ * @param {object} options
+ * @param {import('../types/provisions').StorageProvisions} options.storageProvisions
+ * @returns {ProviderAddHandler}
+ */
+export function createProviderAddHandler(options) {
```
Gozala, line 23: Nit: This obviously follows the style of code that is in the repo already, but I do find amount of indirection adds so much unnecessary complexity e.g. I would much rather have something like:

```ts
/**
 * @param {object} input
 * @param {Ucanto.Invocation<import('@web3-storage/capabilities/types').ProviderAdd>} input.invocation
 * @param {{ storageProvisions: import('../types/provisions').StorageProvisions }} input.context
 */
export const add ({ invocation, context: { storageProvisions } }) => {
   // ...
}
```

And in the service defs something like

```js
Server.provide(Provider.add, ({ invocation }) => add({ invocation, context })
```

Than all this layers and closures all over.

Also for what it's worth I'm going to add `context` to the ucanto provider handles to remove the need for all those closures.

packages/access-api/migrations/0006_add_storage_provider_consumer_schema.sql
gobengo: It's meant to just be an auto incrementing id that other tables could join to.
I considered not even having it. wdyt about having it at all?

packages/access-api/src/types/provisions.ts
```diff
@@ -0,0 +1,30 @@
+import * as Ucanto from '@ucanto/interface'
+
+export type AlphaStorageProvider = 'did:web:web3.storage:providers:w3up-alpha'
+
+/**
+ * action which results in provisionment of a space consuming a storage provider
+ */
+export interface StorageProvisionCreation {
```
Gozala, line 8: Nit: I would prefer if we did not had `Storage` prefix, protocol is meant to allow you to add capability providers which can be for storage or other things sometimes for multiple things. It is true that `did:web:web3.storage:providers:w3up-alpha` is mostly for storage, yet it's not just storage it's also `upload/*` and `access/*`.

packages/access-api/test/helpers/ucanto-test-utils.js
```diff
@@ -1,11 +1,20 @@
...
 import * as assert from 'assert'
+import * as principal from '@ucanto/principal'
+
+/**
+ * @typedef HelperTestContext
+ * @property {Ucanto.Signer<Ucanto.DID<'key'>>} issuer
+ * @property {Ucanto.Signer<Ucanto.DID>} service
+ * @property {Ucanto.ConnectionView<Record<string, any>>} conn
+ * @property {import('miniflare').Miniflare} mf
+ */
```
Gozala, line 12: Nit: Can we just stick typedefs like these into TS files, they are so much easier to read.

packages/access-api/test/provider-add.test.js
```diff
@@ -0,0 +1,293 @@
...
+    })
+    it(`can be invoked by did:mailto w/ session ./update`, async () => {
+      const agentA = await principal.ed25519.generate()
+      const accountDID = /** @type {Ucanto.DID<'mailto'>} */ (
+        'did:mailto:example.com:foo'
+      )
+      const account = { did: () => accountDID }
+      const space = await principal.ed25519.generate()
+      const service = await providerAddHandlerVariant.audience
+      const agentACanSignAsAccountDID = await ucanto.delegate({
```
Gozala, line 64: I'm sure you know this, but this is no longer going to work. Instead you'd have to go through `access.Authorize` or alternatively create delegation from `absentee` account and then attestation from the `service` if you want to simplify the test.

Gozala: It seems that `testAuthorizeClaimProviderAdd` basically tests the same thing so I think you may want to remove this test entirely.

packages/capabilities/src/provider.js
```diff
@@ -0,0 +1,60 @@
...
+      return (
+        fail(equalWith(child, parent)) ||
+        fail(equal(child.nb.provider, parent.nb.provider, 'provider')) ||
+        fail(equal(child.nb.consumer, parent.nb.consumer, 'consumer')) ||
+        true
+      )
+    },
+  }),
+  derives: equalWith,
+})
```
Gozala, line 60: We no longer need the `top` and `base` stuff because new ucanto understands `*` patterns natively now.

```suggestion
export const add = capability({
  can: 'provider/add',
  with: AccountDID,
  nb: struct({
    provider: StorageProvider,
    consumer: DID.match({ method: 'key' }),
  }),
  derives: (child, parent) => {
    return (
      fail(equalWith(child, parent)) ||
      fail(equal(child.nb.provider, parent.nb.provider, 'provider')) ||
      fail(equal(child.nb.consumer, parent.nb.consumer, 'consumer')) ||
      true
    )
  },
})
```

packages/capabilities/src/provider.js
```diff
@@ -0,0 +1,60 @@
...
+ * by did:key in the `with` field.
+ */
+export const provider = top.derive({
+  to: capability({
+    can: 'provider/*',
+    with: DID,
+    derives: equalWith,
+  }),
+  derives: equalWith,
+})
```
Gozala, line 29: No longer need to do the `derive` stuff as ucanto now understands the `*` patterns natively. Also `equalWith` is obsolete because it's a default behavior.

```suggestion
export const provider = capability({
  can: 'provider/*',
  with: DID,
})
```

packages/capabilities/test/capabilities/provider.test.js
```diff
@@ -0,0 +1,68 @@
+import assert from 'assert'
+import { access } from '@ucanto/validator'
+import { Verifier } from '@ucanto/principal/ed25519'
+import * as Provider from '../../src/provider.js'
+import { bob, service, mallory } from '../helpers/fixtures.js'
+import * as ucanto from '@ucanto/core'
+
+describe('provider/add', function () {
+  it.skip('can invoke as account with ./update', async function () {
```
Gozala, line 9: This does not work because it requires `access/authorize` thing as you did in other tests.

packages/access-api/src/models/provisions.js
```diff
@@ -0,0 +1,99 @@
...
+        provider: item.provider,
+        issuer: item.account,
+      }
+    })
+    await this.#db
+      .insertInto(this.tableNames.provisions)
+      .values(rows)
+      .executeTakeFirstOrThrow()
+  }
+
```
Gozala, line 88: I think it would be nice to have method that gives you providers you have on space in shape of `{ [provider]: [subscriber, ...subscriber[]] }`. Which is more or less what I was asking to expose from the `space/info`.

Gozala (changes requested):

packages/access-api/src/types/provisions.ts
gobengo: At this layer of the types I wanted to model this particular use case really explicitly (e.g. with `space` and `account` vs `consumer` and `issuer`). In practice those get translated into a sql table that I think can accomodate the more general thing you're talking about over time. We could have many of these narrower types that mode [domain event](https://www.martinfowler.com/eaaDev/DomainEvent.html) like this that write into the same underlying thing.

> protocol is meant to allow you to add capability providers which can be for storage or other things sometimes for multiple things.
Because the underlying table accomodates this, I don't want to change this particular action type now. I'm not against revising it or adding other events in subsequent PR. or e.g. renaming the model to not have storage prefix but still maybe having this particular event be specific to adding a storage provider.

packages/access-api/migrations/0006_add_storage_provider_consumer_schema.sql
```diff
@@ -0,0 +1,21 @@
...
+        -- DID of the actor that is consuming the provider. e.g. a space DID
+        consumer TEXT NOT NULL,
+        -- DID of the provider e.g. a storage provider
+        provider TEXT NOT NULL,
+        -- DID of the actor that issued this provision
+        issuer TEXT NOT NULL,
```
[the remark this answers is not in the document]
gobengo: @Gozala The issuer isn't necessarily going to be billed, e.g. if there are several provisions for the same space. 

I don't want to attach any semantics to this other than 'the id of the thing that created this'. In ucan parlance that's `issuer` so that's why I used it. I think because when I used `author` you didn't like it, and I considered using [attributedTo](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-attributedto) but wanted to avoid the camelCase.

> It also may not be an issuer it's the with field that matters.

I can see how that might lead to confusion since the actor to attribute the capability to is not necessarily the final `issuer`.

> I think I would call it either account or subscriber

The thing is I tried to make this table generic enough to support this field not necessarily being an account did nor a subscriber...

> but something like sponsor or patron could also probably work.

I also considered `sponsor` so maybe we can agree there... I will update to use that.

packages/access-api/src/service/index.js
```diff
@@ -181,6 +198,18 @@ export function service(ctx) {
       fail() {
         throw new Error('test fail')
       },
+      /**
+       * @param {Ucanto.Invocation<Ucanto.Capability<'testing/space-storage', Ucanto.DID<'key'>, Ucanto.Failure>>} invocation
+       */
+      'space-storage': async (invocation) => {
```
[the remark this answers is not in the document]
gobengo: yeah. I thought about it but didn't want to extend the public api without discussion.

In this case I am very supportive because that will also be helpful to upload-api and things like https://github.com/web3-storage/w3infra/issues/134#user-content-implementation

packages/access-api/src/types/provisions.ts
gobengo: I removed 'Storage' prefix from the model object itself https://github.com/web3-storage/w3protocol/pull/462/commits/e62faaaa3c69c75e8f81a7ec35e9abccb4192da6

packages/access-api/test/helpers/ucanto-test-utils.js
gobengo: https://github.com/web3-storage/w3protocol/pull/462/commits/2e90305660f9b238c37a2193c09d92dcfe38b450

created a test/helpers/types.ts file we can use for test-related types like this

packages/access-api/migrations/0006_add_storage_provider_consumer_schema.sql
gobengo: added a comment. https://github.com/web3-storage/w3protocol/pull/462/commits/184c55a8dace6f56b92fbc7a99d10ca81975d666

fwiw it will not be autoincrementing it will be a 64-bit ROWID https://www.sqlite.org/lang_createtable.html#rowid

packages/access-api/migrations/0006_add_storage_provider_consumer_schema.sql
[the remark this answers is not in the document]
gobengo: changed to sponsor https://github.com/web3-storage/w3protocol/pull/462/commits/754463da0a58ad02b42eac45091e0fda8780bd8c

packages/access-api/src/service/index.js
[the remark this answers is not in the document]
gobengo: can I do it as a fast follow once I have verified that the space registration via `provider/add` works on staging at all?
https://github.com/web3-storage/w3protocol/issues/480

packages/access-api/src/service/provider-add.js
gobengo: this is a good idea and I'll try it out across all the relevant things I wrote in that style to use the less-closures style you suggest. https://github.com/web3-storage/w3protocol/issues/481

packages/access-api/test/provider-add.test.js
gobengo: > I'm sure you know this, but this is no longer going to work

Why won't it work?
It seems to work in the tests incl after #477 and with latest ucanto.

> It seems that testAuthorizeClaimProviderAdd basically tests the same thing so I think you may want to remove this test entirely.

I'm down to do that. It is slightly different because that other test doesn't explicitly test that `./update` works (even if that's currently what comes out of `access/claim`).

packages/capabilities/src/provider.js
gobengo: prettier!

gobengo: https://github.com/web3-storage/w3protocol/pull/462/commits/a199c11a4915b0858473f31e94072e881b8b737c

packages/capabilities/src/provider.js
gobengo: iirc even `provider/*` automatically derives `provider/add`? Correct me if I'm wrong.
I removed this whole variable entirely in https://github.com/web3-storage/w3protocol/pull/462/commits/a199c11a4915b0858473f31e94072e881b8b737c

packages/capabilities/test/capabilities/provider.test.js
gobengo: fixed in #477 (into this branch) I think

packages/access-api/src/models/provisions.js
gobengo: https://github.com/web3-storage/w3protocol/issues/480#issuecomment-1454235566

By `subscriber` I assume you mean the account DIDs?

gobengo:
@Gozala I made a few changes based on your suggestions, and linked to the commits under your feedback comments.
For some I created issues to handle as followup.
Let me know if I did that for any you wanted before merging this.

packages/access-api/migrations/0006_add_storage_provider_consumer_schema.sql
```diff
@@ -0,0 +1,22 @@
...
+
+/*
+goal: add a table to keep track of the storage provider for a space.
+to enable https://github.com/web3-storage/w3protocol/issues/459
+*/
+
+CREATE TABLE
+    -- provision: the action of providing or supplying something for use
+    -- use case: representing the registration of a storage provider to a space
+    IF NOT EXISTS provisions (
```
Gozala, line 11: Nit: In the past when I had to do more DB work using plural names for tables was considered a bad practice, [stack overflow still seems to think the singular is a the way to go](https://stackoverflow.com/questions/338156/table-naming-dilemma-singular-vs-plural-names) with bunch of reasons there. 

I'm not going to get upset if we use plural, but might be worth considering singular names instead.

packages/access-api/src/types/provisions.ts
Gozala, line 8: Nit: I find naming to be very confusing, can we simply call this

```suggestion
export interface ProvisionModel {
```

Or better yet

```suggestion
export interface Provision {
```

packages/access-api/src/types/provisions.ts
```diff
@@ -0,0 +1,30 @@
...
+  space: Ucanto.DID<'key'>
+  account: Ucanto.DID<'mailto'>
+  provider: AlphaStorageProvider
+}
+
+/**
+ * stores instances of a storage provider being consumed by a consumer
+ */
+export interface Provisions {
```
Gozala, line 17: Nit: I would also find it more intuitive if it was called following instead

```suggestion
export interface ProvisionStore {
```

packages/access-api/test/provider-add.test.js
Gozala: I'm not sure why this is passing, but ucanto no longer knows anything about `./update`. In fact now that you've pointed out it is passing tests, I'm concerned that it does as I would expect it to fail instead.

Gozala (changes requested):
Can you just add `cid` field to the table and otherwise I think it's all good.

packages/access-api/migrations/0006_add_storage_provider_consumer_schema.sql
gobengo: in general I don't mind singular at all, and maybe even slightly prefer it, but here I made it plural because `delegations` `accounts` `spaces` already was.

I don't have any objection to batch changing them to singular in one big other PR

packages/access-api/migrations/0006_add_storage_provider_consumer_schema.sql
gobengo: https://github.com/web3-storage/w3protocol/pull/462/commits/40a4f1612d23c1d61d8bc68821f801d3c15587a1

packages/access-api/test/provider-add.test.js
gobengo: We looked at this together and realized it passed because this test only tests the 'handler' which is not wrapped in a ucanto `Server.provide` that would have checked the sigs.

I removed the test cuz its misleading. https://github.com/web3-storage/w3protocol/pull/462/commits/69038913304ec4e7d488bdde5846f24cedebfd88

packages/access-api/src/models/provisions.js
```diff
@@ -0,0 +1,99 @@
...
+    void (/** @type {Provisions} */ (this))
+  }
+
+  /** @type {Provisions['count']} */
+  async count(...items) {
+    const { size } = await this.#db
+      .selectFrom(this.tableNames.provisions)
+      .select((e) => e.fn.count('provider').as('size'))
+      .executeTakeFirstOrThrow()
+    return BigInt(size)
```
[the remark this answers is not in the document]
gobengo: I made it a BigInt to match the interface. I used bigint in the interface for consistency with the `count()` method on delegations, which I made return a BigInt per @alanshaw review. https://github.com/web3-storage/w3protocol/pull/427#discussion_r1115836294

packages/access-api/src/types/provisions.ts
gobengo: I called Delegations `DelegationsStorage`. When I did alan commented that it didn't look like the other types that were called e.g. `Delegations`. I'm not against changing these names in subsequent PRs. If we call it ProvisionStore we should also change DelegationsStorage to DelegationsStore IMO

gobengo: renamed to ProvisionsStorage and idc if we rename to something else after. https://github.com/web3-storage/w3protocol/pull/462/commits/9b9a14d1d610c4c76860c1bdab14e05956db38f0

(My original intention was to eventually have like `type Provision = StorageProvision | OtherKindOfProvision`)

packages/access-api/src/types/provisions.ts
gobengo: https://github.com/web3-storage/w3protocol/pull/462/commits/c1e37f9408c2f75527b01e57fdf526cd2fe2d2fc

Gozala (approved):

packages/access-api/src/types/provisions.ts
alanshaw: I agree with @gozala but I want consistency in the code base. Can you please open another PR to bring the others in line?

PR #467, opened 2023-03-02
feat: includes proofs chains in the delegated authorization chain

Gozala:
This pr cleans up bunch of mess in the access/authorize code path and starts including proof chains in the delegated authorization so agents can actually utilize delegations.

### DIFF
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -31,7 +31,7 @@ export function service(ctx) {
       claim: (...args) => {
         // disable until hardened in test/staging
         if (ctx.config.ENV === 'production') {
-          throw new Error(`acccess/claim invocation handling is not enabled`)
+          throw new Error(`access/claim invocation handling is not enabled`)
         }
         return accessClaimProvider({
           delegations: ctx.models.delegations,
@@ -41,7 +41,7 @@ export function service(ctx) {
       delegate: (...args) => {
         // disable until hardened in test/staging
         if (ctx.config.ENV === 'production') {
-          throw new Error(`acccess/delegate invocation handling is not enabled`)
+          throw new Error(`access/delegate invocation handling is not enabled`)
         }
         return accessDelegateProvider({
           delegations: ctx.models.delegations,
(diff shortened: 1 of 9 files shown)


packages/access-api/src/service/index.js
```diff
@@ -31,7 +31,7 @@ export function service(ctx) {
       claim: (...args) => {
         // disable until hardened in test/staging
         if (ctx.config.ENV === 'production') {
-          throw new Error(`acccess/claim invocation handling is not enabled`)
+          throw new Error(`access/claim invocation handling is not enabled`)
```
gobengo, line 34: thanks for fixing my typo :)

packages/access-api/test/helpers/ucanto-test-utils.js
```diff
@@ -38,13 +38,14 @@ export function createTesterFromContext(createContext, options) {
  * using a service-issued voucher/redeem invocation
  *
  * @param {Iterable<Resolvable<Ucanto.Principal>>} spaces
- * @param {Ucanto.Signer<Ucanto.DID>} issuer
- * @param {Ucanto.ConnectionView<Record<string, any>>} conn
+ * @param {object} options
+ * @param {Ucanto.Signer<Ucanto.DID>} options.service
```
gobengo, line 42: In the future I expect to have `spaceRegistrationInvocation` create a `provider/add` issued by an agent (not the 'service'), which is why I called it the more generic `issuer`. I can always change it back if/when we use non-service-issuers for this.

packages/capabilities/src/access.js
```diff
@@ -74,33 +74,62 @@ export const authorize = capability({
...
+      }
     }],
     exp: null
     sig: "..."
   }
  * ```
  */
 export const session = capability({
-  can: './update',
+  can: 'ucan/attest',
```
gobengo, line 127: would be ideal to link to relevant spec for this, e.g. so someone could figure out how to build a valid/useful `Link` for `.nb.proof`

gobengo (approved):

packages/capabilities/src/access.js
Gozala: The link on line 105 add this, or do you mean something else ?

PR #469, opened 2023-03-03
fix: adjust migration 0005 to keep delegations table but create new used delegations_v2

gobengo:
Motivation:
* similar to https://github.com/web3-storage/w3protocol/pull/468

I think this is the one we should actually do, even though it is a little ugly to call the table `delegations_v2`.
If d1 can't do DROP TABLE, then I think the safest thing to do is to avoid it and always only ever create new ones. We can migrate to new tables and update the table name access-api uses in the same commit/pr.


gobengo:
not sure quite yet if we'll need to do this

Gozala (approved):

Issue #471, opened 2023-03-03
error result invoking access/claim against staging: 'TypeError: Input should be a non-empty Uint8Array.\n' 


gobengo:
Discovered While Testing
* https://github.com/gobengo/w3protocol-test/pull/6

```
    ok 7 - can invoke access/authorize against staging # SKIP
      ---
      duration_ms: 0.077875
      ...
    claimResult {
      name: 'HandlerExecutionError',
      cause: {
        name: 'TypeError',
        stack: 'TypeError: Input should be a non-empty Uint8Array.\n' +
          '    at Kl (d1-beta-facade.entry.js:15342:11)\n' +
          '    at AW (d1-beta-facade.entry.js:20990:11)\n' +
          '    at _a56.find (d1-beta-facade.entry.js:20974:13)\n' +
          '    at async HU (d1-beta-facade.entry.js:15567:18)\n' +
          '    at async d1-beta-facade.entry.js:15584:41\n' +
          '    at async EB (d1-beta-facade.entry.js:14474:12)\n' +
          '    at async vB (d1-beta-facade.entry.js:14462:12)\n' +
          '    at async bB (d1-beta-facade.entry.js:14456:42)\n' +
          '    at async YC (d1-beta-facade.entry.js:15634:11)\n' +
          '    at async bd.fetch (d1-beta-facade.entry.js:9706:17)',
        message: 'Input should be a non-empty Uint8Array.'
      },
      error: true,
      stack: 'HandlerExecutionError: service handler {can: "access/claim"} error: Input should be a non-empty Uint8Array.\n' +
        '    at EB (d1-beta-facade.entry.js:14476:13)\n' +
        '    at async vB (d1-beta-facade.entry.js:14462:12)\n' +
        '    at async bB (d1-beta-facade.entry.js:14456:42)\n' +
        '    at async YC (d1-beta-facade.entry.js:15634:11)\n' +
        '    at async bd.fetch (d1-beta-facade.entry.js:9706:17)\n' +
        '    at async Object.fetch (d1-beta-facade.entry.js:21017:13)',
      message: 'service handler {can: "access/claim"} error: Input should be a non-empty Uint8Array.',
      capability: {
        can: 'access/claim',
        with: 'did:key:z6MkkUymFxGqXWRqFNiaU4jCa2efexfDCNADqD7sQK3RbU2j'
      }
    }
```

gobengo:
[...]

PR #476, opened 2023-03-03
fix: DbDelegationsStorage#find throws UnexpectedDelegation w/ { row } if failed bytesToDelegations

gobengo:
Motivation:
* aid in debugging #471 
* hope this
  * confirms that this error is coming from `DbDelegationsStorage#find` (I see '.find' in the hard-to-read sentry stack trace)
  * gives insight into what the db row is like (esp `row.bytes`) that can't be parsed

### DIFF
--- a/packages/access-api/src/models/delegations.js
+++ b/packages/access-api/src/models/delegations.js
@@ -18,6 +18,8 @@ import {
  * @typedef {import("../types/database").Database<Tables>} DelegationsDatabase
  */
 
+export const delegationsTable = /** @type {const} */ ('delegations_v2')
+
 /**
  * DelegationsStorage that persists using SQL.
  * * should work with cloudflare D1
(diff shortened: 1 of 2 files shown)


Gozala:
This seems like a 🩹 that will not address actual problem, more specifically even if there is one malformed delegation stored in the DB for the audience it will make `access/claim` for that `audience` unusable without someone going and deleting that record from the table.

I propose we do something different here instead:

- We should still return delegations that aren't corrupt to the user
- We should mark delegations that are corrupt to skip those in the future
- We should make ourselves aware of corrupt delegations

I think the way to go about it is either:

1. Have another column in the table so we can mark invalid delegations
2. Write bad records into another table and delete them from the primary one.
3. Do nothing in the DB but report malformed delegations in sentry without crashing

Gozala (approved):
I don't mind landing this, but as per my comment I don't think it solves the problem

gobengo:
> 3. Do nothing in the DB but report malformed delegations in sentry without crashing

This is what I intend to do once I understand what's actually causing this. For now there is no good reason I'd expect to encounter bad bytes here, so I want to fail fast to understand. Once I understand why this might be happening, I will PR to do something like the above.

gobengo:
helped lead to resolution https://github.com/web3-storage/w3protocol/issues/471#issuecomment-1454220590

PR #478, opened 2023-03-03
fix: delegations model tries to handle if row.bytes is Array not Buffer (e.g. cloudflare)

gobengo:
## Rationale

This is the error I get from staging atm (on `access/claim`)

```
    claimResult {
      name: 'HandlerExecutionError',
      cause: {
        row: {
          cid: 'bafyreihqswtsrxvsp2pqf5a2jjiyb6zks3hynbfkhga7odwbat4ptxrf5e',
          bytes: [Array],
          issuer: 'did:mailto:dag.house:bengo',
          audience: 'did:key:z6MkkUymFxGqXWRqFNiaU4jCa2efexfDCNADqD7sQK3RbU2j',
          expiration: null,
          updated_at: '2023-03-03T16:41:54.999Z',
          inserted_at: '2023-03-03T16:41:54.999Z'
        },
        name: 'UnexpectedDelegation',
        stack: 'UnexpectedDelegation: failed to create delegation from row\n' +
```

Notably `bytes` is `Array` here.
Locally and in node.js, I think miniflare gives back a `Buffer` and (in nodejs) [this catches it](https://github.com/web3-storage/w3protocol/blob/a6dafcb0c44b20fa87bbe81d7a982fb13387084e/packages/access-api/src/utils/d1.js#L77). But on cloudflare. This PR would cast the bytes column on delegations model to Uint8Array if the db/kysely hand it back as an Array.

### DIFF
--- a/packages/access-api/src/models/delegations.js
+++ b/packages/access-api/src/models/delegations.js
@@ -169,7 +169,7 @@ export function createDelegationRowUpdate(d) {
  * @returns {ArrayBuffer|undefined} - undefined if unable to convert
  */
 export function delegationsTableBytesToArrayBuffer(sqlValue) {
-  if (isBuffer(sqlValue)) {
+  if (ArrayBuffer.isView(sqlValue)) {
     return new Uint8Array(
       sqlValue.buffer,
       sqlValue.byteOffset,
--- a/packages/access-api/src/models/delegations.js
+++ b/packages/access-api/src/models/delegations.js
@@ -3,6 +3,7 @@ import {
   delegationsToBytes,
   bytesToDelegations,
 } from '@web3-storage/access/encoding'
+import { isBuffer } from '../utils/common.js'
 
 /**
  * @typedef {import('@web3-storage/access/src/types').DelegationTable} DelegationRow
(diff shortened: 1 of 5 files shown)


packages/access-api/src/models/delegations.js
```diff
@@ -162,3 +163,20 @@ export function createDelegationRowUpdate(d) {
     bytes: delegationsToBytes([d]),
   }
 }
+
+/**
+ * @param {Array<number> | Buffer | unknown} sqlValue - value from kysely 'bytes' table - in node it could be a Buffer. In cloudflare it might be an Array
+ * @returns {ArrayBuffer|undefined} - undefined if unable to convert
+ */
+export function delegationsTableBytesToArrayBuffer(sqlValue) {
+  if (isBuffer(sqlValue)) {
```
Gozala, line 172:
```suggestion
  if (ArrayBuffer.isView(sqlValue)) {
```

packages/access-api/test/helpers/context.js
```diff
@@ -58,7 +58,7 @@ export async function context({ env = {}, globals } = {}) {
     bindings,
     d1Persist: undefined,
     buildCommand: undefined,
-    log: new Log(LogLevel.ERROR),
+    log: new Log(LogLevel.DEBUG),
```
gobengo, line 61:
```suggestion
    log: new Log(LogLevel.ERROR),
```

Gozala (approved):

packages/access-api/src/models/delegations.js
```diff
@@ -162,3 +162,20 @@ export function createDelegationRowUpdate(d) {
     bytes: delegationsToBytes([d]),
   }
 }
+
+/**
+ * @param {Array<number> | Buffer | unknown} sqlValue - value from kysely 'bytes' table - in node it could be a Buffer. In cloudflare it might be an Array
```
alanshaw, line 167: > value from kysely 'bytes' table

Not sure I understand this comment! Isn't kysely the client? What is a "bytes" table?

Issue #393, opened 2023-01-25
Implement `access/delegate` in `access-api`


alanshaw:
https://github.com/web3-storage/specs/blob/main/w3-access.md#accessdelegate

* [x] https://github.com/web3-storage/specs/issues/45
* [x] #414
* [x] space registration (because access/delegate checks for this) https://github.com/web3-storage/w3protocol/issues/459
* [x] https://github.com/web3-storage/w3protocol/pull/483
* [x] #425 access-api access/delegate handler stores nb.delegations so that they can be claimed by access/claim
  * unblocks https://github.com/web3-storage/w3protocol/issues/394
  * [x] https://github.com/web3-storage/w3protocol/issues/464 - discovered while tesitng in staging
* [x] demo `access/delegate` in observable against staging https://observablehq.com/d/95bfec64fbb3c2d1

gobengo:
#status need review on https://github.com/web3-storage/w3protocol/pull/420

gobengo:
@alanshaw fyi I just updated the OP link to the spec from a `feat/` branch to the main branch link

old: https://github.com/web3-storage/specs/blob/feat/w3-authorization/w3-access.md#accessdelegate
new: https://github.com/web3-storage/specs/blob/main/w3-access.md#accessdelegate

gobengo:
#status see https://github.com/web3-storage/w3protocol/issues/425#issuecomment-1426055978

gobengo:
#status I can't show off `access/delegate` working e2e until we have a way of registering a space, which we're implementing as https://github.com/web3-storage/w3protocol/issues/459

gobengo:
this is done in access-api trunk, and `@web3-storage/access` (access-client) makes use of it as of https://github.com/web3-storage/w3protocol/issues/395

PR #483, opened 2023-03-07
fix: access/delegate checks hasStorageProvider(space) in a way that provider/add allows access/delegate

gobengo:
Motivation:
* https://github.com/web3-storage/w3protocol/issues/393
* [discovered](https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1678146419356999?thread_ts=1678146205.801039&cid=C02BZPRS9HP) while testing https://observablehq.com/d/95bfec64fbb3c2d1@421

This makes it so that `access/delegate` will detect that a storage provider has been registered with a space via `provider/add`. Previously it would only allow spaces who had completed a `voucher/redeem` flow. Now it will support both. In the long run we will disable the `voucher/redeem` part.


packages/access-api/test/provider-add.test.js
```diff
@@ -97,6 +99,100 @@ for (const accessApiVariant of /** @type {const} */ ([
...
+            can: 'provider/add',
+            with: accountDid,
+          },
+          {
+            can: 'access/delegate',
+            with: space.did(),
+          },
+        ],
+      })
+      const sessionProofs = [
```
Gozala, line 127: Not sure I understand what you're trying to do here, but `ucan/attest` should only be considered when validating ucan issued by non did:key and an issuer different from the service.

UCAN been attested does not meet this criteria so it should have no effect on the validator.

packages/access-api/test/provider-add.test.js
```diff
@@ -97,6 +99,100 @@ for (const accessApiVariant of /** @type {const} */ ([
...
+          capability: {
+            can: 'provider/add',
+            with: accountDid,
+            nb: {
+              provider: 'did:web:web3.storage:providers:w3up-alpha',
+              consumer: space.did(),
+            },
+          },
+          proofs: [
+            // space says issuer can provider/add with this account
```
Gozala, line 157: Hmm... But space should have no authority over the account here.

packages/access-api/test/provider-add.test.js
```diff
@@ -97,6 +99,100 @@ for (const accessApiVariant of /** @type {const} */ ([
...
+              ],
+            }),
+            ...sessionProofs,
+          ],
+        })
+        .delegate()
+      const addStorageProviderResult = await accessApiVariant.invoke(
+        addStorageProvider
+      )
+      assertNotError(addStorageProviderResult)
```
Gozala, line 175: I suspect this is another instance where validator is not used, if that is not the case I would be very concerned. And if validator is disabled indeed, perhaps it's better to enable it so we don't end up in situation where tests pass but deployed code does not.

gobengo: I will check that, but I don't think so.
`accessApiVariant` looks to be built by the `createTesterFromContext` that tests against miniflare. And I traced that to a path that does `Server.provider(Provider.add, ...`. I will check outside of github though

packages/access-api/test/provider-add.test.js
gobengo: I'm trying to test whatever we expect to use for this https://github.com/web3-storage/w3cli/issues/46

packages/access-api/test/provider-add.test.js
```diff
@@ -97,6 +99,100 @@ for (const accessApiVariant of /** @type {const} */ ([
...
+    )
+  ) {
+    it('provider/add allows for access/delegate', async () => {
+      const space = await principal.ed25519.generate()
+      const issuer = await accessApiVariant.issuer
+      const service = await accessApiVariant.audience
+      const accountDid = /** @type {const} */ ('did:mailto:example.com:foo')
+      const serviceSessionAttest = await ucanto.delegate({
+        issuer: service,
+        audience: issuer,
```
Gozala, line 115: Swapped names here are pretty confusing btw

packages/access-api/test/provider-add.test.js
gobengo: We met in person, and I think we verfied that the validator is being used.

packages/access-api/test/provider-add.test.js
```diff
@@ -97,6 +99,100 @@ for (const accessApiVariant of /** @type {const} */ ([
...
+
+  if (
+    ['provider/add', 'access/delegate'].every((f) =>
+      accessApiVariant.features.has(f)
+    )
+  ) {
+    it('provider/add allows for access/delegate', async () => {
+      const space = await principal.ed25519.generate()
+      const issuer = await accessApiVariant.issuer
+      const service = await accessApiVariant.audience
```
gobengo, line 111: when @gozala and I were debugging earlier, he asked a good question e.g. to make sure that e.g. "service isn't same as issuer" or something.

I don't think they are, but one thing to keep in mind is that `service` may have been created using `withDID`:

note: this can be created in
https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/test/helpers/context.js#LL50C3-L52C4
like
```
  const servicePrincipal = Signer.parse(bindings.PRIVATE_KEY).withDID(
    bindings.DID
  )
```

packages/access-api/test/provider-add.test.js
gobengo: I agree. https://github.com/web3-storage/w3protocol/pull/492/commits/6d31d286deab582df5dde336c55f3e7368280db3

(I want to rename the `tester.issuer` to `tester.agent` or something too, but will do in standalone PR)

packages/access-api/test/provider-add.test.js
gobengo: https://github.com/web3-storage/w3protocol/pull/492/files#diff-b90906f041e93cd2ae9feab2b5b6e0e6b2cf0ac36d5917b5e2e67cdbf04b406cR108

packages/access-api/test/provider-add.test.js
gobengo: 🤦 good call
https://github.com/web3-storage/w3protocol/pull/483/commits/90afea4e7eb68353d0938e5adda76fe0cd8a12c0

packages/access-api/test/provider-add.test.js
```diff
@@ -97,6 +99,95 @@ for (const accessApiVariant of /** @type {const} */ ([
...
+      const accountDid = /** @type {const} */ ('did:mailto:example.com:foo')
+
+      const accountAuthorizesAgentClaim = await ucanto.delegate({
+        issuer: principal.Absentee.from({ id: accountDid }),
+        audience: agent,
+        capabilities: [
+          {
+            with: 'ucan:*',
+            can: '*',
```
Gozala, line 120: Nit: you could keep original capability set here, not implying you should.

packages/access-api/test/provider-add.test.js
```diff
@@ -97,6 +99,95 @@ for (const accessApiVariant of /** @type {const} */ ([
...
+      // storage provider added. So we should be able to delegate now
+      const accessDelegate = await ucanto
+        .invoke({
+          issuer: agent,
+          audience: service,
+          capability: {
+            can: 'access/delegate',
+            with: space.did(),
+            nb: {
+              delegations: {},
```
Gozala, line 168: Is this supposed to be empty ? because if it is you're not delegating anything here.

packages/access-api/test/provider-add.test.js
```diff
@@ -97,6 +99,95 @@ for (const accessApiVariant of /** @type {const} */ ([
...
+            await ucanto.delegate({
+              issuer: space,
+              audience: agent,
+              capabilities: [
+                {
+                  can: 'access/delegate',
+                  with: space.did(),
+                },
+              ],
+            }),
```
Gozala, line 183: This isn't used by anyhing as far as I can tell. And `sessionProofs` should be enough to call `access/delegate` with `space.did()`, so motivation here is unclear.

Gozala (changes requested):

packages/access-api/test/provider-add.test.js
gobengo: yes, the specific delegation isn't important. Just want to show that the `access/delegate` invocation is accepted sans error

packages/access-api/test/provider-add.test.js
gobengo: how would session proofs be enough to act on `with=space.did()`?

gobengo: without this proof the test fails because agent is now allowed to `access/delegate` with space

gobengo: we spoke in person and as a result I realized that the `sessionProofs` weren't needed since we have that one, and we removed them https://github.com/web3-storage/w3protocol/pull/483/commits/940d24cd9f2626e69fc2e2210559c4ad9516576d

packages/access-api/test/provider-add.test.js
gobengo: noted. I considered that but did it this way on purpose

gobengo:
we addressed last feedback in person and got to a place of mutual understanding that this is a useful test for the original goal: to ensure that `access/delegate` can be invoked on a space that has had a storage provider registered via `provider/add`

PR #485, opened 2023-03-07
feat: Migrate store/* & upload/* APIs

Gozala:
This pull request moves `store/*` and `upload/*` capability providers from w3infra repo to this one.

The goal is to consolidate implementations so that it's easier to update ucanto and service implementations without having to sync between repos. w3infra should be able to just pull `upload-api` package and expose it through AWS lambdas.

This also migrates tests and makes them agnostic of the runtime. w3infra could run same test suite but pass AWS specific context.


packages/upload-api/test/store.js
```diff
@@ -0,0 +1,645 @@
...
+        with: spaceDid,
+        nb: { link, size: data.byteLength },
+        proofs: [proof],
+      })
+      .execute(connection)
+
+    assert.equal(retryStoreAdd.error, undefined)
+  },
+
+  'store/add fails when size to large to PUT': async (assert, context) => {
```
gobengo, line 323:
```suggestion
  'store/add fails when size too large to PUT': async (assert, context) => {
```

gobengo (approved):

alanshaw (commented):
🚀

PR #506, opened 2023-03-08
fix: incompatibility with w3infra

Gozala:


gobengo:
lgtm so far except tests failing

PR #507, opened 2023-03-08
fix: types so that w3infra would have been evident

Gozala:


packages/upload-api/src/types.ts
```diff
@@ -84,7 +84,13 @@ export interface CarStoreBucket {
   createUploadUrl: (
     link: UnknownLink,
     size: number
-  ) => Promise<{ url: URL; headers: { 'x-amz-checksum-sha256': string } }>
+  ) => Promise<{
+    url: URL
+    headers: {
+      'x-amz-checksum-sha256': string
+      'content-length': string
```
gobengo, line 91: what motivated adding this?

gobengo (approved):

packages/upload-api/src/types.ts
Gozala: Because w3infra passes `content-length` as number which was not included here. This made type checker pass but test fail because it checked against string. If this type existed this would have been caught without back and forth.

PR #509, opened 2023-03-08
fix: switch upload-api to node16

Gozala:
Running tests in w3infra failed because `Blob` is not in node16. Change ci to use node 16 for upload-api to avoid this problem.

### DIFF
--- a/.github/workflows/upload-api.yml
+++ b/.github/workflows/upload-api.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Setup
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 16
           registry-url: https://registry.npmjs.org/
           cache: 'pnpm'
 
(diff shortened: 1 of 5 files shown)


.github/workflows/upload-api.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Setup
         uses: actions/setup-node@v3
         with:
-          node-version: 18
```
gobengo, line 37: I'd love to matrix test w/ both 16 and 18 if possible

gobengo (approved):

.github/workflows/upload-api.yml
Gozala: I'm not sure it is worth it, because doing it in ucanto increases test time significantly. If we ever find that something does not work in node 18 lets add then.

Issue #497, opened 2023-03-07
assert that `space/info` succeeds for spaces w/ storage provider registered via `provider/add` invocation


gobengo:
Motivation:
* @Gozala asked me to
* it'll make it so that when w3infra invokes `space/info` as part of handling `store/add` that its not rejected

Gozala:
Long term we should probably do https://github.com/web3-storage/w3protocol/issues/495, but for now this would unblock the store/* and upload/* with new account stuff.

PR #510, opened 2023-03-08
feat: space/info will not error for spaces that have had storage provider added via provider/add

gobengo:
Motivation:
* #497 
* @Gozala and I talked about doing this minimally just to make sure that when w3infra calls this as part of handling `store/add` that it will work.

### DIFF
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -240,9 +240,9 @@ export function service(ctx) {
  * @returns {Promise<boolean>}
  */
 async function spaceHasStorageProvider(space, spaces, provisions) {
-  return Boolean(
+  return (
     (await spaceHasStorageProviderFromProviderAdd(space, provisions)) ||
-      (await spaceHasStorageProviderFromVoucherRedeem(space, spaces))
+    (await spaceHasStorageProviderFromVoucherRedeem(space, spaces))
   )
 }
 
--- a/packages/access-api/src/service/index.js
+++ b/packages/access-api/src/service/index.js
@@ -1,6 +1,7 @@
 import * as ucanto from '@ucanto/core'
 import * as Ucanto from '@ucanto/interface'
 import * as Server from '@ucanto/server'
+import * as validator from '@ucanto/validator'
 import { Failure } from '@ucanto/server'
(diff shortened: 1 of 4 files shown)


packages/access-api/src/service/index.js
```diff
@@ -219,3 +232,36 @@ export function service(ctx) {
...
 }
+
+/**
+ * @param {Ucanto.DID<'key'>} space
+ * @param {Spaces} spaces
+ * @param {import('../types/provisions.js').ProvisionsStorage} provisions
+ * @returns {Promise<boolean>}
+ */
+async function spaceHasStorageProvider(space, spaces, provisions) {
+  return Boolean(
```
Gozala, line 243: Nit: since other two return booleans another cast here seems redundant

packages/access-api/test/provider-add.test.js
```diff
@@ -187,6 +192,75 @@ for (const accessApiVariant of /** @type {const} */ ([
...
+                  can: 'space/info',
+                  with: space.did(),
+                },
+              ],
+            }),
+          ],
+        })
+        .delegate()
+      const spaceInfoResult = await accessApiVariant.invoke(spaceInfo)
+      assertNotError(spaceInfoResult)
```
Gozala, line 261: Can you also plz check that `spaceInfoResoult.did` is what we expect it is.

packages/access-client/src/types.ts
```diff
@@ -70,6 +70,15 @@ export interface SpaceTable {
 }
 export type SpaceRecord = Selectable<SpaceTable>
 
+export type SpaceInfoResult =
+  // w3up spaces registered via provider/add will have this
+  | {
+      // space did
+      did: DID<'key'>
+    }
+  // deprecated and may be removed if voucher/redeem is removed
```
Gozala, line 79:
```suggestion
  // deprecated and may be removed if voucher/redeem is removed
  /** @deprecated */
```

Gozala (approved):

packages/access-api/src/service/index.js
gobengo: https://github.com/web3-storage/w3protocol/pull/510/commits/ac535ea5e17ab74aba029d2baef7531112fe02b0

Issue #496, opened 2023-03-07
disable github action that deploys PR source branches to 'dev' env 


gobengo:
Why?
* I don't think anyone uses it atm
* It makes PR workflow runs take longer to be green
* It tries to run the feature branches' migrations against the dev env, which is unlikely to work very well. We'll have one shared db state across feature branches that may all have different histories for the order of migrations.

gobengo:
https://github.com/web3-storage/w3protocol/blob/main/.github/workflows/access-api.yml#L44

gobengo:
If we have a need for cloudflare deploys pre-PR-merge, we can accomplish that using cloudflare workers whose names are derrived from the branch name, and which are deleted when the branch is deleted.

example of how: https://github.com/web3-storage/w3protocol/blob/api-gateway/.github/workflows/api-gateway.yml

PR #512, opened 2023-03-08
ci: remove deploy-dev from .github/workflows/access-api.yml

gobengo:
Motivation:
* resolve https://github.com/web3-storage/w3protocol/issues/496


Gozala (approved):

Gozala:
Could you maybe create a workflow with manual dispatch that does this, which may be useful on case by case basis yet does not slow us down otherwise

gobengo:
@Gozala yes let's discuss https://github.com/web3-storage/w3protocol/issues/521#issuecomment-1462930368

Issue #482, opened 2023-03-06
access-api provider/add fails when it has the same CID as an already-invoked invocation


gobengo:
<img width="1240" alt="Screenshot 2023-03-06 at 3 48 10 PM" src="https://user-images.githubusercontent.com/171782/223282581-9f2269bd-205a-4f27-86b0-00dc3907927f.png">

discovered while testing using [this observable](https://observablehq.com/d/95bfec64fbb3c2d1@421#cell-419).

[sentry](https://protocol-labs-it.sentry.io/issues/3981704075/?query=is%3Aunresolved&referrer=issue-stream)
> Error: ERROR 9015: SQL engine error: query error: Error code 1555: A PRIMARY KEY constraint failed (UNIQUE constraint failed: provisions.cid)

todo
* [x] add test to access-api for invoking provider/add twice with same CID
  * assert it leads to no error

PR #517, opened 2023-03-09
fix: DbProvisionsStorage putMany doesnt error on cid col conflict

gobengo:
Motivation:
* https://github.com/web3-storage/w3protocol/issues/482

* [x] first draft

After 2023-03-09 review with @Gozala :
* [x] add test for updating Provision with same invocation cid but different derived 
fields. it should result in error
* [x] make test pass, probably by changing the kysely to be 'on conflict, do nothing as long as the derived columns are identical, otherwise error'
  * it wasn't really possible to do that on conflict logic as far as I could tell, at least through our `kysely` orm
  * kysely doesn't expose methods for all the [`on conflict` clauses that sqlite should support](https://www.sqlite.org/lang_conflict.html)
  * I pursued something like [this using CTE](https://stackoverflow.com/a/49601238), only to find out [sqlite doesn't support `insert` in cte](https://sqlite.org/forum/forumpost/b44ca81f043bed9f)

### DIFF
--- a/packages/access-api/src/models/provisions.js
+++ b/packages/access-api/src/models/provisions.js
@@ -14,9 +14,9 @@ export function createProvisions(storage = []) {
     const hasRowWithSpace = storage.some(({ space }) => space === consumerId)
     return hasRowWithSpace
   }
-  /** @type {Provisions['putMany']} */
-  const putMany = async (...items) => {
-    storage.push(...items)
+  /** @type {Provisions['put']} */
+  const put = async (item) => {
+    storage.push(item)
   }
   /** @type {Provisions['count']} */
   const count = async () => {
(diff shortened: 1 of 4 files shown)


gobengo:
added some notes on why I did it the way I did in issue above

packages/access-api/src/models/provisions.js
```diff
@@ -68,24 +68,66 @@ export class DbProvisions {
...
+      const cause =
+        error && typeof error === 'object' && 'cause' in error && error.cause
+      const code =
+        cause &&
+        typeof cause === 'object' &&
+        'code' in cause &&
+        typeof cause.code === 'string' &&
+        cause.code
+      if (code !== 'SQLITE_CONSTRAINT_PRIMARYKEY') throw error
+      primaryKeyError = error
```
gobengo, line 102: add comment here

gobengo:
@Gozala and I spoke in person and he suggested my last change but said to go ahead and merge this

PR #528, opened 2023-03-10
feat: provision provider type is now the DID of the w3s service

gobengo:
* previously the only allowed storage provider on `provider/add` was a `w3up-alpha` one. I never planned on it being a ucan issuer
* this changes the storage provider id to be the same as the running service, e.g. `did:web:web3.storage` in prod.
  * This makes the `Provision` type generic on `ServiceId`

### DIFF
--- a/packages/capabilities/src/provider.js
+++ b/packages/capabilities/src/provider.js
@@ -11,8 +11,12 @@
 import { capability, DID, literal, struct } from '@ucanto/validator'
 import { equalWith, fail, equal } from './utils.js'
 
-export const StorageProvider = literal(
-  'did:web:web3.storage:providers:w3up-alpha'
+export const Web3StorageId = literal('did:web:web3.storage').or(
+  literal('did:web:staging.web3.storage')
+)
+
+export const Provider = Web3StorageId.or(DID.match({ method: 'key' })).or(
+  DID.match({ method: 'web' })
 )
 
 export const AccountDID = DID.match({ method: 'mailto' })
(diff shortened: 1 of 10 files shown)


packages/capabilities/src/provider.js
```diff
@@ -11,8 +11,12 @@
 import { capability, DID, literal, struct } from '@ucanto/validator'
 import { equalWith, fail, equal } from './utils.js'
 
-export const StorageProvider = literal(
-  'did:web:web3.storage:providers:w3up-alpha'
+export const Web3StorageId = literal('did:web:web3.storage').or(
+  literal('did:web:staging.web3.storage')
+)
```
Gozala, line 16: Nit: Can we keep it simpler by saying it's `did:web` ? I'm not convinced locking down possible options at the type level is going to be of advantage.

gobengo:
> I don't mind it landing it as is, I'm just bit concerned that it might introduce more hassle than help catch us issues, specifically I think on local test runs my DID is different (remember it had .local somewhere)

let's wait on it then. We may not need it.

Gozala:
> let's wait on it then. We may not need it, and can always reopen and use this if/when we do change the provider id

I think I was not very clear with my comment, sorry for that. My concern was regarding:

1. Use of generics as opposed to just less concrete type like `DID<"web">` (I don't mind it however, we could always loosen it up).
2. Similar sentiment about the `Web3StorageId` type which limits it to two specific DIDs which I don't think even accounts to the one used in testing. I think it would be better here to also just use `DID({ method: 'web' })` and just throw if `cap.with !== server.id.did()`.

In other words I think you can either land it as is, or loosen it up a bit and land after.

gobengo:
I'm supportive of getting rid of the literals in provider add capability parser and just using `did:web:*`. let's iterate toward that in main branch

PR #530, opened 2023-03-11
define `access/confirm` handler and use it in ucanto-test-utils registerSpaces + validate-email handler

gobengo:
Previously a lot of logic to handle `access/confirm` was in the `validate-email` flow, since that is the most common place we'd receive that invocation (after clicking email sent by `access/authorize` handler).

However, this logic can be expressed as a `ServiceMethod` on invocation of `access/confirm`, and validate-email can call that.
This allows us to also self-issue `access/confirm` in some tests, e.g. in ucanto-test-utils `registerSpaces` and send it to our service to handle, which is only enabled when node env is TEST for now.

### DIFF
--- a/packages/access-api/src/routes/validate-email.js
+++ b/packages/access-api/src/routes/validate-email.js
@@ -1,6 +1,6 @@
 import {
-  stringToDelegation,
   delegationsToString,
+  stringToDelegation,
 } from '@web3-storage/access/encoding'
 import * as Access from '@web3-storage/capabilities/access'
 import QRCode from 'qrcode'
@@ -11,10 +11,12 @@ import {
   ValidateEmailError,
   PendingValidateEmail,
 } from '../utils/html.js'
-import * as ucanto from '@ucanto/core'
 import * as validator from '@ucanto/validator'
-import { Verifier, Absentee } from '@ucanto/principal'
-import { collect } from 'streaming-iterables'
+import { Verifier } from '@ucanto/principal'
+import * as delegationsResponse from '../utils/delegations-response.js'
+import * as accessConfirm from '../service/access-confirm.js'
+import { provide } from '@ucanto/server'
+import * as Ucanto from '@ucanto/interface'
 
 /**
  * @param {import('@web3-storage/worker-utils/router').ParsedRequest} req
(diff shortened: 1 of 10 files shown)


packages/capabilities/src/types.ts
```diff
@@ -36,6 +36,11 @@ export interface AccessClaimFailure {
   error: true
 }
 
+export interface AccessConfirmSuccess {
+  delegations: Record<string, Ucanto.ByteView<Ucanto.Delegation>>
```
gobengo, line 40: should I make this

```
  delegations: Ucanto.ByteView<Array<Ucanto.Delegation>>
```
?

I did it this way so it is consistent with other times we encode a set of delegations in response (`delegations-response.js` module).

Figured later we can change to another type as same time as response to access/claim. (changing `delegationsResponse.encode/decode` in one place ideally)

packages/access-api/src/routes/validate-email.js
```diff
@@ -151,60 +153,29 @@ async function authorize(req, env) {
...
-    // Store the delegations so that they can be pulled with access/claim
-    // The fact that we're storing proofs chains that we pulled from the
-    // database is not great, but it's a tradeoff we're making for now.
-    await env.models.delegations.putMany(delegation, attestation)
-
-    const authorization = delegationsToString([delegation, attestation])
-    // Send delegations to the client through a websocket
-    await env.models.validations.putSession(authorization, agent.did())
+    if (confirmResult.error) {
+      throw new Error('error confirming')
```
travis, line 173: I forget, does passing `cause` to this do anything? is `confirmResult.error` swallowed if we don't pass it along or does it get printed/preserved somewhere?

packages/access-api/test/helpers/ucanto-test-utils.js
```diff
@@ -92,6 +135,42 @@ export async function spaceRegistrationInvocation(
...
+ * @param {number} lifetimeInSeconds
+ */
+export async function accountRegistrationInvocation(
+  service,
+  account,
+  agent,
+  audience = service,
+  lifetimeInSeconds = 60 * 15
+) {
+  const registerSpace = await Access.confirm
```
travis, line 156: should this be named `registerAccount` ?

packages/access-api/test/helpers/ucanto-test-utils.js
```diff
@@ -92,6 +135,42 @@ export async function spaceRegistrationInvocation(
...
+    .invoke({
+      issuer: service,
+      audience,
+      with: service.did(),
+      lifetimeInSeconds,
+      // We link to the authorization request so that this attestation can
```
travis, line 162: >  We link to the authorization request

what's this referring to? I see the issuer set to be the account, the audience to the agent, and the attestations set to full open, but can't figure out in what sense this is a "link to the authorization request"

packages/access-api/src/service/access-confirm.js
```diff
@@ -0,0 +1,91 @@
...
+    expiration: Infinity,
+  })
+
+  // Store the delegations so that they can be pulled with access/claim
+  // The fact that we're storing proofs chains that we pulled from the
+  // database is not great, but it's a tradeoff we're making for now.
+  await ctx.models.delegations.putMany(delegation, attestation)
+
+  const authorization = delegationsToString([delegation, attestation])
+  // Send delegations to the client through a websocket
```
travis, line 85: this comment is not accurate - all this does is stash the delegation in kv. can probably just drop this comment or rephrase to "send delegations to validations model where it will be made available to claimers"

packages/access-api/src/routes/validate-email.js
```diff
@@ -151,60 +153,29 @@ async function authorize(req, env) {
...
-      ),
-    })
-
-    const attestation = await Access.session.delegate({
-      issuer: env.signer,
-      audience: agent,
-      with: env.signer.did(),
-      nb: { proof: delegation.cid },
-      expiration: Infinity,
+    const confirm = provide(
```
travis, line 157: is this all going to be backwards compatible with the current implementation? backwards compatibility isn't necessarily a hard requirement, but I am interested in tracking what's likely to break for older clients

travis (changes requested):
mostly just clarifying questions, but a couple of requests that would be good to address

packages/access-api/src/routes/validate-email.js
gobengo: My understanding is yes, it will be. All the old tests pass.
(this whole path is only exercised when clicking confirmation email sent by `access/authorize`, so it will be backward compatible with the not-very-many implementations of the client side of that, e.g. the observable I made)

packages/access-api/src/service/index.js
```diff
@@ -45,6 +47,21 @@ export function service(ctx) {
           config: ctx.config,
         })(...args)
       },
+      confirm: Server.provide(
+        Access.confirm,
+        async ({ capability, invocation }) => {
+          // only needed in tests
+          if (ctx.config.ENV !== 'test') {
+            throw new Error(`access/confirm is disabled`)
+          }
```
Gozala, line 56: I would not limit this to tests only.

packages/access-api/src/routes/validate-email.js
gobengo: > does passing cause to this do anything?

Which `cause`?

> is confirmResult.error swallowed if we don't pass it along

yes

packages/access-api/src/service/access-confirm.js
gobengo: this is the same as the comment was before.
I agree that would be a more accurate comment. Open to accepting a suggestion here, or anyone changing it as a followup.

packages/capabilities/src/types.ts
Gozala: I would not bother here and now, but generally sounds good.

Gozala (approved):

packages/access-api/src/routes/validate-email.js
travis: whoop sorry, meant to say "does passing this along as `cause` do anything?"

like:

```
throw new Error('error confirming', {cause: confirmError.result})
```

I think I've seen this used in one or another of our codebases but hadn't confirmed it actually does anything useful

packages/access-api/src/routes/validate-email.js
travis: OH right, the separate codepath for this should make it all work fine, cool!

packages/access-api/src/service/access-confirm.js
travis, line 85:
```suggestion
  // Save delegations for the validation process
```

travis: ok I'd change it to this for now!_

packages/access-api/src/routes/validate-email.js
gobengo: It's a new js feature. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause
And yeah I should probably use it here

gobengo: https://github.com/web3-storage/w3protocol/pull/530/commits/9080cfb57d3aae099b21d48a5597118b99258fba

gobengo: @travis one thing to keep in mind is that the version of typescript in use needs to have updated builtin `Error` types to allow passing `error.cause`. I think most of our repos are upgraded to it now, but some may not be (yet)

packages/access-api/test/helpers/ucanto-test-utils.js
gobengo: ty https://github.com/web3-storage/w3protocol/pull/530/commits/b4fc3c7c9e9c38564487b24d896b3643b06fc5b6

packages/access-api/test/helpers/ucanto-test-utils.js
gobengo: I don't think it makes sense either. It's from copypasta. Good catch. i removed it

packages/access-api/src/routes/validate-email.js
travis: ok thx!

travis (approved):

PR #539, opened 2023-03-14
feat: use new ucanto allows api for `canDelegateCapability` and implement space registration

travis:
* use the new `allows` function in `ucanto` to reimplemeent `canDelegateCapability`
* move space registration from `newCreateSpace` to `registerSpace` where it belongs
  * use agent's current account in registerSpace rather than requiring an email


packages/access-client/src/agent.js
```diff
@@ -43,43 +42,49 @@ function emailToSessionPrincipal(email) {
...
 }
 
 /**
- * @param {Ucanto.Signer<Ucanto.DID<'key'>>} space
+ * @param {Ucanto.Signer<Ucanto.DID<'key'>>} issuer
+ * @param {Ucanto.DID} space
  * @param {Ucanto.Principal<Ucanto.DID<'key'>>} device
  */
-async function createSpaceSaysDeviceCanAccessDelegateWithSpace(space, device) {
+async function createSpaceSaysDeviceCanAccessDelegateWithSpace(
```
gobengo, line 76: may want to call this function now/soon `createIssuerSays...` since the authorization will now no longer always be from Space.

packages/access-client/src/agent.js
```diff
@@ -625,14 +597,14 @@ export class Agent {
...
-        consumer: space.did(),
+        consumer: space,
       },
     })
   }
 
   /**
    *
-   * @param {Signer.EdSigner} space - TODO is this type correct?
+   * @param {Ucanto.DID} space - TODO is this type correct?
```
gobengo, line 607: this can be `Ucanto.DID<'key'>` but also can change after merge

packages/access-client/src/delegations.js
```diff
@@ -62,13 +65,16 @@ export function validate(delegation, opts) {
...
  */
 export function canDelegateCapability(delegation, child) {
-  for (const parent of delegation.capabilities) {
-    // TODO is this right?
-    if (
-      (parent.with === child.with || parent.with === 'ucan:*') &&
-      canDelegateAbility(parent.can, child.can)
-    ) {
-      return true
+  // @ts-ignore ignore this until we bring in ucanto as a normal dependency rather than linking to a local version
```
gobengo, line 68: We should be able to remove this right?

gobengo, line 68:
```suggestion
  
```

gobengo (approved):

PR #550, opened 2023-03-16
feat: Agent#requestAuthorization

gobengo:
Motivation:
* move towards unbundling `authorize()` into e.g. `requestAuthorization()` + (future) `wait.....`

### DIFF
--- a/packages/access-client/src/agent.js
+++ b/packages/access-client/src/agent.js
@@ -487,26 +487,43 @@ export class Agent {
    * signed by the passed email address.
    *
    * @param {`${string}@${string}`} email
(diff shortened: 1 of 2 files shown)


packages/access-client/src/agent.js
```diff
@@ -507,24 +507,6 @@ export class Agent {
     if (res?.error) {
       throw new Error('failed to authorize session', { cause: res })
     }
-
```
gobengo, line 510: don't modify this function, just make a new one called `requestAuthorization`

travis (approved):

PR #554, opened 2023-03-16
chore: test access-client-agent same agent, multiple accounts, addProvider

gobengo:


packages/access-client/src/agent.js
```diff
@@ -476,30 +478,6 @@ export class Agent {
...
-  /**
-   * Request authorization of a session allowing this agent to issue UCANs
-   * signed by the passed email address.
-   *
-   * @param {`${string}@${string}`} email
-   * @param {object} opts
-   * @param {AbortSignal} [opts.signal]
-   * @param {Iterable<{ can: Ucanto.Ability }>} opts.capabilities
-   */
-  async requestAuthorization(email, opts) {
```
gobengo, line 488: note: removing `requestAuthorization` method in favor of a [use case](https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html#use-cases) called `requestAuthorization` that operates on an `AccessClient`, but doesn't require adding a new method to `AccessClient` interface

packages/access-client/src/agent.js
```diff
@@ -535,46 +515,7 @@ export class Agent {
 
     // claim delegations here because we will need an ucan/attest from the service to
     // pair with the session delegation we just claimed to make it work
-    await this.claimDelegations()
-  }
-
-  async claimDelegations() {
```
gobengo, line 541: note: removing `claimDelegations` instance method in favor of `claimDelegations(agent: AccessAgent)` use case fn

gobengo: I want to do this for:
* `addProvider`
* `delegateSpaceAccessToAccount`
* (tbh) `authorize`

travis (approved):

PR #560, opened 2023-03-17
fix: Agent#authorize races websocket and polling access/claim 

gobengo:
TIL:
* the websocket definitely doesn't really work in access-api tests, because access-ws is an entirely different cf worker! it's not running in those access-api tests
* so 100% of the time the websocket immediately terminates with ECONNREFUSED, but afaict that error isn't really bubbled up when Agent calls `await ws.open`. This still doesn't bubble up that error, but does allow for that open to be abortable
* and then authorize() races that websocket and polling access claim. If access/claim polling wins, it aborts the websocket, and everything kinda maybe works

End result
* this re-enables a test against Agent#authorize, which passes due to the polling  on access/claim
  * in the wild maybe the websocket will win the race sometimes

(so it can be used in tests)

### DIFF
--- a/packages/access-client/src/agent.js
+++ b/packages/access-client/src/agent.js
@@ -24,12 +24,13 @@ import {
   validate,
   canDelegateCapability,
 } from './delegations.js'
-import { AgentData, getSessionProofs } from './agent-data.js'
+import { AgentData, getSessionProofs, isSessionProof } from './agent-data.js'
 import { createDidMailtoFromEmail } from './utils/did-mailto.js'
 import {
   addProvider,
   claimDelegations,
   requestAuthorization,
+  expectNewClaimableDelegations,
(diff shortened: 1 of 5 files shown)


packages/access-client/src/agent.js
```diff
@@ -704,7 +754,7 @@ export class Agent {
    */
   async #waitForDelegation(opts) {
     const ws = new Websocket(this.url, 'validate-ws')
-    await ws.open()
+    await ws.open(opts)
```
gobengo, line 757: `authorize()` would hang here because abort signal not passed to this

gobengo: (and even if it were, it wouldn't respect it without the other change I put in there)

travis (approved):

PR #564, opened 2023-03-17
fix: don't let space delegation proofs expire

travis:
otherwise the account will eventually lose access to the space

### DIFF
--- a/packages/access-client/src/agent.js
+++ b/packages/access-client/src/agent.js
@@ -56,7 +56,7 @@ async function createIssuerSaysAccountCanAdminSpace(
     },
   ],
   proofs = [],
-  expiration = Infinity
+  expiration
 ) {
   return ucanto.delegate({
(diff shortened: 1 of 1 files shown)


packages/access-client/src/agent.js
```diff
@@ -54,25 +55,29 @@ async function createIssuerSaysAccountCanAdminSpace(
       with: space,
     },
   ],
-  proofs = []
+  proofs = [],
+  expiration = Infinity
```
gobengo, line 59: can we either:
* not default this to Infinity (it's a risky deafult)
* since this function is kinda just a wrapper around `ucanto.delegate` and iirc we only use this function one time, can we just inline this all at point of use?

travis: I think keeping them as separate functions is a huge win for readability so I'll just get rid of this default

PR #568, opened 2023-03-17
fix: only use spaces with key dids

travis:
this is not ideal, but I chatted with @gozala and we agreed that this makes sense for now

### DIFF
--- a/packages/access-client/src/agent.js
+++ b/packages/access-client/src/agent.js
@@ -744,14 +744,16 @@ export async function addSpacesFromDelegations(access, delegations) {
       agent: access,
     })
   }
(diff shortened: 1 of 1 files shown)


packages/access-client/src/agent.js
```diff
@@ -744,14 +744,16 @@ export async function addSpacesFromDelegations(access, delegations) {
       agent: access,
     })
   }
+  // TODO: we need a more robust way to determine which spaces a user has access to
+  // it may or may not involve look at delegations - will file an issue
```
gobengo, line 748:
```suggestion
  // it may or may not involve look at delegations
```

gobengo (approved):

PR #575, opened 2023-03-20
fix: set currentSpace inside addSpace if currentSpace isn't set

travis:
I noticed this wasn't happening because w3ui's w3console was in a state where the space picker was populated with many spaces, but currentSpace wasn't set.

I think we should avoid this situation at this level because it's most convenient for me, but open to moving this assurance to another level if.

### DIFF
--- a/packages/access-client/src/agent-data.js
+++ b/packages/access-client/src/agent-data.js
@@ -111,12 +111,15 @@ export class AgentData {
 
   /**
    * @deprecated
-   * @param {import('@ucanto/interface').DID} did
+   * @param {import('@ucanto/interface').DID<'key'>} did
    * @param {import('./types').SpaceMeta} meta
    * @param {import('@ucanto/interface').Delegation} [proof]
    */
   async addSpace(did, meta, proof) {
     this.spaces.set(did, meta)
+    if (!this.currentSpace) {
+      this.currentSpace = did
+    }
     await (proof ? this.addDelegation(proof) : this.#save(this.export()))
   }
 
--- a/packages/access-client/src/agent.js
+++ b/packages/access-client/src/agent.js
@@ -754,7 +754,7 @@ export async function addSpacesFromDelegations(access, delegations) {
 
     for (const [did, value] of Object.entries(allows)) {
       if (did.startsWith('did:key') && value['space/*']) {
-        data.addSpace(/** @type {Ucanto.DID} */ (did), {
+        data.addSpace(/** @type {Ucanto.DID<'key'>} */ (did), {
           isRegistered: true,
         })
       }


packages/access-client/src/agent.js
```diff
@@ -754,7 +754,7 @@ export async function addSpacesFromDelegations(access, delegations) {
 
     for (const [did, value] of Object.entries(allows)) {
       if (did.startsWith('did:key') && value['space/*']) {
-        data.addSpace(/** @type {Ucanto.DID} */ (did), {
+        data.addSpace(/** @type {Ucanto.DID<'key'>} */ (did), {
```
alanshaw, line 757: This needs to be `await`ed.

packages/access-client/src/agent-data.js
```diff
@@ -111,12 +111,15 @@ export class AgentData {
...
   /**
    * @deprecated
-   * @param {import('@ucanto/interface').DID} did
+   * @param {import('@ucanto/interface').DID<'key'>} did
    * @param {import('./types').SpaceMeta} meta
    * @param {import('@ucanto/interface').Delegation} [proof]
    */
   async addSpace(did, meta, proof) {
     this.spaces.set(did, meta)
+    if (!this.currentSpace) {
```
gobengo, line 120: It's not obvious to me that calling `addSpace` would also have the side effect of changing `this.currentSpace`, either from the comments or tests.

Why should this have that side effect, instead of leaving this to the caller to decide whether or not to call `Agent#setCurrentSpace`? The latter seems like the most flexible solution.

packages/access-client/src/agent-data.js
```diff
@@ -111,12 +111,15 @@ export class AgentData {
 
   /**
    * @deprecated
-   * @param {import('@ucanto/interface').DID} did
+   * @param {import('@ucanto/interface').DID<'key'>} did
```
gobengo, line 114: It could be useful to indicate *which* did (`space`). (e.g. it might be ambiguous considering the `proof` @param itself is composed of several DIDs)

packages/access-client/src/agent-data.js
gobengo: (and, even if it is desirable to introduce the side effect here, why would this do it via assigning `this.currentSpace =` vs calling `this.setCurrentSpace`?)

packages/access-client/src/agent-data.js
```diff
@@ -111,12 +111,15 @@ export class AgentData {
...
    * @param {import('./types').SpaceMeta} meta
    * @param {import('@ucanto/interface').Delegation} [proof]
    */
   async addSpace(did, meta, proof) {
     this.spaces.set(did, meta)
+    if (!this.currentSpace) {
+      this.currentSpace = did
```
alanshaw, line 121: Side effects make me nervous. I'd feel better about it if the agent class called setCurrent after adding a space(s)...at this level there's no choice for anyone using the data class.

Please if you decide not to move then can you at least add a comment to the method in the agent class to say that it will set current space if one is not set? 🙏

travis:
eh, if both of you feel like adding this here is too icky I'm not gonna do it - will find another place to do this!

alanshaw:
FWIW I think you made the right decision 😘

travis:
thanksssss me too, thx to both of you for pushing this in the right direction!

Issue #571, opened 2023-03-20
store access-api delegation bytes in R2 not D1


gobengo:
## Motivation:
* the [`bytes` column on delegations table](https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/migrations/0004_add_accounts_table.sql#L13) should go away because
  * cloudflare D1 has a relatively small (~100Mb) db storage size
  * we shouldn't put unbounded amount of bytes anywhere in the d1 database
  * delegation `bytes` can be arbitrarily large
  * so in order to lower risk we'll fill up the D1 data store, we shouldn't put delegation `bytes` in the d1 database
* related to: https://github.com/web3-storage/w3protocol/issues/475

### Solutions Considered
* move delegation `bytes` out of D1 and into a non-d1 cloudflare data store
  * pros
    * cheap to try to remove the biggest data from d1 - `bytes`
    * changes max amount of delegations stored from `O(100MB/(avgByteLength * n))` to `O(100Mb/n)`
  * cons
    * reading delegations will be slower. Before the read just had to query D1, but now it will query D1 for delegation CIDs, and then read the bytes for those CIDs from R2
    * migrating this data will be more complicated. When it's in d1, we can use d1 migrations and sql to manage data migrations like e.g. [adding a delegations expiry index](https://github.com/web3-storage/w3protocol/issues/526). If the underlying delegation bytes are not in D1, we will need to make our own data migration system that spans the data stores we do use.
      * in practice today, the d1 migrations tools are alpha, but we will no longer benefit from those tools [if/when cloudflare improves them](https://developers.cloudflare.com/d1/platform/migrations/#plans-for-the-future)
    * non-bytes data will still be in D1 and subject to size limitation
* avoid using d1 entirely
  * this may make sense in the long run, e.g. using a datastore with a much bigger size limit, but I don't think we can do this quickly today without looking beyond cloudflare's data storage products (e.g. to AWS like DyanmoDB or Aurora)
  * pros
    * avoid D1 and its limits entirely
  * cons
    * would involve modifying models/DBs/functionality in access-api other than just `delegations`
    * complexity of using cloudflare workers with non-cloudflare data store
      * e.g. what would localdev use?
* store delegation bytes using web3.storage `store/add` protocol
  * insufficient because the delegations may be non-public, but all things stored in web3.storage are retrievable by CID alone

### data store needs for delegation bytes
  * stored delegations are non-public and we only serve them to authorized invocations e.g. `access/claim` invocations
[...]

Issue #526, opened 2023-03-10
access/claim should not claim expired delegations


gobengo:
probably add an 'expiry' col to 'delegations' table

gobengo:
query by expiry here https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/src/routes/validate-email.js#L186

gobengo:
given audience and `can`, which resources?
query by expiry. query by `with`

gobengo:
I will do this early next week. It's annoying others https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1678498011796599?thread_ts=1678494780.346789&cid=C02BZPRS9HP

PR #578, opened 2023-03-20
feat: move access-api delegation bytes out of d1 and into r2

gobengo:
Motivation:
* first PR on https://github.com/web3-storage/w3protocol/issues/571

todo
* [x] initial `DbDelegationsStorageWithR2` that passes tests as a `DelegationsStorage`
* [x] rm unused access-ws package https://github.com/web3-storage/w3protocol/pull/596
* [x] access-api should use `DbDelegationsStorageWithR2` and not `DbDelegationsStorage` (without r2) and still pass tests https://github.com/web3-storage/w3protocol/pull/599

informed by review
* [x] add r2 bucket name binding to wrangler.toml for staging, production https://github.com/web3-storage/w3protocol/pull/578#discussion_r1145479149

unblocks:
* `DbDelegationsStorageWithR2` supports deletes
  * we don't need this to support current `DelegationsStorage`, but probably will before long, but doesn't need to block this PR
* optimization where r2 stores all the CIDs with varying multibases but for the same hash only once
* safe to remove `DbDelegationsStorage`, `DelegationsV2Row`, `DelegationsV2Tables`


packages/access-api/package.json
```diff
@@ -27,6 +27,8 @@
     "@web3-storage/access": "workspace:^",
     "@web3-storage/capabilities": "workspace:^",
     "@web3-storage/worker-utils": "0.4.3-dev",
+    "it-to-buffer": "^3.0.1",
```
gobengo, line 30: remove these since I think I didnt use them

packages/access-api/package.json
```diff
@@ -55,14 +58,14 @@
     "git-rev-sync": "^3.0.2",
     "hd-scripts": "^4.0.0",
     "is-subset": "^0.1.1",
-    "miniflare": "^2.11.0",
+    "miniflare": "2.12.1",
```
gobengo, line 61: I will try reverting back to `^` range now that I worked out conflicts across versions

packages/access-api/package.json
```diff
@@ -103,6 +106,7 @@
             "IterableIterator",
             "KVNamespace",
             "PromiseLike",
+            "R2Bucket",
```
gobengo, line 109: I think I don't need this now that I found where to import the type from

packages/access-api/src/models/delegations.js
```diff
@@ -5,19 +5,24 @@ import {
...
+
+/**
+ * @template {import('../types/access-api-d1').DelegationsV2Table | import('../types/access-api-d1').DelegationsV3Table} DelegationRow
  * @typedef {Omit<DelegationRow, 'inserted_at'|'updated_at'|'expires_at'>} DelegationRowUpdate
  */
 
 /**
- * @typedef Tables
- * @property {DelegationRow} delegations_v2
+ * @typedef V2Tables
```
gobengo, line 17: Duplicate with `AccessApiD1TablesV2`.
Either way, I will remove 'v2' tables now that there is a working `V3Tables`

packages/access-api/src/models/delegations.js
```diff
@@ -116,14 +117,18 @@ class UnexpectedDelegation extends Error {
...
 /**
- * @param {Pick<DelegationRow, 'bytes'>} row
+ * @param {Pick<import('../types/access-api-d1').DelegationsV2Table, 'bytes'>} row
  * @returns {Ucanto.Delegation}
  */
 function rowToDelegation(row) {
   /** @type {Ucanto.Delegation[]} */
   let delegations = []
+  // kysely/sqlite/d1/various-miniflare-versions sometimes only give an Array here.
+  // (GenericPlugin tries to ensure Uint8Array, but it can't reliably detect when to cast Array -> Uint8Array)
```
gobengo, line 127: When I upgraded miniflare, it changed the way it was returning sqlite bytes from before. Before, the `GenericPlugin` hugo wrote would get the bytes back in a way that `isBuffer` would succeed. Now it doesn't (which is actually nice, because that isBuffer check didn't work on cloduflare anyway).

I should try upgrading kysely-d1 to latest to see how that affects things. https://github.com/aidenwallis/kysely-d1/pull/19

packages/access-api/src/models/delegations.js
```diff
@@ -152,7 +157,7 @@ function rowToDelegation(row) {
 
 /**
  * @param {Ucanto.Delegation} d
- * @returns {DelegationRowUpdate}
+ * @returns {DelegationRowUpdate<import('../types/access-api-d1').DelegationsV2Table>}
  */
 export function createDelegationRowUpdate(d) {
```
gobengo, line 162: remove once v2 table removed

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +195,160 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
     return Uint8Array.from(sqlValue)
   }
 }
+
+/**
+ * @typedef {`delegations_v3`} DelegationsTableWithoutBytesName
```
gobengo, line 200: remove this

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +195,160 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+  /**
+   * @param {import('../types/database').Database<AccessApiD1TablesV3>} db
+   * @param {R2Bucket} dags
+   * @param {keyof AccessApiD1TablesV3} delegationsTableName
+   */
+  // eslint-disable-next-line no-useless-constructor
+  constructor(db, dags, delegationsTableName = delegationsV3Table) {
+    this.#db = db
+    this.#delegationsTableName = delegationsTableName
+    this.#dags = dags
```
gobengo, line 229: add type assertion that `this` implements `DelegationsStorage` type

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +195,160 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+   * store items
+   *
+   * @param  {Array<Ucanto.Delegation>} delegations
+   * @returns {Promise<void>}
+   */
+  async putMany(...delegations) {
+    if (delegations.length === 0) {
+      return
+    }
+    await writeDelegations(this.#dags, delegations)
```
gobengo, line 242: We might succeed here, but then fail on the insert, in which case we will be storing more than is necessary in the R2Bucket, but we should 

If d1 had access to sql transactions, we might be able to use that here to do the write to r1 after trying the sql insert inside a transaction, but we don't so this is I think the best we can do without a much more involved [2pc](https://en.wikipedia.org/wiki/Two-phase_commit_protocol)

packages/access-api/src/models/delegations.js
gobengo: done in https://github.com/web3-storage/w3protocol/pull/598

packages/access-api/src/models/delegations.js
gobengo: done https://github.com/web3-storage/w3protocol/pull/578/commits/51e28356016c0eca719be2c0ba3391999270c99f#diff-d9b54b256c8932e4236c04f8c4011720a34850959bf3643834ab025849f8f4f3R227

packages/access-api/wrangler.toml
```diff
@@ -25,6 +25,10 @@ binding = "__D1_BETA__"
 database_name = "spaces-dev"
 database_id = "7c676e0c-b9e7-4711-97c8-7b1c8eb229ae"
 
+[[r2_buckets]]
```
gobengo, line 28: Assuming this binding name works for everyone, I will need to add a corresponding different bucket_name for each of env.staging, env.production

packages/access-api/src/bindings.d.ts
```diff
@@ -53,6 +49,11 @@ export interface Env {
   SPACES: KVNamespace
   VALIDATIONS: KVNamespace
   W3ACCESS_METRICS: AnalyticsEngine
+  /**
+   * primary r2 bucket used by access-api.
+   * will be used for storing env.models.delegations CARs
+   */
+  ACCESS_API_R2: R2Bucket
```
alanshaw, line 56: Just have a bucket specific to the things it's storing? We can create many of these - we don't have to have only 1 for everything we store.

packages/access-api/wrangler.toml
```diff
@@ -25,6 +25,10 @@ binding = "__D1_BETA__"
...
+binding = "ACCESS_API_R2"
+bucket_name = "access-api-dev-1"
```
alanshaw, line 30: Proposal:
```suggestion
binding = "DELEGATIONS"
bucket_name = "access-api-delegations-dev-0"
```

alanshaw (approved):
LGTM but personally I would name the `binding` and `bucket_name` differently.

packages/access-api/wrangler.toml
gobengo: sgtm! I will create buckets following similar pattern for non-`dev` envs

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +191,169 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+      return
+    }
+    await writeDelegations(this.#dags, delegations, this.#getDagsKey)
+    const values = delegations.map((d) => createDelegationRowUpdateV3(d))
+    // @todo - if this fails, undo writeDelegations that dont need to be stored
+    await this.#db
+      .insertInto(this.#delegationsTableName)
+      .values(values)
+      .onConflict((oc) => oc.column('cid').doNothing())
+      .executeTakeFirst()
```
Gozala, line 240: Nit: Why `executeTakeFirst` ?

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +191,169 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+  /**
+   * @param {import('../types/delegations').Query} query
+   */
+  async *find(query) {
+    const { audience } = query
+    const delegations = this.#delegationsTableName
+    const selection = await this.#db
+      .selectFrom(delegations)
+      .selectAll()
+      .where(`${delegations}.audience`, '=', audience)
```
Gozala, line 257: Nit: Would be a good idea to allow filtering by the expiry as well.

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +191,169 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+      .execute()
+    for await (const row of selection) {
+      yield this.#rowToDelegation(row)
+    }
+  }
+
+  async *[Symbol.asyncIterator]() {
```
Gozala, line 264: Do we need this ? I we don't have we just drop it until we do.

packages/access-api/src/models/delegations.js
Gozala: Also might be a good idea to make `audience` optional field on query.

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +191,169 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+
+  /**
+   * @param {import('../types/delegations').Query} query
+   */
+  async *find(query) {
+    const { audience } = query
+    const delegations = this.#delegationsTableName
+    const selection = await this.#db
+      .selectFrom(delegations)
+      .selectAll()
```
Gozala, line 256:
```suggestion
      .select(['cid'])
```

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +191,169 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+      .where(`${delegations}.audience`, '=', audience)
+      .execute()
+    for await (const row of selection) {
+      yield this.#rowToDelegation(row)
```
Gozala, line 260: Nit: Should we perhaps not throw on first error and try to yield as much as we can and then throw ?

packages/access-api/migrations/0007_add_delegations_v3.sql
```diff
@@ -0,0 +1,19 @@
...
+/*
+goal: add a new table to store delegations in
+which doesn't have a 'bytes' column.
+
+context: we're going to start storing bytes outside of the database (e.g. in R2)
+*/
+
+CREATE TABLE
+    IF NOT EXISTS delegations_v3 (
+        cid TEXT NOT NULL PRIMARY KEY,
```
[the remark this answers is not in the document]
Gozala: Please note if this is a delegation CID or a CAR cid otherwise it's confusing.

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +191,169 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+
+/**
+ * @param {R2Bucket} bucket
+ * @param {Iterable<readonly [key: string, value: Uint8Array ]>} entries
+ */
+async function writeEntries(bucket, entries) {
+  await Promise.all(
+    [...entries].map(async ([key, value]) => {
+      return bucket.put(key, value)
+    })
```
Gozala, line 357:
```suggestion
    [...entries].map(([key, value]) => bucket.put(key, value))
```

packages/access-api/test/delegations-storage.test.js
```diff
@@ -133,3 +81,130 @@ async function createDelegation(opts = {}) {
...
   })
 }
+
+for (const variant of [
+  {
+    name: 'DelegationsStorage with sqlite',
+    ...createDbDelegationsStorageVariant(),
+  },
+  {
+    name: 'DelegationsStorage with sqlite+kv',
```
Gozala, line 91:
```suggestion
    name: 'DelegationsStorage with sqlite+R2',
```

packages/access-api/src/models/delegations.js
```diff
@@ -116,14 +116,14 @@ class UnexpectedDelegation extends Error {
...
- * @param {Pick<DelegationRow, 'bytes'>} row
+ * @param {Pick<import('../types/access-api-cf-db').DelegationsV2Row, 'bytes'>} row
  * @returns {Ucanto.Delegation}
  */
 function rowToDelegation(row) {
   /** @type {Ucanto.Delegation[]} */
   let delegations = []
   try {
-    delegations = bytesToDelegations(row.bytes)
+    delegations = bytesToDelegations(new Uint8Array(row.bytes))
```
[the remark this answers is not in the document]
gobengo: after upgrading wrangler/miniflare/etc, it stopped giving back `Buffer` for `row.bytes` and now gives `Array<number>` :/

packages/access-api/src/bindings.d.ts
gobengo: https://github.com/web3-storage/w3protocol/pull/603

packages/access-api/wrangler.toml
gobengo: https://github.com/web3-storage/w3protocol/pull/603

packages/access-api/src/models/delegations.js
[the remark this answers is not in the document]
gobengo: oh I think I did find a way to ensure it's Uint8Array by here. I will check

packages/access-api/src/models/delegations.js
```diff
@@ -163,6 +163,18 @@ export function createDelegationRowUpdate(d) {
   }
 }
 
+/**
+ * @param {Ucanto.Delegation} d
+ * @returns {DelegationRowUpdate<import('../types/access-api-cf-db').DelegationsV3Row>}
+ */
+export function createDelegationRowUpdateV3(d) {
+  return {
+    cid: d.cid.toV1().toString(),
```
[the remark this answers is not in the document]
gobengo: > and delegations can be encoded in DAG-PB.

is this a typo?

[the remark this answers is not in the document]
gobengo: > Looks like CIDs will retain original base encoding when parsed so if you want to normalize you should pass base32 as an argument.

good catch. I thought it would normalize to base32 for all v1. I will do it explicitly.

packages/access-api/src/models/delegations.js
gobengo: let's add filtering by expiry as followup. We had discussed previously related to https://github.com/web3-storage/w3protocol/issues/526

packages/access-api/src/models/delegations.js
gobengo: I'd prefer to fail fast. afaict there is no good reason to keep going if this errors. something is very wrong

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +191,169 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+    .select((e) => e.fn.count('cid').as('size'))
+    .executeTakeFirstOrThrow()
+  return BigInt(size)
+}
+
+/**
+ * @param {{ cid: string }} ucan
+ */
+function carFileKeyer(ucan) {
+  return /** @type {const} */ (`${ucan.cid.toString()}.car`)
```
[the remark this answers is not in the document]
gobengo: here I expect `ucan.cid` to be a delegation CID not a CAR cid

packages/access-api/src/config.js
```diff
@@ -35,6 +35,12 @@ export function loadConfig(env) {
     }
   }
 
+  if (typeof env.ACCESS_API_R2 !== 'object') {
+    throw new TypeError(
+      `expected env.ACCESS_API_R2 to be an R2Bucket object, but got ${typeof env.ACCESS_API_R2}`
```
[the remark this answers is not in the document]
gobengo: it shouldn't happen under normal circumstances, i.e. there is no wrangler magic to recommend. the `wranger.toml` will set this up in a good way for known envs. This would catch it if you somehow opted out of that calling this function.

packages/access-api/src/models/delegations.js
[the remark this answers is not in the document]
gobengo: I removed it

packages/access-api/migrations/0007_add_delegations_v3.sql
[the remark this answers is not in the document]
gobengo: i noted         /* cidv1 dag-ucan/dag-cbor sha2-256 */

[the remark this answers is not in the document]
gobengo: thought about switching to binary too, but decided against it (this pr at least). strings will be easier to query and copypasta.

packages/access-api/src/models/delegations.js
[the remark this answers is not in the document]
gobengo: ended up removing this function anyway

packages/access-api/src/models/delegations.js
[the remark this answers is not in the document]
gobengo: I added tests asserting the cid kinds for cid column in d1 as well as the cid part of r2 keys https://github.com/web3-storage/w3protocol/pull/578/commits/e379699486b621aa1e87b407f4d30fd95e510799#diff-80f191e42400b9b26178ff477cf9edac5ac2e8106246b9be714368ed719525baR134

packages/access-api/src/models/spaces.js
```diff
@@ -32,7 +32,7 @@ export class Spaces {
       inserted_at: (v) => new Date(v),
       updated_at: (v) => new Date(v),
     })
-    this.d1 = /** @type {Kysely<import('../bindings').D1Schema>} */ (
+    this.d1 = /** @type {Kysely<{ spaces: SpaceTable }>} */ (
```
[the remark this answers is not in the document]
gobengo: I looked and it should be easy to resolve any conflicts, and there may be none. This is just narrowing the type to only what's needed.
It also is paving way for moving kysely/database-related types from access-client (e.g. so we can remove access-client dep on kysely).

packages/access-api/src/models/accounts.js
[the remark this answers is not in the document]
gobengo: see https://github.com/web3-storage/w3protocol/pull/578#discussion_r1145768955

gobengo:
> I'd like R2 path changes so they are not misleading given current naming schema.

I think I addressed this in comments, but if I didn't and we want a different r2 key format, we can make a followup pr and test change without any overhead.

packages/access-api/wrangler.toml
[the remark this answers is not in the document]
gobengo: `bucket` part seemes redundant, but I changed to `w3up-delegations-{env}-{i}`

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +191,169 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+    if (!carBytesR2) {
+      throw new Error(`failed to read car bytes for cid ${cidString}`)
+    }
+    const carBytes = new Uint8Array(await carBytesR2.arrayBuffer())
+    const delegations = bytesToDelegations(carBytes)
+    if (delegations.length !== 1) {
+      throw new Error(
+        `expected 1 delegation in CAR, but got ${delegations.length}`
+      )
+    }
```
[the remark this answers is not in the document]
gobengo: I changed this to just check that `delegations` contains an entry whose cid equals parsed row.cid

packages/access-api/src/models/delegations.js
```diff
@@ -179,3 +191,169 @@ export function delegationsTableBytesToArrayBuffer(sqlValue) {
...
+    .selectFrom(delegationsTable)
+    .select((e) => e.fn.count('cid').as('size'))
+    .executeTakeFirstOrThrow()
+  return BigInt(size)
+}
+
+/**
+ * @param {{ cid: string }} ucan
+ */
+function carFileKeyer(ucan) {
```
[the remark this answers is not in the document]
gobengo: I will change this today in followup

PR #579, opened 2023-03-21
feat: ucan invocation post handled car

vasco-santos:
Adds POST of handled CAR invocations to w3infra's https://github.com/web3-storage/w3infra/pull/171

Implementation details:
- Performed in `waitUntil`. We have no way to rollback previous operations if it fails. We should have a failure tolerance here with CF Queue. Let's create an issue and sort this out in follow up.
- Only makes sure in tests that endpoint returns 200. We need to discuss on how we should perform integration tests for this!

Note that `fetchMock` was not working with current version of `miniflare`. Upgrading miniflare creates issues with D1 which I was not able to solve easily. The issues looked to be with `undici` version behind current miniflare release. Upgraded in the lock file manually just `undici` and is now working! 

Needs:
- [x] secret set

### DIFF
--- a/packages/access-api/src/bindings.d.ts
+++ b/packages/access-api/src/bindings.d.ts
@@ -72,7 +72,7 @@ export interface RouteContext {
     validations: Validations
   }
   uploadApi: ConnectionView
-  uploadApiUrl: URL
+  ucanInvocationPostURL: URL
   ucanInvocationPostBasicAuth: string
 }
 
--- a/packages/access-api/src/bindings.d.ts
+++ b/packages/access-api/src/bindings.d.ts
@@ -46,6 +46,7 @@ export interface Env {
   SENTRY_DSN: string
   POSTMARK_TOKEN: string
   POSTMARK_SENDER?: string
+  UCAN_INVOCATION_POST_BASIC_AUTH: string
 
   DEBUG_EMAIL?: string
   LOGTAIL_TOKEN: string
@@ -71,6 +72,8 @@ export interface RouteContext {
     validations: Validations
   }
   uploadApi: ConnectionView
+  uploadApiUrl: URL
+  ucanInvocationPostBasicAuth: string
 }
 
 export type Handler = _Handler<RouteContext>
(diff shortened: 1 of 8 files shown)


packages/access-api/src/bindings.d.ts
```diff
@@ -71,6 +72,8 @@ export interface RouteContext {
     validations: Validations
   }
   uploadApi: ConnectionView
+  uploadApiUrl: URL
```
alanshaw, line 75: I know these are one and the same but we should confiugure it has a separate service here.

```suggestion
  ucanInvocationPostURL: URL
```

packages/access-api/src/routes/raw.js
```diff
@@ -17,9 +21,24 @@ export async function postRaw(request, env) {
...
 
+  const body = new Uint8Array(await request.arrayBuffer())
   const rsp = await server.request({
-    body: new Uint8Array(await request.arrayBuffer()),
+    body,
     headers: Object.fromEntries(request.headers.entries()),
   })
+
+  // Process CAR with invocations asynchronously
+  ctx.waitUntil(
```
alanshaw, line 31: Should we store only successfully invoked requests? i.e. should we check `res.error`? Can we even? I think no right here.

I assume that later we'll receive a receipt that we can also store so we can know about failures?

cc @Gozala

alanshaw (changes requested):

packages/access-api/src/routes/raw.js
vasco-santos: In the UCAN Stream we will see the contents of the `nb` and if it has an error we will skip it. But, if we can skip it here would be better for sure

packages/access-api/src/routes/raw.js
```diff
@@ -17,9 +21,24 @@ export async function postRaw(request, env) {
...
+    (async () => {
+      await fetch(new URL('/ucan', env.uploadApiUrl), {
```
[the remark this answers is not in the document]
vasco-santos: If we fail we won't rollback what is there as I mentioned in PR description... I would like to see this in a Queue but we still need to get access to CF Queue.

We can wrap in retry and block on the user. But given there is no rollback mechanism I believe operations can then fail forever when user retries given that they already succeeded. Thoughts?

meanwhile adding the retry in waitUntil

vasco-santos:
Closing in favour of https://github.com/web3-storage/w3protocol/pull/592

PR #592, opened 2023-03-21
feat: write invocations and receipts into ucan log

### LINKED ISSUE #581: Pipe invocations and results into w3infra
We need to hook invocations and their results into https://github.com/web3-storage/w3protocol/pull/579

Gozala:
Fixes #581

### DIFF
--- a/packages/access-api/src/bindings.d.ts
+++ b/packages/access-api/src/bindings.d.ts
@@ -63,6 +63,7 @@ export interface RouteContext {
   config: ReturnType<typeof loadConfig>
   url: URL
   email: Email
+  ucanLog: UCANLog
   models: {
     accounts: Accounts
     delegations: Delegations
@@ -71,6 +72,11 @@ export interface RouteContext {
     validations: Validations
   }
   uploadApi: ConnectionView
+}
+
+interface UCANLog {
+  logInvocations: (car: Uint8Array) => Promise<void>
+  logReceipt: (receipt: Uint8Array) => Promise<void>
 }
 
 export type Handler = _Handler<RouteContext>
--- a/packages/access-api/src/routes/root.js
+++ b/packages/access-api/src/routes/root.js
@@ -18,9 +18,54 @@ export async function postRoot(request, env) {
     },
   })
 
-  const rsp = await server.request({
-    body: new Uint8Array(await request.arrayBuffer()),
+  const body = new Uint8Array(await request.arrayBuffer())
+  const invocations = await server.decoder.decode({
+    body,
     headers: Object.fromEntries(request.headers.entries()),
   })
-  return new Response(rsp.body, { headers: rsp.headers })
+
+  const [, ...results] = await Promise.all([
+    env.ucanLog.logInvocations(body),
+    ...invocations.map((invocation) => execute(invocation, server, env)),
+  ])
+
+  const response = await server.encoder.encode(results)
+  return new Response(response.body, {
+    headers: response.headers,
+  })
+}
+
+/**
+ *
+ * @param {Server.Invocation} invocation
+ * @param {Server.ServerView<*>} server
+ * @param {import('../bindings.js').RouteContext} env
+ * @returns {Promise<Server.Result<unknown, Server.API.Failure>>}
+ */
+const execute = async (invocation, server, env) => {
+  /** @type {[Server.Result<*, Server.API.Failure>]} */
+  const [result] = await Server.execute([invocation], server)
+  const out = result?.error ? { error: result } : { ok: result }
+
+  // Create a receipt payload for the invocation conforming to the spec
+  // @see https://github.com/ucan-wg/invocation/#8-receipt
+  const payload = {
+    run: invocation.cid,
+    out,
+    fx: { fork: [] },
+    meta: {},
+    iss: env.signer.did(),
+    prf: [],
+  }
+  // create a receipt by signing the payload with a server key
+  const receipt = CBOR.codec.encode({
+    ...payload,
+    s: await env.signer.sign(CBOR.codec.encode(payload)),
+  })
+  // Send the receipt to the ucan log. Notice that if this fails, the the whole
+  // batch of invocations may fail but there is no way to do a better handling
+  // at this layer. It is up to the `ucanLog` to handle IO errors.
+  await env.ucanLog.logReceipt(receipt)
+  // then we just return the result as is.
+  return result
 }
--- a/packages/access-api/src/utils/context.js
+++ b/packages/access-api/src/utils/context.js
@@ -68,6 +68,10 @@ export function getContext(request, env, ctx) {
     signer,
     config,
     url,
+    ucanLog: {
+      async logInvocations() {},
+      async logReceipt() {},
+    },
     models: {
       delegations: new DbDelegationsStorage(
         createD1Database(config.DB, {


packages/access-api/src/bindings.d.ts
```diff
@@ -73,6 +74,11 @@ export interface RouteContext {
   uploadApi: ConnectionView
 }
 
+interface UCANLog {
+  logInvocations: (car: Uint8Array) => Promise<void>
+  logReceipt: (receipt: Uint8Array) => Promise<void>
```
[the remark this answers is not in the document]
Gozala: Invocations are run concurrently so in order to log receipts together we'd have to await for all of them to finish before we can log receipts. It seems like logging receipt right after invocation is complete is a better idea than having to await for the other invocations.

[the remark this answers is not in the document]
gobengo: > we'd have to await for all of them to finish before we can log receipts

Yikes. I definitely don't want to do that.

> It seems like logging receipt right after invocation is complete is a better idea than having to await for the other invocations.

`logReceipts` could still be called several times as receipts are available (even one at a time). I'm only suggesting that the API afford for batching, not that we wait for all invocations to create only one batch.

vasco-santos:
> I do not seem to have privileges to add vars and secrets in this, so someone with enough privileges should add UCAN_LOG_URL or UCAN_LOG_BASIC_AUTH so someone needs to do it to make this work.

@Gozala set those and made you org owner

vasco-santos (commented):
This is looking good! Thanks for tackling this @Gozala 

I agree we should move this into `ucanto` as follow up. Let me work on the changes in w3infra side before approving to make sure everything fits nicely

packages/access-api/src/routes/root.js
```diff
@@ -18,9 +22,84 @@ export async function postRoot(request, env) {
...
+  const [result] = await Server.execute([invocation], server)
+  const out = result?.error ? { error: result } : { ok: result }
+
+  // Create a receipt payload for the invocation conforming to the spec
+  // @see https://github.com/ucan-wg/invocation/#8-receipt
+  const payload = {
+    ran: invocation.cid,
+    out,
+    fx: { fork: [] },
+    meta: {},
```
vasco-santos, line 90: Currently there is no way to identify which invocation capability was the root for this receipt without opening the CAR file. This is not ideal from a stream of such events.

We should be able to filter out the events we want to handle per capability so that we do not need to read all the CARs to get to know about them. For example, let's consider a stream consumer (with batches of 20) that tracks number of `store/add` successful invocations. This consumer would be required to go through all the receipts received, read all the CAR files from the invocations and check the capability. If from these 20 CARs there are no `store/add` operations to address, all this processing was useless. 

Thinking about adding capability from invocation in `meta` as a good way to flag this. We would read CARs mentioned in receipts with capability we care about (and could still validate that capability is really what is given in `meta`).

Thoughts?

packages/access-api/package.json
```diff
@@ -8,6 +8,7 @@
   "private": true,
   "scripts": {
     "lint": "tsc --build && eslint '**/*.{js,ts}' && prettier --check '**/*.{js,ts,yml,json}' --ignore-path ../../.gitignore",
+    "fix:lint": "eslint --fix '**/*.{js,ts}'",
```
alanshaw, line 11:
```suggestion
    "lint:fix": "eslint --fix '**/*.{js,ts}'",
```

packages/access-api/src/routes/root.js
```diff
@@ -18,9 +22,84 @@ export async function postRoot(request, env) {
...
+  const out = []
+
+  for (const receipt of results) {
+    out.push(receipt.data.out.error || receipt.data.out.ok)
+    // we don't await the logReceipt call because we want to return the response
+    // to the client as soon as possible. We will however keep the worker alive
+    // until we have logged the receipt. If logging fails we will log the error
+    // as we have no way to handle it here nor we can rollback the invocation.
+    forks.push(
+      env.ucanLog.logReceipt(receipt).catch((error) => env.log.error(error))
```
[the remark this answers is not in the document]
alanshaw: The logger should handle that. Create an issue if the logger is not already sending errors to sentry 🙏.

packages/access-api/src/routes/root.js
alanshaw: > Currently there is no way to identify which invocation capability was the root for this receipt without opening the CAR file. This is not ideal from a stream of such events.

It's not that bad...right? It's a local bucket so won't be costly. Can we see how it pans out and change if necessary? I'd rather not rely on something in a `meta` field...

alanshaw (approved):
LGTM - thanks @Gozala

packages/access-api/src/routes/root.js
alanshaw: ...or maybe a better idea - the UCAN log service should validate that the receipt belongs to a stored invocation. It shoudl read the invocation and annotate what it sends to the stream with what you need.

vasco-santos: > ...or maybe a better idea - the UCAN log service should validate that the receipt belongs to a stored invocation. It shoudl read the invocation and annotate what it sends to the stream with what you need.

I think this should be good yes, I will report back once I try it out

vasco-santos: > It's not that bad...right? It's a local bucket so won't be costly. Can we see how it pans out and change if necessary? I'd rather not rely on something in a meta field...

Depends, if we could use [Kinesis filters already](https://github.com/web3-storage/w3infra/blob/main/stacks/ucan-invocation-stack.js#L215) that would be a bumper if we could not filter out consumers right away. But your solution above should work well

packages/access-api/src/routes/root.js
```diff
@@ -18,9 +22,84 @@ export async function postRoot(request, env) {
...
+
+  /** @type {[Server.RequestDecoder, Server.ResponseEncoder]} */
+  // We will use encoder / decoder pair based on the content-type header this
+  // way we do not need to have separate routes.
+  const [decoder, encoder] =
+    headers['content-type'] === 'application/car' ? [CAR, CBOR] : [json, json]
+
+  const invocations = await decoder.decode({
+    body,
     headers: Object.fromEntries(request.headers.entries()),
```
vasco-santos, line 36:
```suggestion
    headers,
```

.github/workflows/reusable-deploy-api.yml
```diff
@@ -24,6 +24,8 @@ on:
         required: true
       LOGTAIL_TOKEN:
         required: true
+      UCAN_LOG_BASIC_AUTH:
```
gobengo, line 27: can this include a description of the format? e.g. in the past I've wondered if it should be base64 encoded or not. If not commented, I assume yes? (i.e. I'd assume it's the value in an http header after 'Basic '.

packages/access-api/src/utils/ucan-log.js
```diff
@@ -0,0 +1,105 @@
...
+   * @param {Uint8Array} car
+   */
+  async logInvocations(car) {
+    await pRetry(
+      () =>
+        fetch(this.url, {
+          method: 'POST',
+          headers: {
+            Authorization: `Basic ${this.auth}`,
+            'Content-Type': 'application/invocations+car',
```
gobengo, line 38: Is this defined anywhere?
nit: If it's ad-hoc, `application/vnd.`... might be appropriate.

packages/access-api/src/utils/ucan-log.js
```diff
@@ -0,0 +1,105 @@
...
+export const debug = () => new UCANLogDebug()
+
+/**
+ * @implements {API.UCANLog}
+ */
+class UCANLog {
+  /**
+   * @param {object} input
+   * @param {URL} input.url
+   * @param {string} [input.auth]
```
gobengo, line 21: it could be helpful to include a comment on what kind of string this is. e.g. should it be base64, base64url, what kind of `auth` (http basic?)

packages/access-api/src/utils/ucan-log.js
```diff
@@ -0,0 +1,105 @@
...
+   */
+  constructor({ url, auth }) {
+    this.url = url
+    this.auth = auth
+  }
+
+  /**
+   * @param {Uint8Array} car
+   */
+  async logInvocations(car) {
```
gobengo, line 31: do tests run in this implementation?

gobengo (approved):
none of my comments should block

.github/workflows/reusable-deploy-api.yml
vasco-santos: Value after Basic encoded as base64

PR #595, opened 2023-03-22
feat: allow multiple providers

alanshaw:
Allows multiple providers to be used with the system and defines web3 and NFT providers in config.


gobengo (approved):
lgtm. I want to make sure @Gozala also approves using these `did:web:nft.storage` as provider DIDs since it's not explicitly mentioned in [this spec](https://github.com/web3-storage/specs/blob/main/w3-provider.md)

packages/access-api/src/types/provisions.ts
```diff
@@ -1,23 +1,21 @@
...
-
 /**
  * action which results in provisionment of a space consuming a storage provider
  */
 export interface Provision<ServiceDID extends Ucanto.DID<'web'>> {
   invocation: Ucanto.Invocation<ProviderAdd>
   space: Ucanto.DID<'key'>
   account: Ucanto.DID<'mailto'>
-  provider: AlphaStorageProvider | ServiceDID
+  provider: ServiceDID
```
Gozala, line 11: Nit: I would personally loose the generic and use `Ucanto.DID<'web'>` here. Which I think would make ts-expect-error in file above go away as well.

packages/access-api/src/types/provisions.ts
```diff
@@ -1,23 +1,21 @@
...
 }
 
 /**
  * stores instances of a storage provider being consumed by a consumer
  */
 export interface ProvisionsStorage<ServiceDID extends Ucanto.DID<'web'>> {
-  service: ServiceDID
+  services: ServiceDID[]
```
Gozala, line 18: Nit: I'd remove these from the interface entirely. When I was designing provider protocol I was assuming that we would have `provider/publish` capability which would allow you to create a provider and consequently allow others to add it if they received `consumer/add` delegation from you. I realize we're far from there, but it still might be good idea to align on it. Implementation could still have such a field.

packages/access-api/src/models/provisions.js
Gozala, line 1: As far as I can tell there are no constraints imposed at the DB layer that would prevent one from adding both nft.storage and web3.storage providers to the same space

https://github.com/web3-storage/w3protocol/blob/52cf7c1f35f01aac66d475d884b87f29348a145c/packages/access-api/migrations/0006_add_storage_provider_consumer_schema.sql#L11-L22

We should impose that constraint at some layer to prevent having spaces with both providers otherwise we'll run into problem where user might add something to a space and we don't know if it's an nft or web3 content.

Gozala (changes requested):
Looks great! But I think we can't land this until we impose constraint that only one of the providers can be added to the space. I provided more detailed comment inline.

My lest urgent concern is that we (as in web3.storage) decides who gets to have a nft.storage provider as opposed to nft.storage itself. It's probably fine if we want to land and unblock work on the client side, but I do think we need to address this sooner than later.

packages/access-api/test/provider-add.test.js
```diff
@@ -1,282 +1,282 @@
...
-          createProviderAddHandler({
-            provisions,
-          })
-        ),
-      }
-    })(),
-  },
-])) {
-  describe(`provider/add ${providerAddHandlerVariant.name}`, () => {
-    it(`can be invoked by did:key`, async () => {
```
Gozala, line 42: Note: This test was removed because it was not testing a right thing and was only passing because it did not used validator.

packages/access-api/test/provider-add.test.js
Gozala, line 1: I have removed bunch of indirection in the tests because those loops with test variants made my head heart and I could not get my head around what what was passed where. Instead I've added some test utility functions so each test can setup whatever it needs test and then test.

alanshaw (commented):
LGTM

Issue #593, opened 2023-03-21
rm /packages/access-ws


gobengo:
Motivation:
* I think it's unused?
* if we don't need it, let's remove it, if only to not have to keep it's `wrangler` et al deps aligned with other packages that are used (dealing with this now adding r2 usage to access-api)

travis:
+1 afaik it's not used - plz delete

PR #596, opened 2023-03-22
chore: remove access-ws and access-client package tests that rely on it

gobengo:
Motivation:
* #593 
* by removing this, [my other PR](https://github.com/web3-storage/w3protocol/pull/578) doesn't have to worry about this package anymore (I had to when I encountered some miniflare/wrangler/workers-types/etc type mismatches when adopting R2Bucket type) 

Notes
* this removes some tests in access-client that tested against this access-ws running in miniflare. If we need those back, we can always checkout the things we want back from these commits. Putting #awake here so it shows up in repo search results
  * My understanding is that the websocket used by access-client agent is different than the access-ws package implemented


travis (approved):
lgtm

PR #602, opened 2023-03-22
fix: call claimAccess at the end of authorizeWithPollClaim

travis:
I'm testing out using `authorizeWithPollClaim` instead of `authorizeWithSocket` to see if it will return more quickly after the email validation link is clicked and therefore better for demos. It worked, but doesn't currently run `claimDelegations` and therefore doesn't infer spaces from the list of claimed delegations.

This feels a bit duplicative because we're already executing `access/claim` several times while we poll, but one more time won't hurt much and this makes it more consistent with `authorizeWithSocket`.

### DIFF
--- a/packages/access-client/src/agent-use-cases.js
+++ b/packages/access-client/src/agent-use-cases.js
@@ -265,7 +265,7 @@ export async function authorizeWithPollClaim(access, email, opts) {
   })
   // Claim delegations here because we will need an ucan/attest from the service to
   // pair with the session delegation we just claimed to make it work.
-  // This is a bit duplicative because we're already invoking `access/claim` in the 
+  // This is a bit duplicative because we're already invoking `access/claim` in the
   // wait function above, but we don't do space inference there so do it one more time
   // for consistency with authorizeWithSocket.
   await claimAccess(access, access.issuer.did(), { addProofs: true })
--- a/packages/access-client/src/agent-use-cases.js
+++ b/packages/access-client/src/agent-use-cases.js
@@ -263,6 +263,12 @@ export async function authorizeWithPollClaim(access, email, opts) {
     ...opts,
     expectAuthorization,
   })
+  // Claim delegations here because we will need an ucan/attest from the service to
+  // pair with the session delegation we just claimed to make it work.
+  // This is a bit duplicative because we're already invoking `access/claim` in the 
+  // wait function above, but we don't do space inference there so do it one more time
+  // for consistency with authorizeWithSocket.
+  await claimAccess(access, access.issuer.did(), { addProofs: true })
 }
 
 /**


packages/access-client/src/agent-use-cases.js
```diff
@@ -263,6 +263,12 @@ export async function authorizeWithPollClaim(access, email, opts) {
     ...opts,
     expectAuthorization,
   })
+  // Claim delegations here because we will need an ucan/attest from the service to
+  // pair with the session delegation we just claimed to make it work.
+  // This is a bit duplicative because we're already invoking `access/claim` in the
+  // wait function above, but we don't do space inference there so do it one more time
+  // for consistency with authorizeWithSocket.
+  await claimAccess(access, access.issuer.did(), { addProofs: true })
```
gobengo, line 271: Can you either define this as a new thing, or else wherever you are tempted to use this, due `authorizeWithPollClaim().then(() => claimAccess(...))`

travis: I have a strong preference for `authorizeWithPollClaim` and `authorizeWithSocket` both behaving the same way here - either they should both automatically `claimAccess` or both not do that. Since `authorizeWithSocket` is already in use in both clients and `authorizeWithPollClaim` is unused I'd like to stick with `authorizeWithSocket`'s semantics for now.

I know you would like to simplify this layer and push responsibility for calling `claimAccess` up to the clients and I'm generally supportive of that, but I think we should open an issue for that refactoring and address it once we're out of crunch mode. 

wdyt?

travis: fwiw I'm also open to defining a new thing, but I think it should be named differently - ie, I think `authorize{WithSocket|WithPollClaim}` should behave the same way - if you'd like to preserve a use case that doesn't `claimAccess` that's fine but lets call it something different to distinguish it from the existing functionality.

travis: one more thought on this - I could call `claimAccess` in `react-keyring/src/providers/Keyring.tsx` but fwiw if feels awkward to be dealing with lower level details like the issuer did - ie, I could go from this:

```tsx
      await authorizeWithPollClaim(agent, email, { signal: controller.signal })
```

to this:

```tsx
      await authorizeWithPollClaim(agent, email, { signal: controller.signal })
      await accessClaim(agent, agent.issuer.did(), {addProofs: true})
```

but that gets this React module in the business of thinking about issuer DIDs and whether it should "add proofs" - I'd rather encapsulate those details down here in the use cases, which suggests to me that we should have some function in here that do what I'm doing for both the websocket and polling cases no matter what.

gobengo: what's important to me is that we maintain there being a way of calling `authorizeWithPollClaim` that does not also do more stuff than that. That way you can compose it with other things. What I don't like about adding more functionality to the function that was there is that it makes the function that was there no longer useful for the case where you don't want to addProofs.

gobengo: >  I could call claimAccess in react-keyring/src/providers/Keyring.tsx but fwiw if feels awkward to be dealing with lower level details like the issuer DID and the addProofs option at that level.

so do it in w3up-client https://github.com/web3-storage/w3ui/issues/295

travis:
> Yes lets remove the side effect from authorizeWithSocket and instead export a authorize function that doesn't expose the implementation details.

Hm, the intention tho was for `authorizeWithSocket` and `authorizeWithClaimPoll` to be drop-in replacements for the old `authorize` function that use different mechanisms for waiting for the email link to be clicked. 

I'm in favor of having a version of these functions that don't do the claim, but would like to keep the current names for backwards compatibility at least in the short term, so maybe something like:

```
authorizeWithSocketNoClaim # executes `access/authorize`, waits on socket
authorizeWithSocket # executes `access/authorize`, waits on socket, executes `access/claim`
authorizeWithPollClaimNoClaim # executes `access/authorize`, waits on socket
authorizeWithPollClaim # executes `access/authorize`, waits on socket, executes `access/claim`
```

I can add big scary comments that talk about the intended future direction here - the names are ugly but honestly that just makes me more confident we'll refactor this soon.


Also worth mentioning - after chatting in Colo this morning I'm inclined to just change the KV polling interval here:

https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/src/routes/validate-ws.js#L39

So I may close this PR for now anyway!

gobengo:
> So I may close this PR for now anyway!

I think there was an element of your motivation to file this which was that `authorizeWithPollClaim` doesn't do what you'd expect it to.
tbh I threw that in just to illustrate that it was possible, and I didn't really realize/expect people would export it from the package and use in other libraries.

Is it okay with you if I remove `authorizeWithPollClaim` it from package exports? (others are free to define something similar in their libraries, and customize it according to their requirements) 

(related, this is why I have a rule to never write `export * from`, but I break my own rule this one time :P and now this happens)

travis:
> Is it okay with you if I remove authorizeWithPollClaim it from package exports? (others are free to define something similar in their libraries, and customize it according to their requirements)

works for me!

Issue #614, opened 2023-03-23
change delegation bytes r2 key to have /delegations prefix


gobengo:
Motivation:
* @Gozala and I think this is best
  * so keys are clearly delegation cids not car cids
  * so we can put other prefixed namespaces in the bucket

PR #617, opened 2023-03-23
feat: r2 keys have /delegations/ prefix

gobengo:
Motivation:
* #614


packages/access-api/src/models/delegations.js
```diff
@@ -71,7 +71,7 @@ export class DbDelegationsStorageWithR2 {
   /** @type {DB} */
   #db
   #delegationsTableName = delegationsV3Table
-  #getDagsKey = carFileKeyer
```
gobengo, line 74: renaming 'keyer' now too because it was a good suggestion by @Gozala from last PR

Gozala (approved):

PR #627, opened 2023-03-23
fix: adjust how access-api uses config to build a ucanLog

gobengo:


packages/access-api/src/utils/context.js
```diff
@@ -36,6 +36,14 @@ export function getContext(request, env, ctx) {
           sender: config.POSTMARK_SENDER,
         })
 
+  if (['staging', 'production'].includes(env.ENV)) {
+    if (!config.UCAN_LOG_BASIC_AUTH) {
+      throw new Error(`config.UCAN_LOG_URL is required but missing`)
```
Gozala, line 41:
```suggestion
      throw new Error(`config.UCAN_LOG_BASIC_AUTH is required but missing`)
```

packages/access-api/src/utils/context.js
```diff
@@ -36,6 +36,14 @@ export function getContext(request, env, ctx) {
...
+    }
+    if (!config.UCAN_LOG_URL) {
+      throw new Error(`config.UCAN_LOG_BASIC_AUTH is required but missing`)
```
Gozala, line 44:
```suggestion
      throw new Error(`config.UCAN_LOG_URL is required but missing`)
```

packages/access-api/src/utils/ucan-log.js
```diff
@@ -97,23 +97,31 @@ class UCANLogDebug {
...
-      throw new Error(
-        `Failed to log receipt for invocation ${receipt.data.ran}: ${error}`
+    const { ucanlog } =
+      /** @type {{ucanlog?: { receipts?: Array<API.ReceiptBlock> }}} */ (
+        globalThis
       )
+    if (typeof ucanlog?.receipts?.push === 'function') {
+      try {
+        ucanlog.receipts.push(receipt)
+      } catch (error) {
```
Gozala, line 120: Do we actually want to error here should we just ignore this ?I'm not sure what that was not the case in first place to be honest.

packages/access-api/src/utils/context.js
```diff
@@ -36,6 +36,14 @@ export function getContext(request, env, ctx) {
           sender: config.POSTMARK_SENDER,
         })
 
+  if (['staging', 'production'].includes(env.ENV)) {
```
vasco-santos, line 39:
```suggestion
  if (['staging', 'production'].includes(config.ENV)) {
```

looks like this is used that way before?

this config with `context.js` and `config.js` is super confusing. We should later on just have one file

packages/access-api/src/utils/ucan-log.js
gobengo: I don't want to ignore any error, I want it to throw.
This throws so I'm happy. (I probably wouldn't wrap the error, but we can change that later)

Gozala (approved):

PR #636, opened 2023-03-23
feat: update access confirmation email

dchoi27:


gobengo (approved):

packages/access-api/src/utils/html.js
```diff
@@ -140,30 +140,24 @@ export const PendingValidateEmail = ({ autoApprove }) => (
...
     <h1>Email Validated</h1>
     <p>{email} was confirmed. You may close this window.</p>
     <div class="box">
       <p>
-        During the beta period, uploads via w3up will only appear via the beta
-        APIs, and not on the web3.storage or NFT.Storage websites, even if the
-        associated email addresses are the same.
+        By registering with either the web3.storage or the NFT.Storage w3up beta, 
+        you agree to the respective Terms of Service 
+        (<a href="https://console.web3.storage/terms">web3.storage ToS</a>,
```
gobengo, line 152: this will link to production TOS even when this page is served on staging, but in practice that seems like NBD

PR #640, opened 2023-03-24
chore: fix massive logo

olizilla:
- make email visually nicer.
- make it so the page doesn't jump when the details are expanded and many other improves.
- add lofi snapshot maker so tweaking that email is less of a pain
- load logo from w3s not github!

Screenshot 2023-03-24 at 11 56 10

Screenshot 2023-03-24 at 11 56 06

License: MIT


alanshaw (approved):
LGTM but linter is angry

PR #666, opened 2023-03-28
feat: get `access/claim` authorization wait function working

travis:
Build on @bengo's `access/claim`-polling-based `waitForDelegationOnSocket` replacement. This will let us replace the websocket-based authorization wait function which makes the UX much more responsive - rather than waiting an indeterminate amount of time after clicking the email link, the process completes almost immediately with this change.

The wait function is also pluggable so we that clients can implement different polling strategies, which should make it easier to test and use a receipt-based version of this in the future.

Added tests for this that use `sinon` to mock out service handlers and ensure they are called the right number of times.


packages/access-client/src/agent-use-cases.js
```diff
@@ -92,20 +92,37 @@ export async function addProvider({ access, space, account, provider }) {
...
+/**
+ * @param {DelegationsChecker} delegationsMatch
  * @param {AccessAgent} access
  * @param {Ucanto.DID} delegee
  * @param {object} [options]
  * @param {number} [options.interval]
  * @param {AbortSignal} [options.abort]
  * @returns {Promise<Iterable<Ucanto.Delegation>>}
  */
-export async function expectNewClaimableDelegations(access, delegee, options) {
```
gobengo, line 102: maybe we should leave `expectNewClaimableDelegations` in to maintain back compat?
idc either way. i.e. if we remove it and need it again, we can add it.
but if we remove it, might want to note that the export will go away (everything in this file is exported, which we might wanna change here by dropping the `export * from "./agent-use-cases.js"` in ./agent.js)

packages/access-client/src/agent-use-cases.js
```diff
@@ -170,6 +185,34 @@ export async function waitForDelegationOnSocket(access, opts) {
   throw new TypeError('Failed to get delegation')
 }
 
+/**
+ * @typedef {{signal?: AbortSignal }} AuthorizationWaiterOpts
+ * @typedef {(accessAgent: AccessAgent, opts: AuthorizationWaiterOpts) => Promise<Iterable<Ucanto.Delegation>> } AuthorizationWaiter
+ */
+
+/**
+ * @type AuthorizationWaiter
```
gobengo, line 194: love having the type assertion in here to automate ensuring which functions are expected to be drop in replacements to one another

gobengo: a line of doc or two on this might helpfully show up in intellisense for the 2+ functions that impl it

packages/access-client/src/agent-use-cases.js
```diff
@@ -201,12 +239,28 @@ export async function authorizeAndWait(access, email, opts) {
...
+ * @param {AccessAgent} accessAgent
+ * @param {`${string}@${string}`} email
+ * @param {object} [opts]
+ * @param {AbortSignal} [opts.signal]
+ * @param {Iterable<{ can: Ucanto.Ability }>} [opts.capabilities]
+ * @param {AuthorizationWaiter} [opts.expectAuthorization] - function that will resolve once account has confirmed the authorization request
+ */
+export async function authorizeWaitAndClaim(accessAgent, email, opts) {
+  await authorizeAndWait(accessAgent, email, opts)
+  await claimAccess(accessAgent, accessAgent.issuer.did(), { addProofs: true })
```
gobengo, line 261: can we make the `addProofs` come from `opts`?
or call this `authorizeWaitAndClaimAndAdd` (claiming and adding are two separable operations)

packages/access-client/src/agent-use-cases.js
gobengo: also fine to me to remove `expectNewClaimableDelegations` here. we can add it back in if needed

packages/access-client/src/agent-use-cases.js
travis: oo I like making it configurable yep

packages/access-client/src/agent-use-cases.js
travis: great call! done

packages/access-client/src/agent-use-cases.js
```diff
@@ -170,6 +189,43 @@ export async function waitForDelegationOnSocket(access, opts) {
...
+/**
+ * Wait for the authorization process to complete by waiting on a
+ * well-known websocket endpoint for the access-api server to
+ * receive and forward a session delegation from the authorization
+ * email flow.
+ *
+ * @type AuthorizationWaiter
+ */
+export async function waitForAuthorizationOnSocket(access, opts = {}) {
+  const delegation = await waitForDelegationOnSocket(access, opts)
```
gobengo, line 206: If we're going to make this a little backwards incompatible, can we just change the `waitForDelegationOnSocket` to match type `AuthorizationWaiter`? and avoid exporting/maintaining both of `waitForDelegationOnSocket` and `waitForAuthorizationOnSocket` with such similar types?

packages/access-client/src/agent-use-cases.js
```diff
@@ -201,12 +252,31 @@ export async function authorizeAndWait(access, email, opts) {
...
+ * @param {object} [opts]
+ * @param {AbortSignal} [opts.signal]
+ * @param {Iterable<{ can: Ucanto.Ability }>} [opts.capabilities]
+ * @param {boolean} [opts.addProofs]
+ * @param {AuthorizationWaiter} [opts.expectAuthorization] - function that will resolve once account has confirmed the authorization request
+ */
+export async function authorizeWaitAndClaim(accessAgent, email, opts) {
+  await authorizeAndWait(accessAgent, email, opts)
+  await claimAccess(accessAgent, accessAgent.issuer.did(), {
+    addProofs: opts?.addProofs ?? true,
```
gobengo, line 276: Without this function having an `AndAdd` at the end, I'd still be surprised for it to default to add.
At the very least, the default to false should be documented on the `opts.addProofs` `@param` jsdoc?

packages/access-client/src/agent-use-cases.js
```diff
@@ -92,20 +92,37 @@ export async function addProvider({ access, space, account, provider }) {
 }
 
 /**
+ * @typedef {(delegations: Ucanto.Delegation<Ucanto.Capabilities>[]) => boolean} DelegationsChecker
+ */
+
+/**
+ * @type DelegationsChecker
+ */
+export function delegationsIncludeSessionProof(delegations) {
```
gobengo, line 101: this is a one liner, do we really need to export it for others?
(might it be fine to export just `isSessionProof`? or neither?)

packages/access-client/src/agent-use-cases.js
travis: yes, though since waitForDelegationOnSocket is definitely in use in a couple places I'd definitely want to do a major version release in this case

packages/access-client/src/agent-use-cases.js
travis: I'd like to give people at least one `DelegationsChecker` implementation that they can use yep - more about providing a publicly exported example of a `DelegationsChecker` so they can implement their own version of `waitForAuthorizationByPolling`. 

I'd be down to hide this, but in that case I don't think it's appropriate to take a `DelegationsChecker` in the publicly exported `waitForAuthorizationByPolling` the way we discussed - I'd want to remove the first parameter there and always pass it `delegationsIncludeSessionProof` - honestly I think this is an inferior option but I'm not strongly against it.

gobengo: > I'd like to give people at least one DelegationsChecker implementation that they can use

sgtm

packages/access-client/src/agent-use-cases.js
travis: tbh I think I'd feel better `@deprecated`ing `waitForDelegationOnSocket` for now and removing it after the next major version release, wdyt?

travis: ok added @deprecated

packages/access-client/src/agent-use-cases.js
travis: Hm, my problem is that from an external point of view, imho, `AndAdd` doesn't add anything useful beyond `Claim` here - there's no reason to claim delegations in a function that returns `void` if you don't save them.

Put another way, `AndAdd` begs the question "add to what" and the answer is an implementation detail that clients don't know anything about. I think the current name is clearer for users of this library and doesn't add a potentially confusing implementation detail to the list of things they need to think about - `Claim` suggests to them that the delegations they're looking for will be available after calling this function. 

All that said having a test here is a good idea, will add!

alanshaw (approved):

packages/access-client/src/agent-use-cases.js
travis: ok test added! https://github.com/web3-storage/w3protocol/pull/666/commits/9146216fe53445b1858108272401a0c19cb6b944

packages/access-client/src/agent-use-cases.js
```diff
@@ -92,20 +92,37 @@ export async function addProvider({ access, space, account, provider }) {
...
+  return delegations.some((d) => isSessionProof(d))
+}
+
+/**
+ * @param {DelegationsChecker} delegationsMatch
  * @param {AccessAgent} access
  * @param {Ucanto.DID} delegee
  * @param {object} [options]
  * @param {number} [options.interval]
  * @param {AbortSignal} [options.abort]
```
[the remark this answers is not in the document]
travis: ah good catch, fixed!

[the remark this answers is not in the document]
travis: oo actually reverting this as it's a public API change - gonna file an issue to make it consistent

gobengo (approved):

Issue #479, opened 2023-03-03
simplify eslint config


gobengo:
* e.g. remove `hd-scripts` config
* https://github.com/web3-storage/w3protocol/actions/runs/4327721071/jobs/7556679430#step:6:13

PR #667, opened 2023-03-28
chore: update w3up-client linter

travis:
make it consistent with other packages in this monorepo so that the top level prettier format command doesn't make unexpected changes when trying to format other packages

### DIFF
--- a/packages/w3up-client/package.json
+++ b/packages/w3up-client/package.json
@@ -58,7 +58,7 @@
     "dist"
   ],
   "scripts": {
-    "lint": "standard",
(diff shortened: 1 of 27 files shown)


packages/w3up-client/package.json
```diff
@@ -89,6 +89,7 @@
     "assert": "^2.0.0",
     "c8": "^7.13.0",
     "docusaurus-plugin-typedoc": "^0.18.0",
+    "hd-scripts": "^5.0.0",
```
gobengo, line 92: medium term I want to remove use of this https://github.com/web3-storage/w3protocol/issues/479

but it does make sense here to be consistent with the other packages ty

gobengo (approved):

Issue #668, opened 2023-03-27
is there a good reason w3up-client tsconfig.json should continue to not include /test?


gobengo:

gobengo:
nope

PR #669, opened 2023-03-28
test: w3up-client tsconfig.json now includes "./test" so its tests have some type checking

gobengo:
Non-goal
* changing src code
* avoidance of all ts-ignore in these newly included files (that can come as iteration)

Context:
* https://github.com/web3-storage/w3protocol/issues/668#issuecomment-1485989307

### DIFF
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -134,7 +134,9 @@ export class Client extends Base {
    * Spaces available to this agent.
    */
   spaces() {
-    return [...this._agent.spaces].map(([did, meta]) => new Space(did, meta))
+    return [...this._agent.spaces].map(([did, meta]) => {
+      return new Space(/** @type {`did:key:${string}`} */ (did), meta)
+    })
   }
 
   /**
@@ -171,7 +173,7 @@ export class Client extends Base {
    */
   async addSpace(proof) {
     const { did, meta } = await this._agent.importSpaceFromDelegation(proof)
-    return new Space(did, meta)
+    return new Space(/** @type {`did:key:${string}`} */ (did), meta)
   }
 
   /**
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -124,10 +124,10 @@ export class Client extends Base {
   /**
    * Use a specific space.
    *
-   * @param {import('./types').DID<'key'>} did
+   * @param {import('./types').DID} did
    */
   async setCurrentSpace(did) {
-    await this._agent.setCurrentSpace(did)
+    await this._agent.setCurrentSpace(/** @type {`did:key:${string}`} */ (did))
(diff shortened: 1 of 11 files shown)


packages/w3up-client/src/client.js
```diff
@@ -124,17 +124,19 @@ export class Client extends Base {
   /**
    * Use a specific space.
    *
-   * @param {import('./types').DID<'key'>} did
```
gobengo, line 127: reverting this ts api interface change https://github.com/web3-storage/w3up-client/commit/ee1cf3a30a79a98c21f2e897d6a26e443b41390f#r106486919

(let's introduce the `DID` -> `DID<'key'>` again in one fell swoop that also adjusts some `@web3-storage/access/agent-data` types that still need to be done)

packages/w3up-client/test/capability/access.test.js
```diff
@@ -33,6 +33,7 @@ describe('AccessClient', () => {
       })
 
       const alice = new Client(await AgentData.create(), {
+        // @ts-ignore
```
gobengo, line 36: I may be able to get rid of these ts-ignore (remember: the whole file was ignored by tsc before) in a followup PR by removing the `voucher` type prop from `@web3-storage/access/types` ServiceType.

travis (approved):
this looks good to me! I'd definitely feel a bit comfier if @alanshaw or @Gozala weighed in since some of this stuff makes changes to capabilities that don't intuitively make sense to me, but it's tests and seems like an improvement over not typechecking this stuff, so +1

gobengo:
> since some of this stuff makes changes to capabilities that don't intuitively make sense to me

@travis it makes changes to some *tests* that have the word 'capability' in their file name, but notably it does not make any changes to the `@web3-storage/capabilities` package, i.e. the canonical capability parsers that sorta define the protocol.

with that said, no worries at all waiting for a review from @alanshaw or @Gozala as well

packages/w3up-client/test/capability/space.test.js
```diff
@@ -51,9 +47,6 @@ describe('SpaceClient', () => {
       assert.equal(service.space.info.callCount, 1)
 
       assert.equal(info.did, space.did())
-      assert.equal(info.agent, alice.agent().did())
-      assert.equal(info.email, 'mailto:alice@example.com')
-      assert.equal(info.product, 'product:test')
```
alanshaw, line 56: Why these removals?

alanshaw (approved):

packages/w3up-client/src/client.js
gobengo: oops this was not 'releasable' due to using `test:` conventional commit

gobengo: This change went out in w3up-client@5.5.0

PR #671, opened 2023-03-29
fix: default authorization abilities

alanshaw:
Until we lift the restriction that we cannot add a space that we don't have admin access (`*`) over please can we default to allowing admin access when we authorize a space?

### DIFF
--- a/packages/access-client/src/agent-use-cases.js
+++ b/packages/access-client/src/agent-use-cases.js
@@ -191,11 +191,7 @@ export async function authorizeAndWait(access, email, opts) {
       })
     }
   const account = { did: () => createDidMailtoFromEmail(email) }
-  await requestAccess(
-    access,
-    account,
-    opts?.capabilities || [{ can: '*' }]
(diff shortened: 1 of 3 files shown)


travis (approved):
This looks good to me, but I would like @Gozala and/or @gobengo to weigh in as I think they had stronger feelings about trying to limit the capabilities we ask for during authorization. 

I don't believe we can grant `*` on a space unless we request `*` here so I don't think there's any way around this for now?

gobengo:
> Until we lift the restriction that we cannot add a space that we don't have admin access (*) over please

What restriction is this referring to? is there a URL for it?

alanshaw:
@gobengo https://github.com/web3-storage/w3protocol/blob/45958cd67a5ecc3f6d1128cfd5040d01633659f3/packages/access-client/src/agent.js#L316-L321

gobengo:
@alanshaw I'm +1 on removing that restriction.

PR #685, opened 2023-03-29
feat: allow importing a space with restricted abilities

alanshaw:
Removes the restriction that importing a space should have `*` capability.

Alternative to https://github.com/web3-storage/w3protocol/pull/671

### DIFF
--- a/packages/access-client/test/agent.test.js
+++ b/packages/access-client/test/agent.test.js
@@ -56,6 +56,46 @@ describe('Agent', function () {
     )
   })
 
+  it('should allow import a space', async () => {
+    const alice = await Agent.create()
+    const bob = await Agent.create()
+
+    const space = await alice.createSpace('videos')
+    await alice.setCurrentSpace(space.did)
+
+    const proof = await alice.delegate({
+      audience: bob,
+      audienceMeta: { name: 'videos', type: 'app' },
+      abilities: ['*']
+    })
+
+    await bob.importSpaceFromDelegation(proof)
+    await bob.setCurrentSpace(space.did)
+
+    const proofs = bob.proofs([{ can: 'store/add', with: space.did }])
+    assert(proofs.length)
+  })
+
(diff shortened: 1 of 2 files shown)


packages/access-client/test/agent.test.js
```diff
@@ -199,7 +239,7 @@ describe('Agent', function () {
       audienceMeta: { name: 'sss', type: 'app' },
     })
 
-    await bob.addProof(delegation)
```
gobengo, line 202: Does it not work with `addProof`? I might expect a new test on `importSpaceFromDelegation` but I don't quite grok why this is a better test now. will you explain?

gobengo (approved):
have a question, but I am happy to have these changes to src and see no blockers

packages/access-client/test/agent.test.js
alanshaw: It's a good question and I should have explained.

Previously this test was circumventing the check that was removed from `importSpaceFromDelegation`. I wanted to use `importSpaceFromDelegation` originally in this test but was unable to, because you couldn't add a space without it being delegated `*`.

It was circumventing the check by adding a proof of `space/info` which `setCurrentSpace` was previously using as a check for it to know whether you could set a space as your current space.

In this PR `setCurrentSpace` just checks the DID against your list of spaces instead. There's two reasons for this:

1. If we're allowing spaces to be added with a specific set of capabilities we can't necessarily guarantee they're always going to be something that `space/info` is derivable from. For example, I'm currently working on a merkle clock implementation where I want to authorize other agents to do `clock/*` on the space, but that capability is not derivable from `space/info`, so I could never set that space as he current space in the agent, or get proofs for `clock/*` invocations.
2. `agent.proofs([cap])` does not currently retrieve proofs for capabilities that are derived from the capability you pass it. This previously worked because we were using `*`.  See https://github.com/web3-storage/w3protocol/blob/main/packages/capabilities/src/utils.js#L103

PR #720, opened 2023-03-31
feat: allow clients to pass uploadFile and uploadDirectory an onUploadProgress callback

travis:
Use the ipfs-utils fetch implementation to enable upload progress reporting.

Users of `upload-client` can now pass an `onUploadProgress` callback to `uploadFile` and `uploadDirectory` which will be handed off to the `fetch` polyfill from `ipfs-utils`. We wrap `onUploadProgress` in order to set `url` on the progress status events we get from `fetch` so that clients can differentiate between upload progress events from different shards, which will be interleaved.


gobengo (changes requested):
1. Could we consider affording for this via allowing user of upload-client to inject their own `fetch` instead of `globalThis.fetch`. Then user could inject the one from ipfs-utils and do whatever they want with it according to their needs, but this library wouldn't need to add a dep on it (just an idea) 
2. Regardless of #1, if you want to add the new options and functionality, will you add tests that asserts how it should behave?
    * Did I hear you right at standup this am that this isn't as useful as originally thought due to many shards, each with their own uploadprogress?

travis:
> 1. Could we consider affording for this via allowing user of upload-client to inject their own `fetch` instead of `globalThis.fetch`. Then user could inject the one from ipfs-utils and do whatever they want with it according to their needs, but this library wouldn't need to add a dep on it (just an idea)

Hmmm, interesting idea... I think this is probably possible, but upload progress reporting does seem like a useful enough feature to make it worth baking into the client rather than pushing the complexity out to the client. I could be convinced 

> 2. Regardless of [Email validator registration and ucan `access/identify`  delegation #1](https://github.com/web3-storage/w3up/issues/1), if you want to add the new options and functionality, will you add tests that asserts how it should behave?

Yes! I've added a couple lines to `store.test.js` that exercise this and make sure we get upload progress events.
    
>    * Did I hear you right at standup this am that this isn't as useful as originally thought due to many shards, each with their own uploadprogress?

Ah sort of - this is actually one of the reasons I think we should bake this into our client: I've just landed some changes to this PR that add a URL to the progress status event so that clients of this library can reliably sort status events they receive into the right buckets and render a progress bar per-shard. I think I'd prefer to provide this functionality here rather than expecting every client to recreate it in their application...

travis:
updated demo video of rendering a progress bar per-shard:

https://user-images.githubusercontent.com/1113/229946475-b082d99b-2fd7-4ae6-9a55-781592f2eab8.mov

packages/upload-client/test/store.test.js
```diff
@@ -62,16 +62,21 @@ describe('Store.add', () => {
...
     })
 
+    let progressStatusCalls = 0
     const carCID = await Store.add(
       { issuer: agent, with: space.did(), proofs, audience: serviceSigner },
       car,
       {
         connection,
+        onUploadProgress: (status) => {
+          progressStatusCalls++
```
gobengo, line 72: nit: will you assert that `typeof status.loaded === 'number' && status.loaded > 0`. That seems like the most useful property for progress bar, but could to assert presence/type of any other properties we expect to be part of long-term public api for this cb.

gobengo (approved):
Thanks for adding a test to catch any regressions I might otherwise accidentally introduce int he future. :) Left a comment with a nonblocking nit.

Would love it if @Gozala had a look too since he suggested the ipfs-utils.

packages/upload-client/test/store.test.js
travis: done!

Issue #623, opened 2023-03-23
access-api stacktraces in sentry are not working


vasco-santos:

gobengo:
related: https://github.com/web3-storage/w3protocol/issues/367

gobengo:
As long as we use d1 which requires `no_bundle=false` (or omitted) in wrangler.toml, I think we'd have to do something like this which pushes source maps to sentry via sentry-cli in our deploy pipeline. https://github.com/robertcepa/toucan-js/tree/master/examples/wrangler-basic#wrangler-basic-starter

gobengo:
I just learned some things:
* there is already some support in the access-api build script to upload the source maps to sentry: https://github.com/web3-storage/w3up/blob/main/packages/access-api/scripts/cli.js#L77
* afaict, `process.env.SENTRY_UPLOAD === 'true'` is true in the github actions that deploy to staging and production. e.g. [this sample deploy seems to have created a release in sentry](https://github.com/web3-storage/w3up/actions/runs/4621902418/jobs/8173939598#step:8:188)
* I see source maps for recent releases in staging and production in the [source maps](https://protocol-labs-it.sentry.io/settings/projects/w3access-api/source-maps/?cursor=10%3A1%3A0&query=&sort=-date_added) settings for access-api in our sentry
* I see [releases](https://protocol-labs-it.sentry.io/releases/?cursor=20%3A0%3A1&project=6536511&sort=date) in sentry from latest deploys
* and [recent errors](https://protocol-labs-it.sentry.io/issues/4030779960/?environment=production&project=6536511&query=is%3Aunresolved&referrer=issue-stream&stream_index=1) seem to show like `release: 5.2.1-production (a4d3409)`
  * So I was surprised that error still doesn't have very much help from source maps.

I noticed this issue https://github.com/web3-storage/w3up/issues/254
mentions upgrading toucan-js, which is the sentry library that access-api uses.
* that [3.0.0 release](https://github.com/robertcepa/toucan-js/releases)
    > This is a complete rewrite of toucan-js. The goal of this update is to reuse more components from [@sentry/core](https://github.com/getsentry/sentry-javascript/tree/master/packages/core), **fix long-standing issues with source maps**, and provide starters using various bundlers (esbuild, rollup, vite, webpack).
* mentions source maps, seems worth trying
[...]

PR #728, opened 2023-04-05
fix: upgrade toucan-js and @web3-storage/worker-utils

gobengo:
Motivation:
* #623 
* test theory that things will work if we upgrade to toucan-js@3, which is a rewrite that sends source maps in a new way


travis (approved):
lgtm - not familiar with toucan but this all seems right as far as I can tell!

Gozala (approved):

PR #737, opened 2023-04-10
feat: change access-api wrangler.toml to be no_bundle=false

gobengo:
Motivation:
* #623 
* verify our assumption that we have to use `no_bundle=true` along with d1.

Experiment
* I wanted to try some new ways of building and whether they would work, but without risking staging, so I used the 'dev' environment from wrangler.toml to try some deploys from local.
* I made access-api use of wrangler more like toucan-js@3 [wrangler-basic](https://github.com/robertcepa/toucan-js/blob/master/examples/wrangler-basic/wrangler.toml)
* I did a sample deploy from my laptop to 'dev' workers environment, with the sentry/toucan `release` called `bengo-dev-0` [sentry errors here](https://protocol-labs-it.sentry.io/issues/?query=+release%3Abengo-dev-0&referrer=issue-list&statsPeriod=14d)
* trigger error via `GET /.debug/error`
  * expectation: this to lead to an in-route error and show up in sentry with helpful stack traces - because this is what [example](https://github.com/robertcepa/toucan-js/blob/master/examples/wrangler-basic/wrangler.toml) implies should work
* trigger error via route that uses d1 binding (e.g. ucanto routes [triggered via w3up+ucanto observable pointing to dev env](https://observablehq.com/d/a76545064f82a998))
  * expectation: d1 error - because supposedly `no_bundle=false` (or omitting `no_bundle`) is incompatible with wrangler d1 bindings

Findings:
* omg it works!
* [error triggered by GET /.debug/error (no d1)](https://protocol-labs-it.sentry.io/issues/4078173859/?query=+release%3Abengo-dev-0&referrer=issue-stream&statsPeriod=14d&stream_index=3)
* [error triggered by using d1 via access protocol via observable](https://protocol-labs-it.sentry.io/issues/4078175792/?query=is%3Aunresolved+release%3Abengo-dev-0&referrer=issue-stream&statsPeriod=14d&stream_index=2)
  * The error did not happen in the `access/authorize` handler. That worked fine, and I got an email from dev env. The error happened when I clicked the email (invoking `access/confirm`), when the `access/confirm` invoacation handler tried to write to d1
  * It got a D1_ERROR, but I could see from the sentry report
    > Error: ERROR 9009: SQL prepare error: no such table: delegations_v3
  * which makes sense, because `dev` doesn't have latest migrations applied that created that table. So I provisioned a new d1 database for this dev env, and ran `npx wrangler --env=dev d1 migrations apply __D1_BETA__`. This would error every couple migrations, but if I re-ran it it would make progress and error again (`Internal error [code: 7501]`), but eventually it would succeed at all migrations.
  * then I re-triggered things via the observable, get email, click email. I saw 'email validated'! 🎉 
  * I continued using the observable to invoke `access/claim` (reads from D1), and got no error all the way through invoking from second device <details>Screenshot 2023-04-10 at 5 48 41 PM</details>

Next steps:
* (opt) feedback on whether we can rely on this from @olizilla 
* develop this PR all the way to
  * make tests pass (tests rely on `./scripts/cli.js` e.g. to run migrations, so I probably need to modify `cli.js` to still work in tests at least)
  * something that will work withour release automation. (locally I had to hardcode env.ACCOUNT_VERSION=bengo-dev-0 just to test the theory)
    * the toucan-js@3 examples hardcode `1.0.0` always. I think I can come up with something better than that quickly tomorrow


packages/access-api/.dev.vars
```diff
@@ -0,0 +1,2 @@
+SENTRY_DSN="https://f8b04d1241bb4b618443b7cc269ee6c1@o609598.ingest.sentry.io/6536511"
```
gobengo, line 1: I tried to gitignore this :/ I disabled this DSN in the sentry admin once I saw this.

packages/access-api/wrangler.toml
```diff
@@ -23,20 +17,22 @@ preview_id = "62e57652625c44a3b1fef2f840ffc882"
 [[d1_databases]]
 binding = "__D1_BETA__"
 database_name = "spaces-dev"
-database_id = "7c676e0c-b9e7-4711-97c8-7b1c8eb229ae"
+database_id = "e3331d3b-d8dd-48b0-ab80-2dba93c2c256"
```
gobengo, line 20: new db because the old one is quite out of date, has no important data, and got hung up on migrations. No one has been using `dev` afaik so I think this should be safe.

packages/access-api/wrangler.toml
```diff
@@ -23,20 +17,22 @@ preview_id = "62e57652625c44a3b1fef2f840ffc882"
...
 
 [[r2_buckets]]
 binding = "DELEGATIONS_BUCKET"
 bucket_name = "w3up-delegations-dev-0"
+preview_bucket_name = "w3up-delegations-dev-0"
```
gobengo, line 25: wrangler asked me to do this when deploy to dev env

packages/access-api/wrangler.toml
```diff
@@ -23,20 +17,22 @@ preview_id = "62e57652625c44a3b1fef2f840ffc882"
...
 
 [vars]
 ENV = "dev"
 DEBUG = "true"
 DID = "did:web:local.web3.storage"
+PRIVATE_KEY="MgCYWjE6vp0cn3amPan2xPO+f6EZ3I+KwuN1w2vx57vpJ9O0Bn4ci4jn8itwc121ujm7lDHkCW24LuKfZwIdmsifVysY="
```
gobengo, line 31: This is from `../../.env.tpl` so already public

gobengo:
I can't think of any downsides. It would accomplish:
* working stack traces in sentry
* access-api use of cloudflare workers is much more like what nftstorage/website and web3-storage/website use

downside
* this uses d1 + `no_bundle=false` - which is technically not allowed in d1 public alpha, but seems like maybe we are detecting it now works even though they haven't updated docs like `https://developers.cloudflare.com/workers/wrangler/bundling/#d1-bindings`

Tomorrow:
* [ ] make sure this is actually using d1... it might be good to be true

gobengo:
I force pushed to this source branch and deleted it because I accidentally pushed a commit containing a sentry dsn I used to test. I disabled the DSN in sentry admin, so it's now useless. I [followed these docs](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository) to delete using `git-filter-repo`.

It should be impossible to find the accidentally pushed DSN, and if it is found, it should be useless

PR #739, opened 2023-04-11
feat: change access-api wrangler.toml to be no_bundle=false. use wrangler bundling

gobengo:
Motivation:
* #623 
* verify our assumption that we have to use `no_bundle=true` along with d1.

Experiment
* I wanted to try some new ways of building and whether they would work, but without risking staging, so I used the 'dev' environment from wrangler.toml to try some deploys from local.
* I made access-api use of wrangler more like toucan-js@3 [wrangler-basic](https://github.com/robertcepa/toucan-js/blob/master/examples/wrangler-basic/wrangler.toml)
* I did a sample deploy from my laptop to 'dev' workers environment, with the sentry/toucan `release` called `bengo-dev-0` [sentry errors here](https://protocol-labs-it.sentry.io/issues/?query=+release%3Abengo-dev-0&referrer=issue-list&statsPeriod=14d)
* trigger error via `GET /.debug/error`
  * expectation: this to lead to an in-route error and show up in sentry with helpful stack traces - because this is what [example](https://github.com/robertcepa/toucan-js/blob/master/examples/wrangler-basic/wrangler.toml) implies should work
* trigger error via route that uses d1 binding (e.g. ucanto routes [triggered via w3up+ucanto observable pointing to dev env](https://observablehq.com/d/a76545064f82a998))
  * expectation: d1 error - because supposedly `no_bundle=false` (or omitting `no_bundle`) is incompatible with wrangler d1 bindings

Findings:
* omg it works!
* [error triggered by GET /.debug/error (no d1)](https://protocol-labs-it.sentry.io/issues/4078173859/?query=+release%3Abengo-dev-0&referrer=issue-stream&statsPeriod=14d&stream_index=3)
* [error triggered by using d1 via access protocol via observable](https://protocol-labs-it.sentry.io/issues/4078175792/?query=is%3Aunresolved+release%3Abengo-dev-0&referrer=issue-stream&statsPeriod=14d&stream_index=2)
  * The error did not happen in the `access/authorize` handler. That worked fine, and I got an email from dev env. The error happened when I clicked the email (invoking `access/confirm`), when the `access/confirm` invoacation handler tried to write to d1
  * It got a D1_ERROR, but I could see from the sentry report
    > Error: ERROR 9009: SQL prepare error: no such table: delegations_v3
  * which makes sense, because `dev` doesn't have latest migrations applied that created that table. So I provisioned a new d1 database for this dev env, and ran `npx wrangler --env=dev d1 migrations apply __D1_BETA__`. This would error every couple migrations, but if I re-ran it it would make progress and error again (`Internal error [code: 7501]`), but eventually it would succeed at all migrations.
  * then I re-triggered things via the observable, get email, click email. I saw 'email validated'! 🎉 
  * I continued using the observable to invoke `access/claim` (reads from D1), and got no error all the way through invoking from second device <details>Screenshot 2023-04-10 at 5 48 41 PM</details>


gobengo:
How I tested this:
* "manual deploy" this source branch to `dev` env
  Screenshot 2023-04-13 at 5 59 51 PM
* created in sentry release [@web3-storage__access-api@6.0.0-dev+0be6257](https://protocol-labs-it.sentry.io/releases/%40web3-storage__access-api%406.0.0-dev%2B0be6257/?project=6536511)
* GET `https://w3access-dev.protocol-labs.workers.dev/.debug/error` to trigger an error
* [error in sentry](https://protocol-labs-it.sentry.io/issues/4078173859/?query=is%3Aunresolved&referrer=issue-stream&stream_index=0) Screenshot 2023-04-13 at 6 04 14 PM

.github/workflows/reusable-deploy-api.yml
```diff
@@ -69,3 +84,18 @@ jobs:
           SENTRY_TOKEN: ${{ secrets.SENTRY_TOKEN }}
           LOGTAIL_TOKEN: ${{ secrets.LOGTAIL_TOKEN }}
           UCAN_LOG_BASIC_AUTH: ${{ secrets.UCAN_LOG_BASIC_AUTH }}
+      - name: create sentry release
+        working-directory: packages/access-api
+        run: |
+          # write ./dist/index.{js,js.map}
+          pnpm exec wrangler --env="$ENV" publish --dry-run --outdir=dist
```
olizilla, line 91: my assumption here is that the step before should already generate these as part of the wrangler action. I'd feel more comfortable here if the artefacts are build once and then sent to the places they need to go, rather than relying on the build definitely always being reproducible (it should be, but, if why tempt fate in our error reporting)

what does `SENTRY_UPLOAD` do in that previous step? is it unused, or does wrangler-action actually upload a thing to sentry (that would be surprising)

packages/access-api/src/utils/context.js
```diff
@@ -44,6 +45,10 @@ export function getContext(request, env, ctx) {
       : UCANLog.debug()
 
   // Sentry
+  // eslint-disable-next-line no-console
+  console.warn('constructing toucan-js', {
```
olizilla, line 49:
```suggestion
  console.log('constructing toucan-js', {
```

packages/access-api/scripts/release.js
```diff
@@ -0,0 +1,52 @@
...
+      const lines = [
+        `/** @type {string|undefined} */`,
+        `export const gitRevShort = '${git.short(__dirname())}'`,
+        `/** @type {string|undefined} */`,
+        `export const name = '${getReleaseName(opts.env)}'`,
+      ]
+      console.log(lines.join('\n'))
+    })
+  cli.parse(argv)
+}
```
olizilla, line 52: this is neat, but note that our other projects do this by injecting these in as variables into the bundle at build time.

packages/access-api/wrangler.toml
```diff
@@ -28,43 +25,52 @@ database_id = "7c676e0c-b9e7-4711-97c8-7b1c8eb229ae"
...
 [build]
-command = "scripts/cli.js build"
 watch_dir = "src"
 
 [miniflare]
 d1_persist = ".wrangler/miniflare"
 
+rules = [
+  { type = "Text", globs = ["package.json"], fallthrough = true }
+]
```
olizilla, line 45: what dis?

packages/access-api/wrangler.toml
```diff
@@ -28,43 +25,52 @@ database_id = "7c676e0c-b9e7-4711-97c8-7b1c8eb229ae"
 [[r2_buckets]]
 binding = "DELEGATIONS_BUCKET"
 bucket_name = "w3up-delegations-dev-0"
+preview_bucket_name = "w3up-delegations-dev-0"
```
olizilla, line 28: some changes in this file seem unrelated to bundling / getting stack traces to work. not a blocker, but can you confirm why these changes are needed in this PR?

olizilla (approved):
Is good. Comments inline but nothing blocking, so approving this on the assumption you will tweak or explain as required.

packages/access-api/wrangler.toml
gobengo: I believe this was required for cloudflare workers to be able to `import` from the package.json.
I think it's a bundler-independent way of providing enough information about whether importing from that json file should result in a string `Text` or a json-parsed object.
https://developers.cloudflare.com/workers/wrangler/configuration/#bundling

packages/access-api/scripts/release.js
gobengo: They do that using `no_bundle=true` and injection via custom esbuild, so this is a way of doing it with `no_bundle=false`.
Yes it's different, but it seems worth it in order to get the stack traces.

If there's another way that works that seems better, lmk and I can try to implement.

.github/workflows/reusable-deploy-api.yml
gobengo: > what does SENTRY_UPLOAD do in that previous step? is it unused, or does wrangler-action actually upload a thing to sentry (that would be surprising)

Nothing atm. (because now `wrangler.toml` doesn't define a custom build command that might read it).

gobengo: > my assumption here is that the step before should already generate these as part of the wrangler action.

Good suggestion. I had thought so too, and tried that, but iirc it didn't write the files.

Let me see if it's just because I need to pass `--outdir=dist` on the previous non-dry-run publish.

packages/access-api/src/utils/context.js
gobengo: I'll remove the console alltogether, as this was useful for debbugging something with `wrangler tail` but no longer

packages/access-api/wrangler.toml
gobengo: I don't think I need it anymore, since now release.node.js imports it in node at build time, and writes a rendered value for cloudflare to import via ESM (so it doesn't need to import the package.json to build the rendered version).

I'll remove it

.github/workflows/reusable-deploy-api.yml
gobengo: Added a commit to only invoke `wrangler publish` once (via `wrangler-action` with custom command to pass `--outdir dist`) https://github.com/web3-storage/w3up/pull/739/commits/3af1c35b4b5158a4c45c346293f7b6b10e97dff2

gobengo:
[error from staging after this deployed there](https://protocol-labs-it.sentry.io/issues/4078173859/?project=6536511&query=is%3Aunresolved&referrer=issue-stream&stream_index=0)

PR #746, opened 2023-04-11
feat!: provider + consumer + subscription 

Gozala:


caps.md
```diff
@@ -0,0 +1,90 @@
...
+  }]
+}
+```
+
+This capability set allows invoker:
+
+1. Insert into `subscription` table record where
+  - `provider` is `with`
+  - `customer` is `nb.customer`
+  - `order` is `nb.order`
```
Gozala, line 60:
```ts
const { cid: order } = CBOR.write({ customer, consumer })
const key = `${order}@${providerDID}`

db.put(key, { provider, customer, order })
```

caps.md
```diff
@@ -0,0 +1,90 @@
...
+
+2. Insert into `consumer` table records where
+- `subscription` is `${nb.order}@${with}`
+- `consumer` is `*`
```
Gozala, line 64:
```ts
const { with: provider, nb: { consumer, order } } = invocation.capability

db.consumer.put(`${order}@${provider}`, { consumer, order, provider })
```

PR #756, opened 2023-04-13
fix: stop polling after authorizeWaitAndClaim is aborted

travis:
the timeout would just keep polling going even after the top level function was aborted - in practice this means we keep polling `access/claim` after a user hits "cancel" in the authorization process

fix this and add a test

### DIFF
--- a/packages/access-client/src/agent-use-cases.js
+++ b/packages/access-client/src/agent-use-cases.js
@@ -217,7 +217,7 @@ export async function waitForAuthorizationByPolling(access, opts = {}) {
     access.issuer.did(),
     {
       signal: opts?.signal,
-      interval: opts?.interval
+      interval: opts?.interval,
     }
   )
   return [...claimed]
--- a/packages/access-client/src/agent-use-cases.js
+++ b/packages/access-client/src/agent-use-cases.js
@@ -191,12 +191,21 @@ export async function waitForDelegationOnSocket(access, opts) {
 }
 
 /**
+ * @template {unknown} [T={}]
+ * @typedef {{ signal?: AbortSignal } & T} AuthorizationWaiterOpts
+ */
+/**
+ * @template {unknown} [U={}]
+ * @typedef {(accessAgent: AccessAgent, opts: AuthorizationWaiterOpts<U>) => Promise<Iterable<Ucanto.Delegation>>} AuthorizationWaiter
+ */
+
+/**
  * Wait for the authorization process to complete by waiting on a
  * well-known websocket endpoint for the access-api server to
  * receive and forward a session delegation from the authorization
  * email flow.
  *
- * @type import('./types.js').AuthorizationWaiter<{}>
+ * @type AuthorizationWaiter
  */
 export async function waitForAuthorizationOnSocket(access, opts = {}) {
   const delegation = await waitForDelegationOnSocket(access, opts)
@@ -208,7 +217,7 @@ export async function waitForAuthorizationOnSocket(access, opts = {}) {
  * `access/claim` capability and waiting for the result to include
  * a session delegation.
  *
- * @type import('./types.js').AuthorizationWaiter<{interval?: number}>
+ * @type AuthorizationWaiter<{interval?: number}>
  */
 export async function waitForAuthorizationByPolling(access, opts = {}) {
   const claimed = await pollAccessClaimUntil(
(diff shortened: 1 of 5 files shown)


packages/access-client/src/types.ts
```diff
@@ -350,3 +351,9 @@ export type InvokeAndExecute = <
     import('./types').Service
   >
 >
+
+export type AuthorizationWaiterOpts<T> = { signal?: AbortSignal } & T
```
gobengo, line 355: Does this need to be generic over `T`? It wasn't before.

```suggestion
export type AuthorizationWaiterOpts = { signal?: AbortSignal }
```

If someone wants to 'mixin' `T`, they can do `AuthorizationWaiterOpts & T`

packages/access-client/src/agent-use-cases.js
```diff
@@ -210,7 +208,7 @@ export async function waitForAuthorizationOnSocket(access, opts = {}) {
  * `access/claim` capability and waiting for the result to include
  * a session delegation.
  *
- * @type AuthorizationWaiter
+ * @type import('./types.js').AuthorizationWaiter<{interval?: number}>
```
gobengo, line 211: previously `AuthorizationWaiter` was only a function that would wait a single time, with no notion of an interval. e.g. when waiting using a socket, there is no interval.

Previously, you could always do `(interval: number) => AuthorizationWaiter`

packages/access-client/src/types.ts
gobengo: I also don't really like moving it to this file at the top of th eproject, because now this type
* is far away from the thing that uses it
* is exported from `@web3-storage/access` and not just `@web3-storage/access/agent`
* doesn't have any jsdoc, so the fact that it's far away from its usage means its less understandable and more likely to be used incorrectly

packages/access-client/src/agent-use-cases.js
```diff
@@ -234,7 +233,7 @@ export async function waitForAuthorizationByPolling(access, opts = {}) {
  * @param {AbortSignal} [opts.signal]
  * @param {boolean} [opts.dontAddProofs] - whether to skip adding proofs to the agent
  * @param {Iterable<{ can: Ucanto.Ability }>} [opts.capabilities]
- * @param {AuthorizationWaiter} [opts.expectAuthorization] - function that will resolve once account has confirmed the authorization request
+ * @param {import('./types.js').AuthorizationWaiter<any>} [opts.expectAuthorization] - function that will resolve once account has confirmed the authorization request
```
gobengo, line 236: * I think we shouldn't make this change because we shouldn't make AuthorizationWaiter generic
* and because it uses `any`. Whenever you are tempted to use `any`, you probaby want `unknown` or something else
  * I would ahve thought we had [this rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-explicit-any.md) enabled that might complain. It's a good rule.

packages/access-client/src/agent-use-cases.js
```diff
@@ -264,9 +263,10 @@ export async function authorizeAndWait(access, email, opts = {}) {
  * @param {`${string}@${string}`} email
  * @param {object} [opts]
  * @param {AbortSignal} [opts.signal]
+ * @param {number} [opts.interval]
```
gobengo, line 266: I don't think we should need to pass this here.
Interval is only needed in order to construct a non-generic `opts.expectAuthorization: AuthorizationWaiter`.

It doesn't seem like it needs to be passed here, because it doesn't need to be passed to `authorizeAndWait` for many (any?) concrete implementation of `expectAuthorization`.

(and then there is no need to change the types)

gobengo (changes requested):

packages/access-client/src/agent-use-cases.js
```diff
@@ -188,8 +191,12 @@ export async function waitForDelegationOnSocket(access, opts) {
 }
 
 /**
- * @typedef {{signal?: AbortSignal }} AuthorizationWaiterOpts
- * @typedef {(accessAgent: AccessAgent, opts: AuthorizationWaiterOpts) => Promise<Iterable<Ucanto.Delegation>> } AuthorizationWaiter
+ * @template {unknown} [T={}]
```
gobengo, line 194:
```suggestion
 * @template [T={}]
```

packages/access-client/src/agent-use-cases.js
```diff
@@ -188,8 +191,12 @@ export async function waitForDelegationOnSocket(access, opts) {
...
+ * @typedef {{ signal?: AbortSignal } & T} AuthorizationWaiterOpts
+ */
+/**
+ * @template {unknown} [U={}]
```
gobengo, line 198:
```suggestion
 * @template [U={}]
```

Unless you explicitly need to constrain the type, it's usually best to omit it

gobengo (approved):
🥇

packages/access-client/src/agent-use-cases.js
travis: oh yea agree - for some reason I thought it was giving me errors without this, but it's working now, whatever I was seeing must have been a red herring, good catch!

alanshaw (approved):

Issue #749, opened 2023-04-11
create capability parsers for an assertion invocation


gobengo:
Why?
* it will result in a building block for applications that process these invocations e.g. for content claims in #743

gobengo:
@Gozala and I just discussed that this is a good idea and will unblock #743

gobengo:
old

PR #759, opened 2023-04-14
feat: 749 create capability parsers for an assertion invocation

gobengo:
Motivation:
* #749


packages/discovery/test/discovery.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  })
+  test('can discovery/assert/location', async () => {
+    const id = await principal.ed25519.generate()
+    const service = discovery.service.create()
+    const server = Server.create({
+      codec: CAR.inbound,
+      id,
+      service,
+    })
+    const claimResult = await server.run(invoke({
```
gobengo, line 30: this works at runtime but not tsc time @Gozala 

I modeled it after this test https://github.com/web3-storage/ucanto/pull/274/files/50f00605e89bd57656aaee6de26316e2d385732c#diff-cdde9fc2d29ed4b39d6ac58052c126c2c661ea3a244c844a0d75ee54394aff98R422

packages/discovery/test/discovery.spec.js
```diff
@@ -0,0 +1,41 @@
...
+      issuer: id,
+      audience: id,
+      capability: {
+        can: /** @type {const} */ ('discovery/assert/location'),
```
Gozala, line 34: Nit: you should not need this annotation here, invoke takes extra steps to make inference treat this as a literal.

packages/discovery/package.json
```diff
@@ -0,0 +1,77 @@
...
+  },
+  "devDependencies": {
+    "@types/mocha": "^10.0.1",
+    "@types/node": "^18.11.18",
+    "@typescript-eslint/eslint-plugin": "^5.58.0",
+    "@typescript-eslint/parser": "^5.58.0",
+    "eslint": "^8.38.0",
+    "eslint-plugin-jsdoc": "^41.1.1",
+    "mocha": "^10.2.0",
+    "typescript": "^5.0.4"
```
Gozala, line 43: This could be causing problems with TS as ucanto is still on [4.9.x](https://github.com/web3-storage/ucanto/blob/main/packages/server/package.json#L50) and there may have been a change in TS that affects inference.

packages/discovery/src/location.js
```diff
@@ -0,0 +1,20 @@
...
+  nb: Schema.struct({}),
+})
+
+export const invoke = async () => {
+  return {
+    ok: {},
+  }
+}
+
+export const provider = provide(capability, invoke)
```
Gozala, line 18: 🤔 because inovoke above has no annotation and `provide` here does not have `invoke` inline it could mislead TS inference. When you inline TS inference will consider `capability` type to infer `invoke` however because `invoke` is already defined TS had inferred it and will no longer update type info for it. I don't know if it would affect `provider` type in this instance, but that is a possibility.

packages/discovery/test/discovery.spec.js
Gozala: It's unclear to me what could be the cause here, but I would suggest:

1. Adding a test to ucanto directly that basically does this & if fails just try to simplify it as much as possible. If it does not fail there it could be a TSC version which is 5.x while ucanto is still on 4.9
2. I made comment on the line where you define a provider for the service in regards to some inference considerations. It may be a cause here, but worth trying to eliminate.
3. The `const` annotation below could also be problem. Ucanto functions were written so it they won't be unnecessary, yet perhaps presence does affect it somehow. Worth eliminating this possibility.
4. Call to invoke here is inlined inside a call to `.run` which in TS affects how inference will work and here it could mean that TS considers `service` type to infer the result of `invoke` in a way that causes problems when it later tries to unify types. Might be worth trying not to inline things and trying it that way.

If it's none of the above, I have no idea why would this not work and would recommend going with in-process serve / client setup [like in many other tests](https://github.com/web3-storage/ucanto/blob/b330244d1505141afb5c53f299b9e923e716e495/packages/server/test/handler.spec.js#L109-L149) unlike `.run` code path it had been around for long time and most (ucanto) tests use that setup. That is to say `.run` is new and perhaps has problems.

Gozala:
I also just looked at TS output from the CI

```
test/discovery.spec.js(30,42): error TS2345: Argument of type 'IssuedInvocationView<{ can: "discovery/assert/location"; with: 'did:key:${string}'; nb: {}; }>' is not assignable to parameter of type 'ServiceInvocation<{ can: "discovery/assert/location"; with: 'did:key:${string}'; nb: {}; }, { discovery: { assert: { location: ServiceMethod<Capability<"discovery/assert/location", 'did:${string}:${string}' & 'did:${string}' & Phantom<{ protocol: "did:"; }>, InferStruct<...>>, {}, {}>; }; }; }>'.
   Type 'IssuedInvocationView<{ can: "discovery/assert/location"; with: 'did:key:${string}'; nb: {}; }>' is not assignable to type 'IssuedInvocation<{ can: "discovery/assert/location"; with: 'did:key:${string}'; nb: {}; }> & IssuedInvocation<any> & (() => IterableIterator<string>)'.
     Type 'IssuedInvocationView<{ can: "discovery/assert/location"; with: 'did:key:${string}'; nb: {}; }>' is not assignable to type '() => IterableIterator<string>'.
       Type 'IssuedInvocationView<{ can: "discovery/assert/location"; with: 'did:key:${string}'; nb: {}; }>' provides no match for the signature '(): IterableIterator<string>'.
```

Specifically first line looks interesting first line of error seems to suggest that `ServiceInvocation` type was inferred properly. However on the next error line that type is gets unpacked not into `IssuedInvocation<{ can: "discovery/assert/location"; with: 'did:key:${string}'; nb: {}; }> & IssuedInvocation<any>` as I would expect [per type definition](https://github.com/web3-storage/ucanto/blob/main/packages/interface/src/lib.ts#L548-L551) but it also has `& (() => IterableIterator<string>)` which is probably why it reports type missmatch.

So I'm guessing [`ServiceInvocations`](https://github.com/web3-storage/ucanto/blob/b330244d1505141afb5c53f299b9e923e716e495/packages/interface/src/lib.ts#L735-L744) somehow gets to infer `IssuedInvocation<any> & (() => IterableIterator<string>)` somehow as opposed to expected `IssuedInvocation<any> & (IssuedInvocation<...> | IssuedInvocation<...>)` 

I suspect ts 5.x changed something there or maybe introduced a bug

packages/discovery/test/discovery.spec.js
gobengo: Got it. I had gotten rid of it in a follow test anyway.

gobengo:
I paused on this but forgot to close the issue.

PR #760, opened 2023-04-17
Revert "feat: change access-api wrangler.toml to be no_bundle=false. use wrangler bundling (#739)"

travis:
This reverts commit d6595169bba974428899a8fd18f596718aaa5e41.

I believe this commit resulted in the Postmark API key not being set in staging and production - reverting to test.


travis:
alternatively, I think the underlying issue should be fixed by adding `POSTMARK_TOKEN` back to https://github.com/web3-storage/w3up/commit/d6595169bba974428899a8fd18f596718aaa5e41#diff-3fca523d2fd08b3c29def7f384792f39a5186fd638e4f9c5b50508a7e44e4c0eR18 - because it's no longer in the `required` list we don't copy it into the `vars` hash and as a result it doesn't get picked up by the code at https://github.com/web3-storage/w3up/commit/d6595169bba974428899a8fd18f596718aaa5e41#diff-3fca523d2fd08b3c29def7f384792f39a5186fd638e4f9c5b50508a7e44e4c0eR41

travis:
fixed in https://github.com/web3-storage/w3up/pull/761

PR #761, opened 2023-04-17
fix: add POSTMARK_TOKEN back to list of required env

travis:
The service is not fully operational without this environment variable - it's unable to send emails to authorize users and so is unusable for anyone who does not already have an existing auth session.

I think it was a mistake to remove this - if it was intentional, then we also need to update the reference to `vars.POSTMARK_TOKEN` below to refer to `env.POSTMARK_TOKEN`.

### DIFF
--- a/packages/access-api/src/config.js
+++ b/packages/access-api/src/config.js
@@ -15,7 +15,7 @@ export function loadConfig(env) {
   const vars = {}
 
   /** @type {Array<keyof env>} */
-  const required = ['DID', 'ENV', 'DEBUG', 'PRIVATE_KEY']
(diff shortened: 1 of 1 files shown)


gobengo (approved):
This works. I verified this var is set in github secrets for env={dev,staging,production}, so I expect this will always be set in our three main environments.

gobengo:
(alternatively could have left it not required, and sourced the maybe-undefined value from `env.POSTMARK_TOKEN`, but this is just as valid afaic)

travis:
> alternatively could have left it not required, and sourced the maybe-undefined value from env.POSTMARK_TOKEN, but this is just as valid afaic

yep - I do like having a big error at startup time if it's not set so I have a soft preference for doing it this way since that's already in place

PR #763, opened 2023-04-17
fix: account for edge cases in polling abort

travis:
My previous implementation (https://github.com/web3-storage/w3up/pull/756) relied on canceling a timeout to stop the polling process. This had a number of edge cases, including one where the signal was aborted before first round of polling happened, which is quite common since we do some network calls before we even get to this code.

Stop worrying about canceling the timeout since we'll just stop polling whenever we execute `poll` in this new implementation - generally this relies on a piece of "persistent" state (in the sense that `signal.aborted` will continue to be true) rather than the assumption that the correct timeout will get canceled, so is quite a bit more robust.

### DIFF
--- a/packages/access-client/src/agent-use-cases.js
+++ b/packages/access-client/src/agent-use-cases.js
@@ -117,7 +117,7 @@ export async function pollAccessClaimUntil(
   const interval = opts?.interval || 250
   const claimed = await new Promise((resolve, reject) => {
     opts?.signal?.addEventListener('abort', (e) => {
-      reject(new Error('pollAccessClaimUntil aborted', { cause: e }))
+      reject(new Error('access/claim polling aborted', { cause: e }))
     })
     poll(interval)
     /**
@@ -125,7 +125,7 @@ export async function pollAccessClaimUntil(
      */
     async function poll(retryAfter) {
       if (opts?.signal?.aborted) {
-        reject(new Error('pollAccessClaimUntil aborted'))
+        reject(new Error('access/claim polling aborted'))
       } else {
         const pollClaimResult = await access.invokeAndExecute(
           w3caps.Access.claim,
--- a/packages/access-client/src/agent-use-cases.js
+++ b/packages/access-client/src/agent-use-cases.js
@@ -115,13 +115,10 @@ export async function pollAccessClaimUntil(
   opts
 ) {
   const interval = opts?.interval || 250
-  /** @type {ReturnType<typeof setTimeout>} */
-  let pollingTimeoutId
   const claimed = await new Promise((resolve, reject) => {
     opts?.signal?.addEventListener('abort', (e) => {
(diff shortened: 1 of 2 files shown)


packages/access-client/src/agent-use-cases.js
```diff
@@ -115,38 +115,37 @@ export async function pollAccessClaimUntil(
...
   const interval = opts?.interval || 250
-  /** @type {ReturnType<typeof setTimeout>} */
-  let pollingTimeoutId
   const claimed = await new Promise((resolve, reject) => {
     opts?.signal?.addEventListener('abort', (e) => {
-      pollingTimeoutId && clearTimeout(pollingTimeoutId)
-      reject(
-        new Error('pollAccessClaimUntilSessionProof aborted', { cause: e })
-      )
+      reject(new Error('pollAccessClaimUntil aborted', { cause: e }))
```
alanshaw, line 120: I'm not sure we should put function names in error messages. These kinds of things often get shown to users. Can we put a human readable message and let the stack trace take care of identifying the origin?

packages/access-client/src/agent-use-cases.js
```diff
@@ -115,38 +115,37 @@ export async function pollAccessClaimUntil(
...
+          const claimedDelegations = Object.values(
+            pollClaimResult.delegations
+          ).flatMap((d) => bytesToDelegations(d))
+          if (delegationsMatch(claimedDelegations)) {
+            resolve(claimedDelegations)
+          } else {
+            setTimeout(() => poll(retryAfter), retryAfter)
+          }
+        } catch (error) {
+          reject(error)
```
alanshaw, line 147: I don't understand why so complicated? Specifically, the delayed recursive function call and mixing `await` with `resolve`/`reject`. Why not just a `while` loop?

```js
while (true) {
  if (opts?.signal?.aborted) throw opts.signal.reason ?? new Error('operation aborted')
  const res = await access.invokeAndExecute(w3caps.Access.claim, { with: delegee })
  if (res.error) throw res
  const claims = Object.values(res.delegations).flatMap(d => bytesToDelegations(d))
  if (delegationsMatch(claims)) return claims
  await new Promise(resolve => setTimeout(resolve, retryAfter))
}
```

alanshaw (changes requested):

packages/access-client/src/agent-use-cases.js
alanshaw: Not blocking btw, looks like what you have here will fix the issue just fine 😇

travis: hm no I agree tho, I'll take a crack at reworking this a bit - I didn't really touch the structure of this code during this fix but it is a bit odd and I'm not sure it's needed - will take your snippet for a spin!

travis: ok yea this worked great, thanks!!

alanshaw (approved):

PR #764, opened 2023-04-19
fix: adjust shard size

alanshaw:
As discussed with @ribasushi, this is a better shard size to set to avoid excessive null padding in the aggregated piece.

### DIFF
--- a/packages/upload-client/src/sharding.js
+++ b/packages/upload-client/src/sharding.js
@@ -2,7 +2,7 @@ import Queue from 'p-queue'
 import { encode } from './car.js'
 import { add } from './store.js'
 
-const SHARD_SIZE = 1024 * 1024 * 100
+const SHARD_SIZE = 1024 * 1024 * 120
 const CONCURRENT_UPLOADS = 3
 
 /**


vasco-santos (approved):

vasco-santos:
@alanshaw should we also limit at the implementation (and spec) level the max shard also according to what we talk with Riba? as https://github.com/web3-storage/w3infra/issues/123

packages/upload-client/src/sharding.js
```diff
@@ -2,7 +2,7 @@ import Queue from 'p-queue'
 import { encode } from './car.js'
 import { add } from './store.js'
 
-const SHARD_SIZE = 1024 * 1024 * 100
+const SHARD_SIZE = 1024 * 1024 * 120
```
ribasushi, line 5: Is the shard case the final size with all the overheads of a car? Or is it something in between?

I am basically highlighting you *might* be leaving a bit more bytes on the table than you'd want.

From Fil's proofs standpoint you can do exactly up to (and inclusive) 133,169,152 bytes ( or 127 * 1024 * 1024 )

alanshaw: The size here is the size of the blocks in the CAR, no CAR related overheads.

vasco-santos:
I think that we will need to tie this limit with the `commP` calculation. That is a bit annoying if we can support at protocol level `commP` being given async after upload is accepted. 

Is there a way we can get to a safe limit without calculating commp to have piecePadding?

cc @Gozala

packages/upload-client/src/sharding.js
vasco-santos: I think that we will need to tie this limit with the commP calculation. That is a bit annoying if we can support at protocol level commP being given async after upload is accepted.

Is there a way we can get to a safe limit without calculating commp to have piecePadding?

cc @Gozala

PR #787, opened 2023-05-03
feat: w3 aggregate protocol client and api implementation

### LINKED ISSUE #772: Implement w3-aggregation API package
Per https://github.com/web3-storage/specs/pull/51

vasco-santos:
This PR is a first iteration for w3 aggregate client and API implementations. It follows same patterns we identified while working on `upload-api` and `upload-client` and is agnostic from the deployment service. `aggregate-api` exposes a lib and a set of tests that implementers with infra should follow, as well as the needed interfaces that must be fulfilled.  It also includes `fx` usage for `aggregate/offer` return value

It follows spec https://github.com/web3-storage/specs/blob/feat/filecoin-spec/w3-aggregation.md

There are two interfaces defined in this service:
- `offerStore` - responsible for receiving offers to be queued. Implementer receives `commD` of aggregate together with Offer content. These will be used to give SPs. Note that implementer should keep track of the offer's `commD` so that it can query over time when a deal is done (or failures), so that the `fx` invocation can be executed and a receipt generated
- `aggregateStore` - store of aggregates already places with SPs (implementor like `spade-proxy` will have this store hooked with spade DB)

It is worth noting that this is a first iteration and not the complete thing at this moment. More specifically, in follow up iterations we need to:
- validate `commitmentProof`  https://github.com/web3-storage/w3up/issues/805
- validate cbor decoded offer with schema https://github.com/web3-storage/w3up/issues/806
- `aggregateGet` types are still unknown at this moment https://github.com/web3-storage/w3up/issues/807

New detailed issues for above will be created before this PR gets merged.

Part of  https://github.com/web3-storage/w3filecoin/issues/19

Still to do:
- [x] change commp from text to CID

Usage:
- `aggregate-api` - https://github.com/web3-storage/spade-proxy/pull/2
- `aggregate-client` - https://github.com/web3-storage/w3filecoin/pull/22

Closes #772 and #773


packages/aggregate-api/src/aggregate/offer.js
```diff
@@ -0,0 +1,91 @@
...
+          ),
+        }
+      }
+
+      // Validate offer content
+      const size = offers.reduce((accum, offer) => accum + offer.size, 0)
+      if (size < MIN_SIZE) {
+        return {
+          error: new Server.Failure(
+            `provided size is not enough to create an offer (${size} < ${MIN_SIZE})`
```
alanshaw, line 36:
```suggestion
            `offer under size, offered: ${size}, minimum: ${MIN_SIZE}`
```

packages/aggregate-api/src/aggregate/offer.js
```diff
@@ -0,0 +1,91 @@
...
+          ),
+        }
+      } else if (size > MAX_SIZE) {
+        return {
+          error: new Server.Failure(
+            `provided size is larger than it can be accepted for an offer (${size} > ${MAX_SIZE})`
```
alanshaw, line 42:
```suggestion
            `offer over size, offered: ${size}, maximum: ${MAX_SIZE}`
```

packages/aggregate-api/src/aggregate/offer.js
```diff
@@ -0,0 +1,91 @@
...
+          ),
+        }
+      } else if (size !== capability.nb.size) {
+        return {
+          error: new Server.Failure(
+            `provided size ${capability.nb.size} does not match computed size ${size}`
```
alanshaw, line 48:
```suggestion
            `offer size mismatch, specified: ${capability.nb.size}, actual: ${size}`
```

packages/aggregate-api/src/offer/arrange.js
```diff
@@ -0,0 +1,29 @@
...
+export function offerArrangeProvider({ arrangedOfferStore }) {
+  return Server.provide(Offer.arrange, async ({ capability }) => {
+    const commitmentProof = capability.nb.commitmentProof
+
+    const status = await arrangedOfferStore.get(commitmentProof)
+
+    if (!status) {
+      return {
+        error: new Server.Failure(
+          `provided commitment proof ${commitmentProof} has no arranged offers`
```
alanshaw, line 18:
```suggestion
          `arranged offer not found for commitment proof: ${commitmentProof}`
```

packages/aggregate-client/src/aggregate.js
```diff
@@ -0,0 +1,97 @@
...
+  const size = offers.reduce((accum, offer) => accum + offer.size, 0)
+  if (size < MIN_SIZE) {
+    throw new Error(
+      `provided size is not enough to create an offer (${size} < ${MIN_SIZE})`
+    )
+  } else if (size > MAX_SIZE) {
+    throw new Error(
+      `provided size is larger than it can be accepted for an offer (${size} > ${MAX_SIZE})`
+    )
+  }
```
alanshaw, line 37: I'd remove this and let the server validate - different implementations might allow different min/max sizes and want to use the same client.

packages/aggregate-client/src/aggregate.js
```diff
@@ -0,0 +1,97 @@
...
+    for (const u of offer.src) {
+      try {
+        new URL(u)
+      } catch {
+        throw new Error(
+          `provided url ${u} for offer CAR ${offer.link.toString()} is invalid`
+        )
+      }
+    }
+  }
```
alanshaw, line 50: Move to server side.

alanshaw (changes requested):

Gozala (approved):
Made some suggestions around error handling. Feel free to address or disregard, I don't think they are blockers.

Gozala:
> While writing this and getting to know more about commP, commD and so on, thinking perhaps we should do some renames from what we decided in spec:

Same here.

> aggregate/offer nb to rename commitmentProof to commD or dataSegementProof ?
CBOR block offer data structure to rename commitmentProof to pieceLink and size to pieceSize ?

I end up adding `piece: { size: Int, link: Link }` to the `store/add` capability.

https://github.com/web3-storage/w3up/commit/6314d266bc613090b0ccfabc4d92adedd3b5915b#diff-32283b314c355aa3cc4bc4945fbde9dbd548b1e35fed4f8e064fc7d9baea9872

I think it makes more sense to adopt `Piece` term used by the [Filecoin spec](https://spec.filecoin.io/systems/filecoin_files/piece/)

In other words what we call `OfferDetails` should really be called `ContentPiece` or something like that and contain ContentArchive (CAR) info and piece info. Something like

```ts
interface ContentPiece {
  link: Link
  src: URL
  piece: {
     link: Link
     size: Int
  }
}
```

> do we also need to store paddedSize ?

I don't think so

> aggregate/get nb to rename commitmentProof to commD or dataSegementProof ?

Likewise I think we should adopt `Segment` term similar to `Piece`. _(commD is a CID of of the Segment unless I'm mistaken, I'll know for sure once I'll get around to implementing this)_.

We may also want to double check the rems with @ribasushi on the next call.

packages/aggregate-api/src/aggregate/offer.js
```diff
@@ -0,0 +1,91 @@
...
+      }).join(fx.link())
+    },
+  })
+}
+
+/**
+ * @param {Server.API.Link<unknown, number, number, 0 | 1>} offerCid
+ * @param {Server.API.Invocation<Server.API.Capability<"aggregate/offer", Server.API.URI<"did:">, Pick<{ offer: Server.API.Link<unknown, number, number, 0 | 1>; commitmentProof: string; size: number & Server.API.Phantom<{ typeof: "integer"; }>; }, "offer" | "commitmentProof" | "size">>>} invocation
+ */
+function getOfferBlock(offerCid, invocation) {
```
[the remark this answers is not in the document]
vasco-santos: ` Otherwise capability handler will simply error and user will not have good feedback on what has happened.`

If no `offer` block present, it will fail with error block not found.

However, I totally agree with validating the schema! We will still change namings with to match `piece` and so on. So, I will create an issue to handle this as follow up

packages/aggregate-api/src/types.ts
```diff
@@ -0,0 +1,77 @@
...
+export interface OfferServiceContext {
+  arrangedOfferStore: ArrangedOfferStore
+}
+
+export interface ServiceContext
+  extends AggregateServiceContext,
+    OfferServiceContext {}
+
+export interface ArrangedOfferStore {
+  get: (commitmentProof: string) => Promise<string | undefined>
```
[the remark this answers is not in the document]
vasco-santos: We should look into this as a major change of APIs. We have all other (access-api, upload-api, ...) as we have here. I would prefer for now to be uniform across the board instead of needing to have different mental models

packages/aggregate-client/src/types.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export interface OfferArrangeResponse {
+  status: string
+}
+
+export interface Offer {
+  link: CARLink
+  size: number
+  commitmentProof: string // TODO: ProofLink
```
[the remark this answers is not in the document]
vasco-santos: I will add this once we move to Piece

packages/aggregate-api/src/aggregate/get.js
```diff
@@ -0,0 +1,27 @@
...
+    if (!aggregateArrangedResult) {
+      return {
+        error: new Server.Failure(
+          `requested aggregate with commitment proof ${commitmentProof} is not known`
+        ),
+      }
+    }
+    return {
+      ok: {
+        deals: aggregateArrangedResult,
```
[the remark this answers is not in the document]
vasco-santos: @alanshaw an aggregate is identified by unique commD. Therefore, on an `aggregateGet` we will only have one aggregate. This PR for now minimally follows the spec on this https://github.com/web3-storage/specs/blob/main/w3-aggregation.md#aggregateget

Anyway, at this point is still unclear the types for these (as mentioned in first nessafe of PR this should be handled as follow up once we integrate with Spade Oracle - #807 ). I am happy to work with you to flush how this should be when we get to that

packages/aggregate-api/src/aggregate/offer.js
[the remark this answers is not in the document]
vasco-santos: https://github.com/web3-storage/w3up/issues/806

vasco-santos:
A few follow ups will be worked on:

- validate `commitmentProof`  https://github.com/web3-storage/w3up/issues/805
- validate cbor decoded offer with schema https://github.com/web3-storage/w3up/issues/806
- `aggregateGet` types are still unknown at this moment https://github.com/web3-storage/w3up/issues/807
- integrate spec changes from https://github.com/web3-storage/specs/pull/62

packages/aggregate-api/src/aggregate/get.js
```diff
@@ -0,0 +1,38 @@
...
+  return {
+    ok: {
+      deals: aggregateArrangedResult,
+    },
+  }
+}
+
+class AggregateNotFound extends Server.Failure {
+  get name() {
+    return /** @type {const} */ ('AggregateGetFailure')
```
alanshaw, line 36:
```suggestion
    return /** @type {const} */ ('AggregateNotFoundFailure')
```

packages/aggregate-api/src/aggregate/offer.js
```diff
@@ -0,0 +1,133 @@
...
+      // TODO: Validate with schema
+    }
+  }
+}
+
+class AggregateOfferInvalidUrlError extends Server.Failure {
+  get name() {
+    return /** @type {const} */ ('AggregateOfferInvalidUrlFailure')
+  }
+}
```
alanshaw, line 121: As far as I can tell most of the time we don't suffix with `Error` or `Failure`, e.g.

https://github.com/web3-storage/w3up/blob/57182b1d193924a6c5eefa3e01ef56719303c42e/packages/capabilities/src/types.ts#L47-L49
https://github.com/web3-storage/w3up/blob/57182b1d193924a6c5eefa3e01ef56719303c42e/packages/upload-api/src/types.ts#L236C14-L238

Perhaps be consistent with this?

alanshaw (approved):
LGTM, just a minor non blocking consistency thing on the failures...

packages/aggregate-api/src/aggregate/offer.js
vasco-santos: good catch!

PR #790, opened 2023-05-11
feat!: merge `@web3-storage/access-api` into `@web3-storage/upload-api`

travis:
image

As part of the transition from D1 to Dynamo for delegation and provision storage, move service providers from `@web3-storage/access-api` to `@web3-storage/upload-api`. 

This change decouples this entire repository from any particular storage implementation and should make it easier for third parties to create w3up-compatible storage services. We now expose 2 storage interfaces - `DelegationsStorage` and `ProvisionsStorage` - and an email provider interface (right now simply called `Email`) that implementors like our very own `w3infra` need to implement. A prototype implementation of these interfaces can be found in https://github.com/web3-storage/w3infra/pull/194 - feedback on that PR was the basis for some of the public interface changes in this PR.

I've also taken this opportunity to drop support for the voucher protocol and stop using the `spaces` table which appears to only support that. I've also dropped the `accounts` table since it didn't seem to be adding anything - the one query we made to it returned the account DID that we used to query the table. Once we have a clearer picture of what information we want to store in an account we can bring that back if that makes sense.

Important notes for reviewers:
* There are definitely some awkward interfaces hanging around after merging these two codebases - please feel free to call them out, and I'll either fix them now or file some fast follow issues for more time-intensive refactors.
* I've deleted `access-api` primarily because it makes the PR much more readable. I'm open to removing it with this PR, but also open to keeping it in - if we decide to keep it for now, I'll restore it just before merging.
* @gozala has identified a number of improvements to the interfaces and underlying design - some of them I hope to incorporate into this round of work and may result in some (hopefully small) changes to this PR, some of which I think make sense for future work
* we should rename upload-api. should we rename upload-api? probably? now or later? I don't have super strong feelings, but we should probably do this soonish?

### TODO

- [ ] finish https://github.com/web3-storage/w3infra/pull/194
- [x] incorporate @gozala's feedback about interfaces from https://github.com/web3-storage/w3infra/pull/194
- [x] move Email implementation to w3infra
- [x] port relevant tests from `access-api` to `upload-api`


packages/upload-api/src/access-claim.js
```diff
@@ -0,0 +1,31 @@
+import * as Server from '@ucanto/server'
```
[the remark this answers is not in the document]
travis: oh yes I think that's a great idea - will do that now!

packages/upload-api/src/provider-add.js
```diff
@@ -1,25 +1,21 @@
...
  * @param {API.Input<Provider.add>} input
- * @param {Context} context
+ * @param {API.ProviderServiceContext} context
  */
 export const add = async (
   { capability, invocation },
-  { models: { provisions } }
+  { provisionsStorage: provisions }
 ) => {
   const {
```
Gozala, line 20: Does not need to be part of this PR, but I would really like if this instead invoked `customer.add` from the service DID, and then invoked `consumer.add` also from the service DID.

That is conceptually what supposed to be happening and provider/add is simply a sugar to save roundtrip.

packages/upload-api/src/space-allocate.js
```diff
@@ -1,29 +1,19 @@
...
- * @param {Context} context
+ * @param {API.SpaceServiceContext} context
  * @returns {Promise<API.Result<{size:number}, API.AllocationError>>}
  */
 export const allocate = async ({ capability }, context) => {
   const { with: space, nb } = capability
   const { size } = nb
-  const { provisions, spaces } = context.models
-  const result = await provisions.hasStorageProvider(space)
+  const result = await context.provisionsStorage.hasStorageProvider(space)
```
Gozala, line 14: Again not necessarily something this PR should solve, but we should instead be checking if space has provision for the `provider: invocation.audience.did()`. Or alternatively it could be `consumer/has` invocation from the service DID.

packages/upload-api/src/consumer/has.js
```diff
@@ -22,5 +22,5 @@ const has = async ({ capability }, context) => {
     )
   }
 
-  return context.models.provisions.hasStorageProvider(capability.nb.consumer)
+  return context.provisionsStorage.hasStorageProvider(capability.nb.consumer)
```
Gozala, line 25: We should checking whether `provider` matching `capability.with` is present as opposed to arbitrary one. Because one might have `nft.storage` provider but that does not allow them to store in `web3.storage` and vice versa.

I realize this is not something changes here alter, but given the migration it might be a good time to make such a change. Not a blocker.

packages/upload-api/src/space.js
```diff
@@ -0,0 +1,48 @@
...
+    const unexpectedSpaceDidFailure = {
+      name: 'SpaceUnknown',
+      message: `can only get info for did:key spaces`,
+    }
+    return {
+      error: unexpectedSpaceDidFailure,
+    }
+  }
+
+  const result = await provisions.hasStorageProvider(spaceDid)
```
Gozala, line 26: As in all other cases we should consider provider in this case it would be `invocation.audience`

packages/upload-api/src/types.ts
```diff
@@ -36,6 +95,53 @@ export interface Service {
...
+    error: ServiceMethod<
+      InferInvokedCapability<typeof Capabilities.Console.error>,
+      never,
+      Failure & { cause: unknown }
+    >
+  }
+  access: {
+    authorize: ServiceMethod<AccessAuthorize, AccessAuthorizeSuccess, Failure>
+    claim: ServiceMethod<AccessClaim, AccessClaimSuccess, AccessClaimFailure>
+    // eslint-disable-next-line @typescript-eslint/ban-types
```
Gozala, line 113: I think this might be obsolete now.

packages/upload-api/test/access-client-agent.js
```diff
@@ -389,24 +397,27 @@ export const test = {
       agent,
       agent.issuer.did()
     )
-
+    // wait 100 milliseconds to let the email be sent
```
Gozala, line 400: `mail.take()` should take care of waiting until mail is there, so this seems obsolete

packages/upload-api/test/access-client-agent.js
```diff
@@ -347,17 +351,21 @@ export const test = {
...
-      email: 'example@dag.house',
+      accountEmail: 'example@dag.house',
     })
 
-    const authorized = authorizeAndWait(agent, email, {
+    const authorized = authorizeAndWait(agent, accountEmail, {
       expectAuthorization: waitForAuthorizationByPolling,
     })
+    // wait 100 milliseconds to let the email be sent
+    await new Promise((resolve) => {
```
Gozala, line 366: `mail.take()` should take care of waiting until mail is there, so this seems obsolete

packages/upload-api/test/delegations-storage-tests.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {object} [opts]
+ * @param {Ucanto.Signer<Ucanto.DID>} [opts.issuer]
+ * @param {Ucanto.Principal} [opts.audience]
+ * @param {Ucanto.Capabilities} [opts.capabilities]
+ * @returns {Promise<Ucanto.Delegation>}
+ */
+async function createDelegation(opts = {}) {
+  const {
+    issuer = await principal.ed25519.generate(),
+    audience = issuer,
```
Gozala, line 18: `issuer` and `audience` are always passed so this seems like unnecessary generalization.

packages/upload-api/test/provider-add.spec.js
```diff
@@ -0,0 +1,447 @@
...
+    const { service } = context
+    try {
+      const proofs = await createAuthorization({ agent, service, account })
+      const addNFTStorage = await Provider.add
+        .invoke({
+          issuer: agent,
+          audience: service,
+          with: account.did(),
+          nb: {
+            provider: 'did:web:nft.storage',
```
Gozala, line 198: Aside: I think we should drop `provider` field in favor of `audience`.

Gozala (approved):

Gozala (changes requested):
Looks good! Only thing I would like to be addressed is the fact that we simply return customer DID from the `customer/has` without any checks. Once that is addressed this is ready to go. No need for another review.

Gozala (changes requested):
Looks good! Only thing I would like to be addressed is the fact that we simply return customer DID from the `customer/has` without any checks. Once that is addressed this is ready to go. No need for another review.

packages/upload-api/src/access/authorize.js
```diff
@@ -50,8 +50,6 @@ export const authorize = async ({ capability }, ctx) => {
     })
     .delegate()
 
-  await ctx.models.accounts.create(capability.nb.iss)
-
```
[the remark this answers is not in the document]
travis: Yea agree - so far I think we can make do with the `ProvisionsStorage` and `DelegationsStorage` interfaces but I think we should revisit that part of the design after I get this landed - we may want some of `CustomersStorage`, `ConsumersStorage`, `AccountsStorage`, etc and may want to get rid of `ProvisionsStorage`. Definitely want to push that out post-merge tho.

packages/upload-api/src/access/claim.js
```diff
@@ -1,29 +1,31 @@
...
   const claimedAudience = invocation.capabilities[0].with
-  const claimed = await collect(delegations.find({ audience: claimedAudience }))
+  const claimedResult = await delegations.find({ audience: claimedAudience })
+  if (claimedResult.error){
+    return claimedResult
+  }
   return {
     ok: {
-      delegations: delegationsResponse.encode(claimed),
+      delegations: delegationsResponse.encode(claimedResult.ok),
```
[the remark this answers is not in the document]
travis: that's an interesting idea. definitely out of scope for this PR because of the client updates it would require, but I'm open to this - would probably save us a bit of money on bandwidth and compute.

packages/upload-api/src/customer/get.js
```diff
@@ -24,7 +24,11 @@ const get = async ({ capability }, context) => {
     return { error: new UnknownProvider(capability.with) }
   }
 
-  const customer = await context.models.accounts.get(capability.nb.customer)
+  // TODO: we used to query the accounts table for this, but that only
+  // returns the DID we already have. we may want to query other tables in the future
+  // but for now I think we can skip this and drop the accounts table entirely
+  // const customer = await context.models.accounts.get(capability.nb.customer)
```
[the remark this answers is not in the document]
travis: got it! I've updated this PR along these lines - I introduced a new `getCustomer` function on `ProvisionsStorage` - since users essentially become "customers" when we provision storage for them this seems reasonable to me, but per my comment above we may want to introduce new interfaces for this soon...

packages/upload-api/src/provider-add.js
travis: yep agreed. also think we shouldn't get this in in this PR, but I'd like to move in this direction as well.

packages/upload-api/src/space-allocate.js
travis: yep agreed - I'd like to keep this out of this PR, but agree that we should move in this direction

packages/upload-api/src/space.js
travis: ok I have added an issue to the repo to track this and the others - would be great to get this in soon https://github.com/web3-storage/w3up/issues/802

packages/upload-api/src/types.ts
travis: I think so yep! will remove

packages/upload-api/test/access-client-agent.js
travis: ah so - I think this worked in the `access-api` testing infrastructure, but I had to hack around it to get this working in `upload-api` and didn't implement that particular bit - I think it would be straightforward to get it back in, so let me give that a shot...

travis: fixed!

packages/upload-api/test/delegations-storage-tests.js
travis: hm I'm not sure what you mean by this, could you expand on this?

travis:
woo hoo! updated per all of these comments.

> Only thing I would like to be addressed is the fact that we simply return customer DID from the customer/has without any checks.

I've done this in https://github.com/web3-storage/w3up/pull/790/commits/4287e8507ad3d0450b16be5bca83ee0c41ea729a, but it's induced a bit of a change that broke tests in a way that I fixed in https://github.com/web3-storage/w3up/pull/790/commits/35cff74d132dcfad0aaea70d71a87b8466e9afea

The idea behind the change is that we shouldn't consider a user a "customer" until we've provisioned storage for them. The test I updated in https://github.com/web3-storage/w3up/pull/790/commits/35cff74d132dcfad0aaea70d71a87b8466e9afea would verify that after a user did an `access/authorize` we'd get a customer record back when we invoked `customer/get` with their DID. 

This is tricky the way things are currently implemented because we don't actually save a "subscription" record until we allocate that customer storage, and that's what we currently use to track customers.

We could introduce a new "customers" table and add a record to it when we invoke `access/authorize` but that's actually not my preference - I'd like to avoid tracking information about customers unless we have a relationship with them that means we'll be storing non-delegation data for them, and I like how the current system operates in that regard.

@Gozala what do you think? Happy to jump on a quick call tomorrow to chat more if that'd be helpful!

travis:
Also @Gozala just want to make sure that https://github.com/web3-storage/w3up/commit/4287e8507ad3d0450b16be5bca83ee0c41ea729a addressed the issue you mentioned above with `customer/get` - if it does could you hit the approve button?

Gozala (approved):

PR #800, opened 2023-05-25
feat: enable reuse of stored invocations for delegations storage

travis:
we'd like to stop storing copies of delegations separately, and to do that we need to pass the CID of the invocation where the delegations do live so that we can grab it and pull delegations out of it when they are requested. This interface change enables this change, which is implemented in w3infra in https://github.com/web3-storage/w3infra/pull/201

### DIFF
--- a/packages/upload-api/src/access/confirm.js
+++ b/packages/upload-api/src/access/confirm.js
@@ -73,7 +73,7 @@ export async function confirm({ capability, invocation }, ctx) {
   // Store the delegations so that they can be pulled with access/claim
   // The fact that we're storing proofs chains that we pulled from the
   // database is not great, but it's a tradeoff we're making for now.
-  await ctx.delegationsStorage.putMany(delegation, attestation)
+  await ctx.delegationsStorage.putMany(invocation.asCID, delegation, attestation)
 
   return {
     ok: {
--- a/packages/upload-api/src/access/confirm.js
+++ b/packages/upload-api/src/access/confirm.js
@@ -73,7 +73,7 @@ export async function confirm({ capability, invocation }, ctx) {
   // Store the delegations so that they can be pulled with access/claim
   // The fact that we're storing proofs chains that we pulled from the
   // database is not great, but it's a tradeoff we're making for now.
-  await ctx.delegationsStorage.putMany(invocation.asCID, delegation, attestation)
+  await ctx.delegationsStorage.putMany(invocation.link(), [delegation, attestation])
 
   return {
     ok: {
--- a/packages/upload-api/src/types/delegations.ts
+++ b/packages/upload-api/src/types/delegations.ts
@@ -14,6 +14,7 @@ export interface DelegationsStorage<
    * @param delegations - delegations to store
    */
   putMany: (
+    cause: Ucanto.Link,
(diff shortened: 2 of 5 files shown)


vasco-santos (approved):
LGTM! Should wait on getting the other PR in before merging ?

packages/upload-api/src/access/confirm.js
```diff
@@ -73,7 +73,7 @@ export async function confirm({ capability, invocation }, ctx) {
   // Store the delegations so that they can be pulled with access/claim
   // The fact that we're storing proofs chains that we pulled from the
   // database is not great, but it's a tradeoff we're making for now.
-  await ctx.delegationsStorage.putMany(delegation, attestation)
+  await ctx.delegationsStorage.putMany(invocation.asCID, delegation, attestation)
```
Gozala, line 76:
```suggestion
  await ctx.delegationsStorage.putMany(invocation.link(), delegation, attestation)
```

I would avoid `asCID`, which I should probably mark as deprecated in the ucanto.

packages/upload-api/src/types/delegations.ts
```diff
@@ -14,6 +14,7 @@ export interface DelegationsStorage<
    * @param delegations - delegations to store
    */
   putMany: (
+    cause: Ucanto.Link,
     ...delegations: Array<Ucanto.Delegation<Ucanto.Tuple<Cap>>>
```
Gozala, line 18: We should just pass this as an array and probably of delegation CIDs as opposed to delegations.

> JS engines actually set limits on number of arguments function can have and I have experienced nasty bugs caused by going over that limit.

packages/upload-api/test/delegations-storage-tests.js
```diff
@@ -46,7 +46,7 @@ export function testVariant(createVariant, test) {
     const delegations = await Promise.all(
       Array.from({ length: count }).map(() => createSampleDelegation())
     )
-    await delegationsStorage.putMany(...delegations)
+    await delegationsStorage.putMany(delegations[0].asCID, ...delegations)
```
Gozala, line 49:
```suggestion
    await delegationsStorage.putMany(delegations[0].link(), ...delegations)
```

packages/upload-api/test/delegations-storage-tests.js
```diff
@@ -76,7 +76,7 @@ export function testVariant(createVariant, test) {
       )
     )
 
-    await delegations.putMany(...delegationsForAlice, ...delegationsForBob)
+    await delegations.putMany(delegationsForAlice[0].asCID, ...delegationsForAlice, ...delegationsForBob)
```
Gozala, line 79:
```suggestion
    await delegations.putMany(delegationsForAlice[0].link(), ...delegationsForAlice, ...delegationsForBob)
```

Gozala (approved):

packages/upload-api/src/types/delegations.ts
Gozala: I'm realizing that we need more details here than delegation CID, however it still probably better to pass records like `{ cause: CID, delegation: CID, audience: DID,  nbf?, exp? }` as opposed to whole Delegation.

travis:
I'm going to merge this into the main PR - feel like I have enough support that any remaining changes can be reviewed as part of that

PR #803, opened 2023-06-05
feat!: integrate commP into upload-client

Gozala:
- Amends `store/add` capability with `piece` field
  - ⚠️ End up replacing `chai` asserts instead of regular `assert` as later did not have `assert.match` in browser
- Integrates `data-segment` lib and uses it to compute `piece` info for uploads.
  - ⚠️ This is breaking change for `store/add` as we now return more details than just CAR CID. We could avoid this but than client will have to way of knowing what the `piece` info ended up being which I think would be more trouble than breaking change.

Note that tests do not actually check that commP generation was correct, because server performs no such verification and commP generation tests live in the data-segment library.

### DIFF
--- a/packages/upload-client/test/index.test.js
+++ b/packages/upload-client/test/index.test.js
@@ -109,7 +109,11 @@ describe('uploadFile', () => {
     assert.equal(dataCID.toString(), expectedCar.roots[0].toString())
   })
 
-  it('allows custom shard size to be set', async () => {
+  it('allows custom shard size to be set', async function () {
+    // on my device it usually takes 4secs but on CI it seems to exceed 5s
+    // timeout. Bump it up to 10s.
+    this.timeout(10_000)
+
     const space = await Signer.generate()
     const agent = await Signer.generate() // The "user" that will ask the service to accept the upload
     const file = new Blob([await randomBytes(1024 * 1024 * 5)])


packages/upload-client/test/index.test.js
```diff
@@ -109,7 +109,11 @@ describe('uploadFile', () => {
     assert.equal(dataCID.toString(), expectedCar.roots[0].toString())
   })
 
-  it('allows custom shard size to be set', async () => {
+  it('allows custom shard size to be set', async function () {
+    // on my device it usually takes 4secs but on CI it seems to exceed 5s
+    // timeout. Bump it up to 10s.
```
vasco-santos, line 114: looks like we could improve this yesterday with worker ?

vasco-santos (commented):
Code here looks great. The breaking API also definitely makes sense for me

Couple of things before getting it  in @Gozala 

- What are next steps? Do we plan do get this hooked up with `upload-api` in follow PR? I am assuming this information would land into dynamoDB `store` table
- Could we get the performance optimizations for browser by the reported slow downs there?

packages/capabilities/src/store.js
```diff
@@ -46,12 +51,16 @@ export const add = capability({
...
      */
-    link: Link,
+    link: Schema.link(),
     /**
      * Size of the CAR file to be stored. Service will provision write target
      * for this exact size. Attempt to write a larger CAR file will fail.
      */
     size: Schema.integer(),
+    /**
+     * The file coin piece details.
```
alanshaw, line 61:
```suggestion
     * The filecoin piece details.
```

packages/capabilities/src/store.js
```diff
@@ -64,16 +73,16 @@ export const add = capability({
...
-      return ok({})
-    }
+    return (
+      and(equalWith(claim, from)) ||
+      and(checkLink(claim.nb.link, from.nb.link, 'nb.link')) ||
+      and(checkQuota(claim.nb.size, from.nb.size, 'nb.size')) ||
+      and(
+        checkLink(claim.nb.piece.link, from.nb.piece.link, 'nb.piece.link')
+      ) ||
+      and(equal(claim.nb.piece.size, from.nb.piece.size, 'nb.piece.size')) ||
```
alanshaw, line 83: I realize this was not added here but `and` does not read well for me - I'm having real trouble understanding what this code means.

packages/capabilities/src/store.js
```diff
@@ -94,7 +103,13 @@ export const remove = capability({
      */
     link: Link,
   }),
-  derives: equalLink,
+  derives: (claimed, delegated) => {
+    return (
+      and(equalWith(claimed, delegated)) ||
+      checkLink(claimed.nb.link, delegated.nb.link, 'nb.link') ||
```
alanshaw, line 109: Why does this not have a call to `and()`?

alanshaw (approved):

vasco-santos:
should we pause this for now until MVP @alanshaw @Gozala ? We previously aligned (right?) that we will compute commP for first iteration

alanshaw:
IMHO we should get this into the client asap, even if we're not using it on the server side.

PR #812, opened 2023-06-07
fix: aggregation workflows

vasco-santos:
Previous PR went out without workflows commited https://github.com/web3-storage/w3up/pull/787 🤦🏼‍♂️


travis:
workflows hooray! it's a little strange that this PR is only running one of the workflows, any idea why that might be @vasco-santos ? This all looks good to me, but I'm going to hold off on approving because I've pushed a couple commits on to the end of it - the first is required to get the build working, the second could use a bit more scrutiny: 

https://github.com/web3-storage/w3up/pull/812/commits/0f2750a8badee5551534d487cb0c08db815ea109

vasco-santos:
@travis looks like only the last commit did not have the aggregate API running. Looking into it

vasco-santos:
It errored to start in last commit https://github.com/web3-storage/w3up/actions/runs/5205914542 due to workflow syntax

vasco-santos:
@travis it ran both workflows now :)

PR #814, opened 2023-06-09
fix: specify module types in exports

travis:
The existing config worked fine in some situations (including when using `file` dependencies rather than normal module version deps!), but for some reason the `w3infra` `upload-api` linter (and some tests) were unable to find the right type files and threw errors like:

```
node_modules/@web3-storage/capabilities/src/types.ts:4:57 - error TS7016: Could not find a declaration file for module './space.js'. '/Users/travis/dev/pl/w3infra/node_modules/@web3-storage/capabilities/src/space.js' implicitly has an 'any' type.

4 import { space, info, recover, recoverValidation } from './space.js'
                                                          ~~~~~~~~~~~~
```

This PR updates the exports following the pattern in `w3up-client` which seems to fix this issue - tested locally using `yalc` (which simulates package publishing).


packages/capabilities/package.json
```diff
@@ -23,44 +23,29 @@
...
+        "*",
+        "dist/*",
+        "dist/src/*",
+        "dist/src/*/index"
       ],
-      "types": [
-        "dist/src/types"
+      "src/*": [
+        "*",
+        "dist/*",
```
Gozala, line 46: Doesn't this cover bottom two ?

Gozala (approved):

packages/capabilities/package.json
travis: My understanding is that for a file at `src/foo/bar/baz` this line would ensure tsc looks in `dist/foo/bar/baz` while the lines below would ensure it looked in `dist/src/foo/bar/baz` and `dist/src/foo/bar/baz/index` so I do think they're differentiated. 

That said I think some of these are unnecessary and I'd like to slim them down soon - I cargo culted this from the `w3up-client` `package.json` and want to ship as-is since it's working, but will take another look at this post-merge...

PR #824, opened 2023-06-29
fix!: update aggregate spec in client and api

vasco-santos:
Updates client and api based on https://github.com/web3-storage/specs/commit/63846e529a77d200af4a4a8bdab4c4ac7320c107

It also starts relying on the https://github.com/web3-storage/data-segment/ lib for the tests by generating commP of randomly generated CARs. In next iteration, tests should also include commP of commPs builder (needs https://github.com/web3-storage/data-segment/pull/8)

Client API for submitting offer was changed to receive piece instead of computing it. Main reason for this is that there is a need to build it outside while keeping track of sizes. It could also happen here and explode if not, even though this is an aggregate offer call and not really the builder. So I would love feedback on API surface preferences. This will not be used by end users as will be abstracted within our services. The API will still need to validate if given commP of commPs is good for the given offer.

BREAKING CHANGE: aggregate capabilities now have different nb properties and aggregate client api was simplified

### DIFF
--- a/packages/aggregate-api/src/aggregate/offer.js
+++ b/packages/aggregate-api/src/aggregate/offer.js
@@ -27,7 +27,7 @@ export const claim = async (
 ) => {
   // Get offer block
   const offerCid = capability.nb.offer
-  const commitmentProof = capability.nb.commitmentProof
+  const piece = capability.nb.piece
   const offers = getOfferBlock(offerCid, invocation)
 
   if (!offers) {
@@ -52,30 +52,15 @@ export const claim = async (
         `offer over size, offered: ${size}, maximum: ${MAX_SIZE}`
       ),
     }
-  } else if (size !== capability.nb.size) {
+  } else if (size !== piece.size) {
     return {
       error: new AggregateOfferInvalidSizeError(
-        `offer size mismatch, specified: ${capability.nb.size}, actual: ${size}`
+        `offer size mismatch, specified: ${piece.size}, actual: ${size}`
       ),
     }
   }
 
-  // Validate URLs in offers src
-  for (const offer of offers.values()) {
-    for (const u of offer.src) {
-      try {
-        new URL(u)
-      } catch {
-        return {
-          error: new AggregateOfferInvalidUrlError(
-            `offer has invalid URL: ${u}`
-          ),
-        }
-      }
-    }
-  }
-
-  // TODO: Validate commP
+  // TODO: Validate commP of commPs
 
   // Create effect for receipt
   const fx = await Offer.arrange
@@ -84,13 +69,13 @@ export const claim = async (
       audience: context.id,
       with: context.id.did(),
       nb: {
-        commitmentProof,
+        pieceLink: piece.link,
       },
     })
     .delegate()
 
   // Write offer to store
-  await offerStore.queue({ commitmentProof, offers })
+  await offerStore.queue({ piece, offers })
 
   return Server.ok({
     status: 'queued',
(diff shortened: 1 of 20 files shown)


packages/aggregate-api/src/aggregate/offer.js
```diff
@@ -99,13 +84,13 @@ export const claim = async (
...
+ * @param {Server.API.Invocation<Server.API.Capability<"aggregate/offer", `did:${string}:${string}` & `did:${string}` & Server.API.Phantom<{ protocol: "did:"; }> & `${string}:${string}` & Server.API.Phantom<{ protocol: `${string}:`; }>, Pick<{ offer: Server.API.Link<unknown, number, number, 0 | 1>; piece: Server.Schema.InferStruct<{ link: Server.Schema.Schema<Server.API.Link<unknown, number, number, 0 | 1>, any>; size: Server.Schema.NumberSchema<number & Server.API.Phantom<{ typeof: "integer"; }>, unknown>; }>; }, "offer" | "piece"> & Partial<Pick<{ offer: Server.API.Link<unknown, number, number, 0 | 1>; piece: Server.Schema.InferStruct<{ link: Server.Schema.Schema<Server.API.Link<unknown, number, number, 0 | 1>, any>; size: Server.Schema.NumberSchema<number & Server.API.Phantom<{ typeof: "integer"; }>, unknown>; }>; }, never>>>>} invocation
```
alanshaw, line 87: Woah...maybe just pass the blocks iterator to the function instead?

packages/aggregate-api/src/offer/arrange.js
```diff
@@ -14,14 +14,14 @@ export const provide = (context) =>
  * @returns {Promise<API.UcantoInterface.Result<API.OfferArrangeSuccess, API.OfferArrangeFailure>>}
  */
 export const claim = async ({ capability }, { arrangedOfferStore }) => {
-  const commitmentProof = capability.nb.commitmentProof
+  const pieceLink = capability.nb.pieceLink
```
alanshaw, line 17: Can we not suffix with types? `piece: Link` is ok? Why is the size not being passed in this instance?

packages/aggregate-client/src/types.ts
```diff
@@ -77,3 +71,18 @@ export interface Connectable {
 export type CARLink = Link<unknown, typeof CAR.codec.code>
 
 export type OfferSrc = ToString<URL>
+
+/**
+ * [Piece CID](https://spec.filecoin.io/systems/filecoin_files/piece/) of some
+ * content.
+ */
+export type PieceCID = ReturnType<typeof CommP.toCID>
```
alanshaw, line 79: Is there not an exported type for this? Or at least can we not define like `CARLink`?

Call this `PieceLink` to be consistent with `CARLink`?

packages/aggregate-api/src/offer/arrange.js
vasco-santos: The size is only needed on the aggregate offer for SP validation. Here is when there was already a resolution and we just want to create the receipt with final result. For searching we do not need the size (and I don't even know if by the state we get from the Oracle we would have access to it easily).

`piece` is Object with link and size everywhere else. That is the reasoning to name it this way here as we decided in https://github.com/web3-storage/specs/pull/62

packages/aggregate-client/src/types.ts
vasco-santos: I think we want to be consistent with filecoin namings. I stole this from @Gozala 's PR https://github.com/web3-storage/w3up/pull/803 but I actually wondered if `data-segment` lib could actually export the type

Gozala: Yeah it should absolutely export a type, I'll do that

packages/aggregate-client/src/aggregate.js
```diff
@@ -50,12 +44,12 @@ export async function aggregateOffer(
  * Get details of an aggregate.
  *
  * @param {import('./types').InvocationConfig} conf - Configuration
- * @param {import('@ucanto/interface').Link<unknown, number, number, 0 | 1>} commitmentProof
+ * @param {import('@ucanto/interface').Link<unknown, number, number, 0 | 1>} subject
```
Gozala, line 47:
```suggestion
 * @param {import('@ucanto/interface').UnknownLink} subject
```

There is an `UnknownLink` type that covers version CID 0 as well.

packages/aggregate-api/src/offer/arrange.js
Gozala: I got to say `pieceLink` also bothered me, I was not going to say anything but since @alanshaw did how about we rename `piece` object into something else ? I think few cases we considered `pieceInfo` ?

Naming is so hard 😭 and it's made harder by the fact that piece CID does not fully identify the piece without the size....

packages/aggregate-api/src/aggregate/offer.js
Gozala: This type is pretty insane, could we maybe have some types defined in .ts file and referenced here instead ? Also I think some of the stuff is redundunt there. And that thecond type parameter is probably just `Server.API.DID`.

Also after second look, I think passing blocks iterator is probably a way to go here indeed.

Gozala (approved):

packages/capabilities/src/aggregate.js
```diff
@@ -11,6 +11,15 @@
...
 
+/**
+ * @see https://github.com/multiformats/go-multihash/blob/dc3bd6897fcd17f6acd8d4d6ffd2cea3d4d3ebeb/multihash.go#L73
+ */
+const SHA2_256_TRUNC254_PADDED = 0x1012
+/**
+ * @see https://github.com/ipfs/go-cid/blob/829c826f6be23320846f4b7318aee4d17bf8e094/cid.go#L104
+ */
+const FilCommitmentUnsealed = 0xf101
+
```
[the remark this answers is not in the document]
vasco-santos: I will wait for that to make it to next release

PR #825, opened 2023-06-29
fix: run format for upload-api

vasco-santos:
Noticed while working on aggregate-api and client that format/lint is failing for `upload-api`. So ran format for it in different branch to put fix it in main without creating noise in other PR

### DIFF
--- a/packages/upload-api/test/helpers/context.js
+++ b/packages/upload-api/test/helpers/context.js
@@ -68,7 +68,9 @@ export const createContext = async (options = {}) => {
  * @param {Types.UcantoServerTestContext} context
  */
 export const cleanupContext = async (context) => {
-  /** @type {CarStoreBucket & {  deactivate: () => Promise<void> }}} */ ;(
-    await context.carStoreBucket
-  ).deactivate()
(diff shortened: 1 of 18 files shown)


packages/upload-api/test/helpers/context.js
```diff
@@ -68,5 +68,9 @@ export const createContext = async (options = {}) => {
  * @param {Types.UcantoServerTestContext} context
  */
 export const cleanupContext = async (context) => {
-  /** @type {CarStoreBucket & {  deactivate: () => Promise<void> }}} */(await context.carStoreBucket).deactivate()
```
vasco-santos, line 71: Following error was happening in build script.

```
Error: packages/upload-api build: test/helpers/context.js(73,5): error TS2339: Property 'deactivate' does not exist on type 'CarStoreBucket'.
packages/upload-api build: Failed
```

See https://github.com/web3-storage/w3up/actions/runs/5412245896/jobs/9836091169

Type was not correctly set and an ignore was also needed because it does not include all the properties expected

alanshaw (approved):

PR #831, opened 2023-07-17
fix!: update aggregation capabilitites to use height instead of size together with client and api

vasco-santos:
Renames `size` to `height` for `piece`/`aggregate` based on described in https://github.com/web3-storage/data-segment/pull/13

Note that:
- given `piece` and `aggregate` are both perfect binary trees, leaf count of can be derived as `2 ** height`, and the size of the tree can be derived by `leafCount * Node.Size` (32).

Needs https://github.com/web3-storage/specs/pull/66

### DIFF
--- a/packages/aggregate-api/src/aggregate/offer.js
+++ b/packages/aggregate-api/src/aggregate/offer.js
@@ -1,11 +1,13 @@
 import * as Server from '@ucanto/server'
 import { CBOR } from '@ucanto/core'
+import { Node, Aggregate as AggregateBuilder } from '@web3-storage/data-segment'
 import * as Aggregate from '@web3-storage/capabilities/aggregate'
 import * as Offer from '@web3-storage/capabilities/offer'
 import * as API from '../types.js'
 
 export const MIN_SIZE = 1 + 127 * (1 << 27)
-export const MAX_SIZE = 127 * (1 << 28)
+// export const MAX_SIZE = 127 * (1 << 28)
+export const MAX_SIZE = AggregateBuilder.DEFAULT_DEAL_SIZE
 
 /**
  * @param {API.AggregateServiceContext} context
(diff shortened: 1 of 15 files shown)


packages/aggregate-api/src/aggregate/offer.js
```diff
@@ -1,11 +1,13 @@
...
 import { CBOR } from '@ucanto/core'
+import { Node, Aggregate as AggregateBuilder } from '@web3-storage/data-segment'
 import * as Aggregate from '@web3-storage/capabilities/aggregate'
 import * as Offer from '@web3-storage/capabilities/offer'
 import * as API from '../types.js'
 
 export const MIN_SIZE = 1 + 127 * (1 << 27)
-export const MAX_SIZE = 127 * (1 << 28)
+// export const MAX_SIZE = 127 * (1 << 28)
+export const MAX_SIZE = AggregateBuilder.DEFAULT_DEAL_SIZE
```
vasco-santos, line 10: @Gozala interesting that these values are different than ones previously obtained with riba. Clues?

packages/aggregate-api/test/aggregate.js
```diff
@@ -174,31 +145,22 @@ export const test = {
...
-    const piece = {
-      link: parseLink(
-        'baga6ea4seaqm2u43527zehkqqcpyyopgsw2c4mapyy2vbqzqouqtzhxtacueeki'
-      ),
-      size,
-    }
+    // Generate Pieces for offer
+    const { pieces, aggregate } = await randomAggregate(100, 128)
+    const badHeight = 31
+    const size = 2n ** BigInt(badHeight) * BigInt(Node.Size)
```
Gozala, line 151: We already have [PaddedSize.toHeight](https://github.com/web3-storage/data-segment/blob/61ee74d4ebc5c8eec8b5d4dcea2b46e844fe1dc8/src/piece/padded-size.js#L55-L60) we probably should just add `PaddedSize.fromHeight` so we don't have to do this math all over the place.

packages/aggregate-client/src/types.ts
```diff
@@ -70,19 +69,12 @@ export interface Connectable {
...
+ * Filecoin piece proof that can be used to derivew `PieceInfo`.
  * [Piece](https://spec.filecoin.io/systems/filecoin_files/piece/) information
  * for this CAR file.
  */
 export interface Piece {
-  link: PieceCID
-  size: number
+  link: PieceLink
+  height: number
 }
```
Gozala, line 80: Looks like we need another type [along with `Piece` and `PieceInfo`](https://github.com/web3-storage/data-segment/blob/61ee74d4ebc5c8eec8b5d4dcea2b46e844fe1dc8/src/api.ts#L119-L165) that we can re-export here.

packages/capabilities/src/aggregate.js
```diff
@@ -44,17 +44,20 @@ export const offer = capability({
...
        * CID of the aggregate piece.
        */
-      link: Schema.link({
-        code: FilCommitmentUnsealed,
-        version: 1,
-        multihash: {
-          code: SHA2_256_TRUNC254_PADDED,
-        },
-      }),
+      link: /** @type {import('./types').PieceLinkSchema} */ (
```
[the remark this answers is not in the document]
Gozala: You can not because there is no way to verify it, so I was expecting that ppl would write manually type cast like `Schema<Link<T, ...>>` that way you you imply to know what you're doing.

I'm not sure what config param in ucanto would look like. That said I have in-progress PR that would allow us to infer types like `Schema.struct({ ... }).link({ codec, hasher, version })` which than would be able to infer the type from the schema.

Gozala (approved):
Looks good.  Let's just make sure that MIN_SIZE and MAX_SIZE are what we expect them to be. I also proposed few minor changes to avoid repeating functionality provided by data-segment lib.

packages/capabilities/src/aggregate.js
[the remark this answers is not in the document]
vasco-santos: ok, will keep as is for the time being

packages/aggregate-client/src/types.ts
vasco-santos: https://github.com/web3-storage/data-segment/pull/17

packages/aggregate-api/test/aggregate.js
vasco-santos: https://github.com/web3-storage/data-segment/pull/18

PR #832, opened 2023-07-18
feat!: introduce new administrative capabilities

travis:
As described in https://hackmd.io/@tvpl/H1whErCI3, introduce a set of capabilities that will: 
1) allow providers to rate limit (especially fully limit, ie block) users 
2) allow providers to get information about subscriptions and consumers

I've taken this opportunity to remove the `AccessVerifier` from the service and now simply use the real access service implementation for space provisioning. 

I'm not entirely happy with the verbosity of the tests for either capabilities or invocation providers, but I'd like to address that in a separate line of work - I'm taking notes!

### TODO
- [x] write tests for capabilities 
- [x] implement in-memory RateLimitsStorage
- [x] use it to write `upload-api` tests
- [ ] move capabilities spec from hackmd to https://github.com/web3-storage/w3up/blob/main/spec/capabilities.md 
- [x] start using rate limits by porting https://github.com/web3-storage/w3up/pull/801/commits/0eea3d1268893062264414b2bbbeb689ac25c6c1 and https://github.com/web3-storage/w3up/pull/801/commits/d230033c1569094738b2dcdd9c6acb5afa889c82 to this new set of capabilities


packages/upload-api/src/types/rate-limits.ts
```diff
@@ -0,0 +1,43 @@
+import * as Ucanto from '@ucanto/interface'
+
+export type RateLimitID = string
```
Gozala, line 3: Could you add a comment explaining what and how it is derived from. As an aside I am not a fan of IDs that you need to keep track of, any reason why remove can't take same `{subject, rate}` pair as key ? It appears a lot more appealing IMO.

packages/capabilities/src/rate-limit.js
```diff
@@ -0,0 +1,71 @@
...
+      and(equalWith(child, parent)) ||
+      and(equal(child.nb.subject, parent.nb.subject, 'subject')) ||
+      and(equal(child.nb.rate, parent.nb.rate, 'rate')) ||
+      ok({})
+    )
+  },
+})
+
+/**
+ * Capability can be invoked by an agent to remove rate limits from a subject.
```
Gozala, line 38:
```suggestion
 * Capability can be invoked by a provider are an authorized delegate to remove rate limits from a subject.
```

packages/capabilities/src/rate-limit.js
```diff
@@ -0,0 +1,71 @@
...
+ * @module
+ */
+import { capability, DID, struct, Schema, ok } from '@ucanto/validator'
+import { equalWith, and, equal } from './utils.js'
+
+// e.g. did:web:web3.storage or did:web:staging.web3.storage
+export const Provider = DID
+
+/**
+ * Capability can be invoked by an agent to add a rate limit to a subject.
```
Gozala, line 18:
```suggestion
 * Capability can be invoked by the provider or an authorized delegate to add a rate limit to a subject.
```

packages/capabilities/src/rate-limit.js
```diff
@@ -0,0 +1,71 @@
...
+    return (
+      and(equalWith(child, parent)) ||
+      and(equal(child.nb.id, parent.nb.id, 'id')) ||
+      ok({})
+    )
+  },
+})
+
+/**
+ * Capability can be invoked by an agent to list rate limits on a subject
```
Gozala, line 56:
```suggestion
 * Capability can be invoked by the provider or an authorized delegate to list rate limits on the given subject
```

packages/capabilities/test/capabilities/rate-limit.test.js
Gozala, line 1: Would be nice to have some tests that check that escalating delegated capabilities like changing rate or subject fails.

packages/upload-api/src/access/authorize.js
```diff
@@ -15,6 +17,23 @@ export const provide = (ctx) =>
...
 export const authorize = async ({ capability }, ctx) => {
+  const isBlocked = await areAnyBlocked(
+    ctx.rateLimitsStorage,
+    emailAndDomainFromMailtoDid(
+      /** @type {import('@web3-storage/did-mailto/dist/src/types').DidMailto} */ (
+        capability.nb.iss
+      )
+    )
+  )
+  if (isBlocked.error || isBlocked.ok) {
```
Gozala, line 28: If there is an error we seem to disregard it and return `AccountBlocked` error. This seems unexpected and some code comment would be helpful here.

packages/upload-api/src/customer/get.js
```diff
@@ -24,11 +24,11 @@ const get = async ({ capability }, context) => {
     return { error: new UnknownProvider(capability.with) }
   }
 
-  const customer = await context.provisionsStorage.getCustomer(
+  const result = await context.provisionsStorage.getCustomer(
     capability.with,
     capability.nb.customer
   )
-  return { ok: { customer: customer.ok ? { did: customer.ok.did } : null } }
+  return result.ok ? { ok: { did: result.ok.did } } : result
```
Gozala, line 31: Just flagging that this is a breaking change, and we need to make sure nothing expects this to never fail.

packages/upload-api/src/provider-add.js
```diff
@@ -29,6 +31,22 @@ export const add = async (
...
+      /** @type {import('@web3-storage/did-mailto/dist/src/types').DidMailto} */ (
+        accountDID
+      )
+    )
+  )
+  if (isBlocked.error || isBlocked.ok) {
+    return {
+      error: {
+        name: 'AccountBlocked',
+        message: `Account identified by {capability.nb.iss} is blocked`,
```
Gozala, line 46: I would expect to report accountDID instead since that is what is checked above. Also there `$` was missing in template literal.

```suggestion
        message: `Account identified by ${accountDID} is blocked`,
```

packages/upload-api/src/upload/add.js
```diff
@@ -2,26 +2,48 @@ import pRetry from 'p-retry'
...
-    const space = Server.DID.parse(capability.with)
+    const space = /** @type {import('@ucanto/interface').DIDKey} */ (
+      Server.DID.parse(capability.with).did()
+    )
     const issuer = invocation.issuer.did()
-    const allocated = await access.allocateSpace(invocation)
-
-    if (allocated.error) {
-      return allocated
+    const hasProviderResult = await hasProvider(
```
Gozala, line 22: Do we not want to check the block list here ?

I also would recommend command style API over defensive style API that is instead of having `hasProvider` you could have `useCapability` so that you don't have to have one branch for propagating an error and other branch for creating an error if answer to question was `false`.

That is also why we had `allocateSpace` as opposed to `hasSpace.

P.S. It is very same idea as with [Parse, don’t validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) _(a good read if you have not come across it before)_.

packages/upload-api/src/utils/did-mailto.js
```diff
@@ -0,0 +1,10 @@
+import * as DidMailto from '@web3-storage/did-mailto'
+
+/**
+ * @param {import("@web3-storage/did-mailto/dist/src/types").DidMailto} mailtoDid
+ */
+export function emailAndDomainFromMailtoDid(mailtoDid) {
```
Gozala, line 6: Nit: Feels like it should be two different functions `toEmail(did): Email` and `toDomain(email: Email): string`

packages/upload-api/src/utils/rate-limits.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param {import("../types").RateLimit} rateLimit
+ */
+const isBlock = (rateLimit) => rateLimit.rate === 0
+
+/**
+ *
+ * @param {import("../types").RateLimit[]} rateLimits
+ */
+const areAnyBlocks = (rateLimits) => rateLimits.some(isBlock)
```
Gozala, line 11: I would have called it `isBlocked`

packages/upload-api/src/utils/rate-limits.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Query rate limits storage and find out if any of the given subjects
+ * have rate set to 0
+ *
+ * @param {import("../types").RateLimitsStorage} storage
+ * @param {string[]} subjects
+ * @return {Promise<import("@ucanto/interface").Result<boolean, import("@ucanto/interface").Failure>>}
+ */
+export async function areAnyBlocked(storage, subjects) {
```
Gozala, line 21: I would instead have something like `moderateRateLimit` and make it either succeed er error as opposed to erroring or returning a boolean.

packages/upload-api/test/provisions-storage.js
```diff
@@ -71,7 +76,65 @@ export class ProvisionsStorage {
...
+          message: `could not find ${subscription}`,
+        },
+      }
+    }
+  }
+
+  /**
+   *
+   * @param {Types.ProviderDID} provider
+   * @param {*} consumer
```
Gozala, line 114: Is this a string or something ? `*` is same as `any` which is not ideal. You could also put `unknown` if you mean some unknown value which I think will work here.

Gozala (approved):
I have provided some feedback and suggestions, nothing critical but I think there was one small bug, the rest is kind of take it or leave it.

Either way I'm approving this as I don't think we need another review round to get this out.

packages/upload-api/src/types/rate-limits.ts
travis: So my reasoning here was that an implementor might want to allow the same `{subject, rate}` pair in multiple rate limits - more than one party may be in charge of adding rate limits in the future (eg, the fraud prevention department flagging and blocking an account because they detected phishing sites being uploaded and, separately, billing blocking an account for non-payment) and in that case I think they would likely not want the removal of one block to always imply the removal of another block.

packages/capabilities/test/capabilities/rate-limit.test.js
travis: oh great idea, done!

packages/upload-api/src/access/authorize.js
travis: yea I went back and forth on this one - I figured it's moderately better to not authorize a new agent if we get an unexpected error - they can always retry, but we can't easily invalidate a set of delegations once we sign off on them. I'll add a code comment for sure, and am definitely open to reconsidering this!

packages/upload-api/src/customer/get.js
travis: ah good call, I'll make this a breaking `feat`. I can't find anywhere that we actually use this capability at the moment, do you remember if we use this anywhere?

packages/upload-api/test/provisions-storage.js
travis: whoop - autogenerated, should be a string, good catch!

packages/upload-api/src/utils/did-mailto.js
travis: done!

packages/upload-api/src/utils/rate-limits.js
travis: hm, this is a function which, given a list of rate limits, return true if any of them is a "block" - ie, if
any of them have a rate of 0 - I'll add a comment to that effect, but I like this name better for that

packages/upload-api/src/utils/rate-limits.js
travis: ah yea I ran with this and changed it to a `ensureRateLimit` function that also accepts a limit - this is more generally useful too which is nice, great idea!

packages/upload-api/src/upload/add.js
travis: oo I like this, and I also like being able to re-use `allocate` the way we were - I've pushed a new commit that makes this happen, and drops the `size` input on `space/allocate` - we weren't using it for anything and the fact that we don't have a `size` in the `upload/add` provider was the main reason I moved away from it.

I suspect we may want to re-introduce it later, either as an optional parameter that triggers a check against a space's allocation or as a separate but similar capability, but this seems ok to me for now - let me know if you think it might be a problem!

https://github.com/web3-storage/w3up/pull/832/commits/22b7a69e9861960206456b9b9802727b44cc79ec

packages/capabilities/src/rate-limit.js
```diff
@@ -0,0 +1,71 @@
...
+      and(equalWith(child, parent)) ||
+      and(equal(child.nb.subject, parent.nb.subject, 'subject')) ||
+      and(equal(child.nb.rate, parent.nb.rate, 'rate')) ||
+      ok({})
+    )
+  },
+})
+
+/**
+ * Capability can be invoked by the provider are an authorized delegate to remove rate limits from a subject.
```
olizilla, line 38: `Capability can be invoked by the provider **or** an authorized delegate`

packages/capabilities/src/types.ts
```diff
@@ -67,6 +79,75 @@ export interface InvalidProvider extends Ucanto.Failure {
...
+  id: string
+}
+export type RateLimitAddFailure = Ucanto.Failure
+
+export type RateLimitRemove = InferInvokedCapability<
+  typeof RateLimitCaps.remove
+>
+export type RateLimitRemoveSuccess = Unit
+
+export interface RateLimitsNotFound extends Ucanto.Failure {
```
olizilla, line 136: nit: RateLimit**s** is plural here but singular in all other types

packages/upload-api/src/space-allocate.js
```diff
@@ -1,19 +1,33 @@
...
+  const { with: space } = capability
+  const rateLimitResult = await ensureRateLimitAbove(
+    context.rateLimitsStorage,
+    [space],
+    0
+  )
+  if (rateLimitResult.error) {
+    return {
+      error: {
+        name: 'InsufficientStorage',
```
olizilla, line 23: `InsufficientStorage` doens't feel like the right name here. It's more of an `AccountBlocked` kind of error.

PR #836, opened 2023-07-21
fix: bump package number to ensure w3up-client depends on >= access@15

travis:
I released these packages in the wrong order, so the new w3up-client still depends on v14, which points at access.web3.storage. tweak the version dep (in a way I'll revert later) to ensure we move past that.


travis:
merging without approval because this is a mechanical hack that I'll be reverting after I release the package

PR #838, opened 2023-07-21
chore: unpin @web3-storage/access dep in w3up-client

travis:


travis:
merging without review because I'm just reverting https://github.com/web3-storage/w3up/pull/836

PR #840, opened 2023-07-25
fix: support for Node.js 20

alanshaw:
Support Node.js 20 in the clients.

Adds a temporary fix for Node.js 20 as `blob.stream()` is currently broken: https://github.com/nodejs/node/issues/48916

Also adds matrix config to CI workflows so that tests are run in both node 18 & 20.


olizilla (approved):
Yeesh

PR #862, opened 2023-08-16
feat: add providers to space/info

travis:
I believe we currently don't have any way for an agent to figure out which providers a space is registered to - add a list of providers to `space/info` to enable this in the CLI and w3console


alanshaw (approved):
LGTM 💪

PR #864, opened 2023-08-16
fix: re-enable upload-api tests

travis:
mocha's globbing behavior is... quirky. tests currently broken, working to fix.


alanshaw (approved):
LGTM - FWIW it's bash that's doing the expansion here - hence the quotes to prevent it and let mocha do it.

travis:
chatted w/ @alanshaw out of band on the test changes I made after approval and I think we're aligned - want to get this merged to get tests running again on main but happy to update from any post-merge comments!

PR #869, opened 2023-08-23
feat: return ID from ProvisionsStorage `put`

travis:
We generate an ID when we create a new "provision" but we don't currently return it - add it to the return value of `put`.

This is a bit muddled because the system is currently organized around "provisions" which are really an amalgamation of "subscriptions" and "consumers" - the plan since the D1 migration has been to break ProvisionsStorage into SubscriptionsStorage and ConsumersStorage - in that world this will be two steps - users will obtain a subscription from a provider and then attach it to a consumer, and it will be much clearer that this is actually the "subscription" ID.

I'd like to tackle this refactoring soon, but for now this makes it easier to write tests and prepare for that future state.


Gozala (approved):
Looks good to me

PR #870, opened 2023-08-25
fix: update docs to bump version

travis:
something seems to have gone wrong with release-please and the latest release doesn't have the latest code. bump the version to fix this, and improve docs at the same time.


travis:
self merging as it's only a docs change, but please feel free to suggest edits and I'll get them in!

PR #874, opened 2023-08-28
chore: remove spec subdirectory

travis:
this is almost all either incorrect or misleading, and has been replaced by specs in https://github.com/web3-storage/specs


olizilla:
This made typedoc unhappy as we reference specs/docs.css in our typedoc config https://github.com/web3-storage/w3up/blob/a906488c8cc36cc981f8703fc156b619fd492e34/tsconfig.json#L50

PR #875, opened 2023-08-28
feat: make agent Service generic

travis:
Right now we hardcode the Service type of an Agent to an old version of the access service types. This makes it possible to specify an arbitrary Service in the types, which should result in better typing behavior in clients.


Gozala (approved):
Looks good to me

PR #908, opened 2023-09-06
feat: adjust shard size

alanshaw:
As discussed in #764 and calculated in https://observablehq.com/@gozala/w3up-shard-size, this PR adjusts the `upload-client` to shard on a CAR size of <= `133,169,152` bytes.

It also fixes a bug where a CAR shard could be created with size greater than the shard size.

supersedes #764


packages/upload-client/src/unixfs.js
```diff
@@ -34,8 +34,12 @@ export function createFileEncoderStream(blob) {
   const unixfsWriter = UnixFS.createWriter({ writable, settings })
   const fileBuilder = new UnixFSFileBuilder('', blob)
   void (async () => {
-    await fileBuilder.finalize(unixfsWriter)
-    await unixfsWriter.close()
+    try {
+      await fileBuilder.finalize(unixfsWriter)
+      await unixfsWriter.close()
```
alanshaw, line 39: Calling close on a writable stream that has errored causes this to throw an uncaught exception.

packages/upload-client/src/sharding.js
```diff
@@ -22,21 +23,30 @@ export class ShardingStream extends TransformStream {
...
     super({
       async transform(block, controller) {
         if (readyShard != null) {
           controller.enqueue(await encode(readyShard))
           readyShard = null
         }
-        if (shard.length && size + block.bytes.length > shardSize) {
+
+        const blockLength = blockEncodingLength(block)
+        if (headerEncodingLength() + blockLength > shardSize) {
```
Gozala, line 36: Nit: I'd personally defined `const maxBlockLength = shardSize - headerEncodingLength()` on line 22 and then would have `if (blockLength > maxBlockLength)` here which would make logic here a lot more clear and avoid calculating on each block.

packages/upload-client/src/sharding.js
```diff
@@ -22,21 +23,30 @@ export class ShardingStream extends TransformStream {
     let shard = []
     /** @type {import('@ipld/unixfs').Block[] | null} */
     let readyShard = null
-    let size = 0
+    let shardBlockLength = 0
```
Gozala, line 26: Nit: I was confused by the name while reading, I think `currentLength` or `byteLength` would be more intuitive.

packages/upload-client/src/sharding.js
```diff
@@ -22,21 +23,30 @@ export class ShardingStream extends TransformStream {
     let shard = []
```
Gozala, line 23: Nit: I think calling it `blocks` would made it easier to follow the logic below

packages/upload-client/src/sharding.js
```diff
@@ -45,10 +55,29 @@ export class ShardingStream extends TransformStream {
...
-        if (rootBlock != null) {
-          controller.enqueue(
-            await encode(shard, options.rootCID ?? rootBlock.cid)
-          )
+        if (rootBlock == null) return
+
+        const rootCID = options.rootCID ?? rootBlock.cid
+        const headerLength = headerEncodingLength(rootCID)
+        // does the shard with the CAR header that _includes_ a root CID
+        // exceed the shard size?
```
Gozala, line 63:
```suggestion
        // If adding CAR root overflows the shard limit we move overflowing blocks
        // into a another CAR.
```

packages/upload-client/src/sharding.js
```diff
@@ -45,10 +55,29 @@ export class ShardingStream extends TransformStream {
...
+          const lastShard = []
+          let lastShardBlockLength = 0
+          while (lastShardBlockLength < overage) {
+            // need at least 1 block in original shard
+            if (shard.length < 2)
+              throw new Error(`block exceeds shard size: ${shard.at(-1)?.cid}`)
+            const block = shard[shard.length - 1]
+            shard.pop()
+            lastShard.unshift(block)
+            lastShardBlockLength += blockEncodingLength(block)
```
Gozala, line 75: Nit: I was very confused when I read this, comment said `1` block but then we had `< 2`, I think changing order of operations would make it easier to follow

```suggestion
            const block = shard[shard.length - 1]
            shard.pop()
            lastShard.unshift(block)
            lastShardBlockLength += blockEncodingLength(block)
            
            // need at least 1 block in original shard
            if (shard.length < 1)
              throw new Error(`block exceeds shard size: ${shard.at(-1)?.cid}`)
```

I also think that error message is somewhat misleading, it's combination of the block and it's CID what exceeds the block size limit, maybe message can be improved here.

packages/upload-client/src/unixfs.js
Gozala: ☝️ This should probably be a code comment, also I don't think just logging an error is much better, browsers will log such errors and in node you can decide what to do. If we just log error things will appear to work but then break don't they ?

We should probably propagate error into returned readable somehow although I have to say I'm not sure how can we do that.

Any idea in what circumstances writable stream would error, it's not even exposed outside so if it errors perhaps it's problem in the `UnixFS` writer that can be fixed.

Gozala (approved):
Provided some optional feedback here and there, thanks for making it happen!

I'm only worried about `console.error` which I'm not sure is better than throwing, that said I think we should land this and deal with that in the followup.

packages/upload-client/src/unixfs.js
alanshaw: I have removed and refactored the test so we don't encounter it.

What was causing it was the tests that were triggering the `ShardingStream` to error because a block is too big and it caused the CAR to exceed the shard size.

PR #911, opened 2023-09-07
fix: add providers to space/info result type

alanshaw:
Missing exported type.

### DIFF
--- a/packages/access-client/src/types.ts
+++ b/packages/access-client/src/types.ts
@@ -72,6 +72,7 @@ export type SpaceRecord = Selectable<SpaceTable>
 export interface SpaceInfoResult {
   // space did
   did: DID<'key'>
+  providers: Array<DID<'web'>>
 }
 
 export interface AccountTable {
--- a/packages/w3up-client/test/capability/space.test.js
+++ b/packages/w3up-client/test/capability/space.test.js
@@ -21,7 +21,7 @@ describe('SpaceClient', () => {
             return {
               ok: {
                 did: /** @type {`did:key:${string}`} */ (space.did()),
-                providers: []
+                providers: [],
               },
             }
           }),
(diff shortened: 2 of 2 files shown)


travis:
Ahh, I was confused about this because I remembered doing it and realized that the type I updated is in `upload-api` - we currently maintain this type in two places, the place you've changed and https://github.com/web3-storage/w3up/blob/main/packages/upload-api/src/types.ts#L335

In other cases we've solved this by defining types in the `capabilities` package (eg https://github.com/web3-storage/w3up/blob/main/packages/capabilities/src/types.ts#L57) but that also has seemed a little weird to me in the past - it's surprising the the capabilities package would define types for the services one might build with them.

TLDR this PR looks great, but I'd love to talk more about how we'd like to approach this going forward. Options include:

1) Define all service result types in the `capabilities` package and use them in clients and services that use these capabilities (this would likely require servers and clients to evolve in lockstep)
2) Define "server" types in the `upload-api` package and "client" types in the `access-client` and `upload-client` packages (this would duplicate many of these types but also allow clients and servers to evolve separately and might have slightly clearer backwards compatibility semantics)
3) ??

alanshaw:
In [content claims](https://www.npmjs.com/package/@web3-storage/content-claims) we publish 1 package e.g.

```js
import * as Client from '@web3-storage/content-claims/client'
import { Assert } from '@web3-storage/content-claims/capability'
```

...and you can access server types from '@web3-storage/content-claims/server' (or re-export them from the client).

So you can imagine `@web3-storage/w3up/client`, `@web3-storage/w3up/capability` and `@web3-storage/w3up/server` but after loads of refactorings...

Might be easiest to just publish a `@web3-storage/w3up-service` package with just the service types.

travis:
> Might be easiest to just publish a @web3-storage/w3up-service package with just the service types.

Yea this is the direction I lean too - should be relatively straightforward to consolidate the various service types out to a dedicated package.

PR #918, opened 2023-09-11
feat: implement `admin/upload/inspect` and `admin/store/inspect` capabilities

travis:
1. implementations of new capabilities proposed in https://github.com/web3-storage/specs/pull/77
2. added a new `getCID` method to each of the `{Upload|Store}Table` interfaces, and updated the in-memory implementations of each.
3. implementations of `admin/upload/inspect` and `admin/store/inspect` invocation handlers
4. tests for all of the above


Gozala (approved):
@travis I'm approving this as implementation of the spec looks good. That said I do think we need to iterate on the spec (sorry for not reviewing it sooner) to avoid taking up generic namespaces for system introspection capabilities.

Feel free to land as is and we can iterate on naming in followups, or we can align on naming and update this pr before landing.

travis:
No worries at all @Gozala! I've updated this for the latest version of the spec - I'll hold off on merging until I've spiked this out full-stack, so feel free to re-review or not - the latest updates are basically just name changes and restructuring the match, but couldn't hurt to have more eyes.

PR #923, opened 2023-09-13
chore: reorganize tests

travis:
1) move `access/*` handler tests into a new directory and name them consistently with more recently added tests
2) move "storage" and "table" tests into their own directory
3) add `handlerTests` and `storageTests` exports so we can make the w3infra re-use of these tests suites cleaner


packages/upload-api/test/lib.js
```diff
@@ -1,8 +1,19 @@
...
+import * as AdminStoreInspect from './handlers/admin/store/inspect.js'
+import * as AdminUploadInspect from './handlers/admin/upload/inspect.js'
+import * as RateLimitAdd from './handlers/rate-limit/add.js'
+import * as RateLimitList from './handlers/rate-limit/list.js'
+import * as RateLimitRemove from './handlers/rate-limit/remove.js'
+import * as Store from './handlers/store.js'
+import * as Upload from './handlers/upload.js'
+
+
+
```
alanshaw, line 13: Why big gappy?

alanshaw (approved):
LGTM

packages/upload-api/test/lib.js
travis: removed!

PR #925, opened 2023-09-14
feat: compute piece CID in client

alanshaw:
This PR alters the client to compute piece CIDs for each shard, and exposes it via the `CARMetadata` provided to the `onShardStored` callback.

I removed the `ShardStoringStream`. In hindsight, the only real utility it was providing was parallelizing the calls to `Store.add(...)`. I swapped in a parallel transform stream that allows us to do the same thing, but also concurrently calculate the piece CID of the shard.

Note: the client does not yet do anything with the piece CID, but in theory it will be trivial to collect them and then send a bunch of invocations to trigger filecoin.


Gozala:
I think we should wasm version as it's noticeably faster, but I'm still struggling to find a way to make it bundler friendly

vasco-santos (approved):
LGTM
We can swap to https://github.com/web3-storage/fr32-sha2-256-trunc254-padded-binary-tree-multihash/ + hookup this to invocation as follow up PRs

PR #926, opened 2023-09-14
feat: reorg tests

travis:
mostly pushing this in order to trigger a release - I incorrectly called my previous commit a `chore`


travis:
self-merging since it's a whitespace change for side-effects

PR #928, opened 2023-09-15
chore: cleanup space DID types and schema

travis:
@gozala and I are discussing the possibility of allowing `did:web`s to be used as spaces, in order to enable us to `access/delegate` capabilities on service DIDs. As I explored this possibility, I noticed that we are currently a bit sloppy with space DID definitions.

This PR cleans up and moves Space DID definition to the capabilities package. It does not make any changes to the interfaces yet, but puts us in a better place to make the change we are discussing, alongside just making our code cleaner.


packages/upload-client/src/util.js
```diff
@@ -0,0 +1,16 @@
...
+ * @typedef {import("@web3-storage/capabilities/src/types").SpaceDID} SpaceDID
+ */
+
+/**
+ * 
+ * @param {string} did 
+ * @returns {SpaceDID}
+ */
+export function ensureSpaceDID(did){
+  if (did.startsWith('did:key')) {
```
Gozala, line 11: This is more accurate IMO

```suggestion
  if (did.startsWith('did:key:')) {
```

Gozala: I think this is somewhat redundant, because it basically does what `SpaceDID.from` does see

https://github.com/web3-storage/ucanto/blob/c8999a59852b61549d163532a83bac62290b629d/packages/core/src/schema/type.ts#L26-L27

If we really want it perhaps we can define it as `export const  ensureSpaceDID = SpaceDID.from.bind(SpaceDID)` instead.

Gozala (approved):

Gozala:
@travis looks like eslint is unhappy, can you please make tree green before landing.

packages/upload-client/src/util.js
travis: oh awesome, replaced ensureSpaceDID with SpaceDID.from, thanks!

PR #934, opened 2023-09-19
fix: top level await error

Gozala:
Fixes [top level await](https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1695053425167059) problem by upgrading playwright-test. I also fixed a some type mismatches causing typecheck to fail

### DIFF
--- a/packages/upload-client/src/index.js
+++ b/packages/upload-client/src/index.js
@@ -133,7 +133,7 @@ async function uploadBlockStream(conf, blocks, options = {}) {
             const digestBytes = new Uint8Array(PIECE_MULTIHASH_SIZE)
             hasher.write(bytes)
             hasher.digestInto(digestBytes, 0, true)
-            const digest = Digest.create(PieceHasher.code, digestBytes)
+            const digest = Digest.decode(digestBytes)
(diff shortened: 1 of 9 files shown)


packages/upload-client/src/index.js
```diff
@@ -129,15 +130,17 @@ async function uploadBlockStream(conf, blocks, options = {}) {
...
           (async () => {
             const hasher = PieceHasher.create()
-            const digestBytes = new Uint8Array(36)
+            const digestBytes = new Uint8Array(PIECE_MULTIHASH_SIZE)
             hasher.write(bytes)
             hasher.digestInto(digestBytes, 0, true)
-            const digest = Digest.create(PieceHasher.code, digestBytes)
-            return Link.create(raw.code, digest)
-          })()
+            const digest = Digest.decode(digestBytes)
```
Gozala, line 136: @alanshaw `digestBytes` already contained multihash prefix and `Digest.create` was adding another prefix.  I have changed to `Digest.decode` instead which will attempt to interprent given bytes as multihash instead.

alanshaw (approved):

PR #937, opened 2023-09-20
refactor!: filecoin capabilities

alanshaw:
These are the refactored filecoin capabilities as per the updated spec https://github.com/web3-storage/specs/pull/81

I will send a seperate PR targeting this PR to fix up `filecoin-client` and `filecoin-api`.


packages/capabilities/package.json
```diff
@@ -31,6 +31,22 @@
       "types": "./dist/src/*.d.ts",
       "import": "./src/*.js"
     },
+    "./filecoin": {
+      "types": "./dist/src/filecoin/index.d.ts",
+      "import": "./src/filecoin/index.js"
+    },
```
vasco-santos, line 37: nit: If the caps names have been decided with `filecoin/...` on the others I think this would make sense. But in current state, I think that it would be more consistent to expose `filecoin/storefront` here.

packages/capabilities/src/filecoin/aggregator.js
```diff
@@ -0,0 +1,76 @@
...
+      and(checkLink(claim.nb.piece, from.nb.piece, 'nb.piece')) ||
+      and(equal(claim.nb.group, from.nb.group, 'nb.group')) ||
+      ok({})
+    )
+  },
+})
+
+/**
+ * Capability that allows an Aggregator to signal a piece has been accepted
+ * for inclusion in an aggregate.
```
vasco-santos, line 48:
```suggestion
 * Capability that allows an Aggregator to signal a piece has been accepted
 * or rejected for inclusion in an aggregate.
```

vasco-santos: The receipt of this can be either success with inclusion proof, or failed with reason

packages/capabilities/src/filecoin/dealer.js
```diff
@@ -0,0 +1,76 @@
...
+import { equalWith, checkLink, and } from '../utils.js'
+
+/**
+ * Capability allowing an Aggregator to request an aggregate to be added to a
+ * deal with a Storage Provider.
+ */
+export const aggregateOffer = capability({
+  can: 'aggregate/offer',
+  /**
+   * DID of the Aggregator.
```
vasco-santos, line 23:
```suggestion
   * DID of an authorized Storefront.
```

vasco-santos: Per previous discussion, it should be Storefront who requests dealer for something to be offered. Aggregation is an implementation detail for storefront perspective.

For now in implementation, we will just use same DID for Aggregator and Storefront, but to make sure we are using the Storefront one

vasco-santos:
@alanshaw please sign this as a breaking change :)

packages/capabilities/src/types.ts
```diff
@@ -158,51 +161,118 @@ export type Space = InferInvokedCapability<typeof space>
 export type SpaceInfo = InferInvokedCapability<typeof info>
 
 // filecoin
-export type FILECOIN_PROCESSING_STATUS = 'pending' | 'done'
-export interface FilecoinAddSuccess {
-  piece: PieceLink
+export interface DataAggregationProof {
```
vasco-santos, line 164:
```suggestion
/** @see https://github.com/filecoin-project/go-data-segment/blob/e3257b64fa2c84e0df95df35de409cfed7a38438/datasegment/verifier.go#L8-L14 */
export interface DataAggregationProof {
```

packages/capabilities/src/types.ts
```diff
@@ -158,51 +161,118 @@ export type Space = InferInvokedCapability<typeof space>
...
+  /**
+   * Proof the piece is included in the aggregate.
+   */
+  inclusion: InclusionProof
+  auxDataType: uint64
+  auxDataSource: SingletonMarketSource
+}
+export interface InclusionProof {
```
vasco-santos, line 172:
```suggestion
/** @see https://github.com/filecoin-project/go-data-segment/blob/e3257b64fa2c84e0df95df35de409cfed7a38438/datasegment/inclusion.go#L30-L39 */
export interface InclusionProof {
```

packages/capabilities/src/types.ts
```diff
@@ -158,51 +161,118 @@ export type Space = InferInvokedCapability<typeof space>
...
+  /**
+   * Commitment proof for piece.
+   */
   piece: PieceLink
-  aggregate?: PieceLink
-}
-export interface AggregateAddFailure extends Ucanto.Failure {
-  name: string
 }
+export type FilecoinOfferFailure = Ucanto.Failure
```
vasco-santos, line 198: We are missing here a `ContentNotFoundError` per spec https://github.com/web3-storage/specs/blob/feat/update-filecoin-spec/w3-filecoin.md#filecoinoffer

vasco-santos (approved):
LGTM

Mostly small details on JS docs. There is specially one comment that should be handled: `ContentNotFoundError` is missing

packages/capabilities/src/types.ts
```diff
@@ -158,51 +161,118 @@ export type Space = InferInvokedCapability<typeof space>
...
+// filecoin dealer
+export interface AggregateOfferSuccess {
+  /**
+   * Commitment proof for aggregate.
+   */
+  aggregate: PieceLink
+  /**
+   * Link to a dag-cbor node with details of content pieces.
+   */
+  pieces: Link
```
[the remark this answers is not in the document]
alanshaw: Good catch

packages/capabilities/package.json
alanshaw: Ok how about this: I've moved all the storefront caps into `storefront.js` and now `index.js` just re-exports the caps in `storefront.js` without the "filecoin" prefix. i.e. `filecoinOffer` => `offer`.

So then most people can do:

```js
import * as Filecoin from '@web3-storage/capabilities/filecoin'

Filecoin.offer
```

Which works with all our existing caps.

And equivalently:

```js
import * as Storefront from '@web3-storage/capabilities/filecoin/storefront'

Storefront.filecoinOffer
```

vasco-santos: cool, this sounds great!

packages/capabilities/src/filecoin/dealer.js
```diff
@@ -0,0 +1,76 @@
...
+  nb: Schema.struct({
+    /**
+     * Commitment proof for the aggregate being offered.
+     */
+    aggregate: PieceLink,
+    /**
+     * CID of the DAG-CBOR encoded block with offer details.
+     * Service will queue given offer to be validated and handled.
+     */
+    pieces: Schema.link(),
```
vasco-santos, line 66: While working on the CRON that will read pending deals to issue `aggregate/accept` for the ones already on chain, I questioned this need.

`pieces` is quite big list, for which we dag encode it, and inline block with the block. We then store it.

But, we also store it on `aggregate/offer`. This list is actually stored both in the UCAN Stream CAR bucket, as well as in the store where Spade reads from.

This brings the question, should we don't care about putting `pieces` here? There will only be one aggregate with that CID, and we already store its content in 2 different places. Sending it to self invoke `aggregate/accept` from Dealer seems useless and requires us to write a new copy of this list on our UCAN Stream CAR bucket. Moreover, makes implementation more complex and expensive. After we track deals and see that a deal is now on chain, we only have the aggregateCid in the deals Table. So, for then invoking `aggregate/accept` we would need to read the list from one of the already written places and put it there.

We could do a middle term as well, and store the `invocationCid` + `taskCid` (of `aggregate/offer` on the `deal-store` and add them here instead).

PR #939, opened 2023-09-29
fix: clarity for bring your own agent example

dchoi27:


packages/w3up-client/README.md
```diff
@@ -175,7 +175,7 @@ Calling `authorize` will cause an email to be sent to the given address. Once a
 
 ##### Bringing your own Agent and delegation
 
-For uses of `w3up-client` in environments where the Agent is not persisted and/or the email verification step would be prohibitive (e.g., serverless backend environment where local Store with the Agent is dropped in between runs, and going through the email verification flow isn't practical), you can manually add a delegation for access to a space created by a different authorized agent (see the [`addSpace` client method](docs-client#addSpace)). An example:
+For uses of `w3up-client` in environments where the Agent is not persisted and/or the email verification step would be prohibitive (e.g., serverless backend environment where local Store with the Agent is dropped in between runs, and going through the email verification flow isn't practical), you can manually add a delegation for access to a Space created by a different authorized agent (see the [`addSpace` client method](docs-client#addSpace)). An example using w3cli, assuming it already is set up with a Space (e.g., already ran `w3 space create`):
```
travis, line 178: the last sentence here is confusing - the example doesn't use w3cli. I think this might be referring to the comment - if that's the case I think it might make more sense to talk about `w3 space create` in the comment that describes creating PROOF env var value

travis (changes requested):
I think the reference to `w3 space create` should be moved down into the comment that describes `PROOF` - what do you think?

packages/w3up-client/README.md
dchoi27: oh you're right it's ambiguous. i previously had it the way you described, but having multiple w3cli commands in-line in the code was confusing

dchoi27: OK how about now?

travis: Looks good, thanks!!

travis (approved):

PR #940, opened 2023-10-02
fix: type error in access client types

travis:
This has had red squigglies in VSCode for a bit, but I assumed it was superfluous until I saw the error in a downstream project I was working on earlier today:

Screenshot 2023-10-02 at 4 21 49 PM

The issue seems to stem from the generic type `R` on `InvokeOptions` - when I use the reified `Resource` rather than the generic `R` it goes away. 

With `R`:

Screenshot 2023-10-02 at 4 19 51 PM
Screenshot 2023-10-02 at 4 19 36 PM

With `Resource`:

Screenshot 2023-10-02 at 4 20 04 PM

I don't fully understand why this fixes the error I was seeing, but a) it does and b) I don't see much value in the generic type here.

Pushing this up to start a discussion about whether this is an acceptable way to fix this issue - would love any thoughts on a better way to fix if anyone has ideas!


travis:
fixed in https://github.com/web3-storage/w3up/pull/962

PR #941, opened 2023-10-03
feat: add `RevocationsStorage`

travis:
Add a new to the `RevocationsStorage` interface:

`getAll` takes a list of delegation CIDs and returns any `Revocation`s in the store whose `revoke` field matches one of the given CIDs

`addAll` takes a list of `Revocation`s and adds them to the revocation store.

Also introduce tests for the described behavior.

### DIFF
--- a/packages/upload-api/src/types/delegations.ts
+++ b/packages/upload-api/src/types/delegations.ts
@@ -43,12 +43,12 @@ export interface DelegationsStorage<
   >
 
   /**
-   * Given a list of invocation CIDs, return a Ucanto Result with a boolean
-   * success value that will be true if any of the identified invocations
+   * Given a list of delegation CIDs, return a Ucanto Result with a boolean
+   * success value that will be true if any of the identified delegations
    * have been revoked and false if all are valid.
    */
   areAnyRevoked: (
-    invocationCids: Ucanto.Link[]
+    delegationCids: Ucanto.Link[]
   ) => Promise<
     Ucanto.Result<Boolean, Ucanto.Failure>
   >
--- a/packages/upload-api/src/types/delegations.ts
+++ b/packages/upload-api/src/types/delegations.ts
@@ -5,6 +5,12 @@ interface ByAudience {
 }
 export type Query = ByAudience
 
+export type Revocation = {
+  iss: Ucanto.DID
+  revoke: Ucanto.Link
+  challenge: string
+}
+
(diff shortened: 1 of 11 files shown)


packages/upload-api/src/types/delegations.ts
```diff
@@ -5,6 +5,16 @@ interface ByAudience {
 }
 export type Query = ByAudience
 
+export interface RevocationMeta {
+  context: Ucanto.Link
```
Gozala, line 9: Perhaps we should call this a `scope` instead, as in UCAN has been revoked in this scope.

```suggestion
  scope: Ucanto.Link
```

packages/upload-api/src/types/delegations.ts
```diff
@@ -5,6 +5,16 @@ interface ByAudience {
 }
 export type Query = ByAudience
 
+export interface RevocationMeta {
```
Gozala, line 8: I'm not sure what does `Meta` imply here, maybe `RevocationDetails` would be less confusing name

packages/upload-api/src/types/delegations.ts
```diff
@@ -34,4 +44,32 @@ export interface DelegationsStorage<
...
+   * Given a list of delegation CIDs, return a Ucanto Result with a map from
+   * some or all of the CIDs to a list of "revocation context CIDs" for a
+   * given CID - that is, a list of delegation CIDs that should no longer
+   * be considered valid proof for the given CID.
+   */
+  getRevocations: (
+    delegationCIDs: Ucanto.Link[]
+  ) => Promise<
+    Ucanto.Result<RevocationsToMeta, Ucanto.Failure>
+  >
```
Gozala, line 58: I wonder if following would be a lot simpler 

```ts
interface RevocationStore {
  getAll(query: Revocation['delegation'][]): Promise<Result<Revocation[], Failure>>
  addAll(revocations: Revocation[]): Promise<Result<Unit, Failure>>
}


interface Revocation {
  delegation: UCANLink
  scope: UCANLink
  cause: UCANLink
}
```

Gozala (approved):
Provided some suggestions, but current version also works

packages/upload-api/src/types/delegations.ts
travis: I like it! will get this updated and shipped

PR #942, opened 2023-10-04
feat: add `store/get` and `upload/get` capabilities

olizilla:
Capability definitions to allow clients to check for inclusion of cid in store shards or upload roots. Also allows client to fetch the additional properties we store. This will, for example, let a client verify all shards for an upload have been removed before removing an upload.

_I'm aiming to simplify things with this PR so that we return insertedAt info on the Upload/Store ListItem types, and return the same object shape for a single object get. If this is pleasing, then I will continue in this direction and make Remove also return the same shape if it removes something_

Depends on approval of the new capabilities in https://github.com/web3-storage/specs/pull/82

TODO:
- [x] spec update https://github.com/web3-storage/specs/pull/82
- [x] tests

License: MIT


packages/capabilities/src/types.ts
```diff
@@ -221,14 +221,22 @@ export interface ChainTrackerInfoFailure extends Ucanto.Failure {
...
 // Store
-export type Store = InferInvokedCapability<typeof store>
-export type StoreAdd = InferInvokedCapability<typeof add>
-export type StoreRemove = InferInvokedCapability<typeof remove>
-export type StoreList = InferInvokedCapability<typeof list>
+export type Store = InferInvokedCapability<typeof StoreCaps.store>
+export type StoreAdd = InferInvokedCapability<typeof StoreCaps.add>
+export type StoreGet = InferInvokedCapability<typeof StoreCaps.get>
+export type StoreRemove = InferInvokedCapability<typeof StoreCaps.remove>
+export type StoreList = InferInvokedCapability<typeof StoreCaps.list>
```
olizilla, line 239: Update to pull each cap off of StoreCaps to be consistent with all the other caps in this file.

packages/capabilities/src/types.ts
```diff
@@ -271,13 +283,21 @@ export interface StoreListItem {
   link: UnknownLink
   size: number
   origin?: UnknownLink
+  insertedAt: string
```
olizilla, line 286: add insertedAt to StoreListItem as we already want to return this info for list usage, and it means we can re-use the type for both list and get operations.

packages/capabilities/src/types.ts
```diff
@@ -271,13 +283,21 @@ export interface StoreListItem {
...
 }
 
-export interface UploadAddSuccess {
+export interface UploadListItem {
   root: UnknownLink
   shards?: CARLink[]
+  insertedAt: string
+  updatedAt: string
```
olizilla, line 293: Add insertedAt and updatedAt props to UploadListItem. We want to share a timestamp with the user about when this upload was added, and these are the props we have. `updatedAt` currently means "when did we last add a shard to this upload", while `insertedAt` means "when did you first add a given root cid as an upload to this space."

packages/upload-api/test/handlers/store.js
```diff
@@ -72,18 +71,16 @@ export const test = {
 
     assert.equal(goodPut.status, 200, await goodPut.text())
 
-    const item = await context.testStoreTable.get(spaceDid, link)
+    const item = await context.storeTable.get(spaceDid, link)
```
olizilla, line 74: `storeTable.get` is now a thing, so we no longer need the `testStoreTable`.

packages/upload-api/test/handlers/store.js
```diff
@@ -72,18 +71,16 @@ export const test = {
...
     if (!item) {
       return assert.equal(item != null, true)
     }
 
     assert.deepEqual(
       {
-        space: item.space,
         link: item.link.toString(),
         size: item.size,
-        issuer: item.issuer,
```
olizilla, line 86: i don't think we need to test things here that are not part of the contract

packages/capabilities/src/types.ts
```diff
@@ -257,6 +265,10 @@ export interface StoreItemNotFound extends Ucanto.Failure {
 
 export type StoreRemoveFailure = StoreItemNotFound | Ucanto.Failure
 
+export type StoreGetSuccess = StoreListItem
+
+export type StoreGetFailure = StoreItemNotFound | Ucanto.Failure
```
alanshaw, line 270: This is returned as `ShardNotFound` in implementation...

packages/upload-api/src/store/get.js
```diff
@@ -0,0 +1,29 @@
+import * as Server from '@ucanto/server'
+import * as Store from '@web3-storage/capabilities/store'
+import * as API from '../types.js'
+
+/**
+ * @param {API.StoreServiceContext} context
+ * @returns {API.ServiceMethod<API.StoreGet, API.StoreGetSuccess, API.Failure>}
```
alanshaw, line 7:
```suggestion
 * @returns {API.ServiceMethod<API.StoreGet, API.StoreGetSuccess, API.StoreGetFailure>}
```

packages/upload-api/src/types.ts
```diff
@@ -372,37 +378,27 @@ export interface DudewhereBucket {
...
-    link: UnknownLink
-  ): Promise<
-    (StoreAddInput & StoreListItem & { insertedAt: string }) | undefined
-  >
+    ) => Promise<ListResponse<StoreAddInput & StoreListItem>>
 }
 
 export interface UploadTable {
-  inspect: (link: UnknownLink) => Promise<UploadGetOk>
+  inspect: (link: UnknownLink) => Promise<UploadInspectOk>
```
alanshaw, line 393: Would be nice if these were named consistently like the others...

```suggestion
  inspect: (link: UnknownLink) => Promise<UploadInspectSuccess>
```

packages/upload-api/src/upload/get.js
```diff
@@ -0,0 +1,29 @@
+import * as Server from '@ucanto/server'
+import * as Upload from '@web3-storage/capabilities/upload'
+import * as API from '../types.js'
+
+/**
+ * @param {API.UploadServiceContext} context
+ * @returns {API.ServiceMethod<API.UploadGet, API.UploadGetSuccess, API.Failure>}
```
alanshaw, line 7:
```suggestion
 * @returns {API.ServiceMethod<API.UploadGet, API.UploadGetSuccess, API.UploadGetFailure>}
```

alanshaw (changes requested):

packages/capabilities/src/store.js
```diff
@@ -77,6 +77,28 @@ export const add = capability({
...
+ * When used as as an invocation, `nb.link` must be specified.
+ */
+export const get = capability({
+  can: 'store/get',
+  with: SpaceDID,
+  nb: Schema.struct({
+    /**
+     * shard CID to fetch info about.
+     */
+    link: Link.optional(),
```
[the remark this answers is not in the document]
olizilla: is interesting point, but i'm gonna leave it separate for now.

packages/capabilities/src/types.ts
olizilla: Fixed!

alanshaw (approved):
LGTM

PR #943, opened 2023-10-05
feat: define ucan/revoke capability

Gozala:
Define `ucan/revoke` capability


packages/capabilities/src/ucan.js
```diff
@@ -0,0 +1,52 @@
...
+ * `ucan/revoke` capability is a replacement for the
+ * [UCAN Revocation](https://github.com/ucan-wg/spec#66-revocation) that had
+ * been proposed to a UCAN working group and had a tentative support from
+ * members.
+ */
+export const revoke = capability({
+  can: 'ucan/revoke',
+  /**
+   * With MUST be a DID of the UCAN issuer that is in the proof chain of the
+   * delegation been revoked.
```
travis, line 33: is this always the DID of the issuer of the "context CID" ?

packages/capabilities/src/ucan.js
```diff
@@ -0,0 +1,52 @@
...
+  with: Schema.did(),
+  nb: Schema.struct({
+    /**
+     * Link of the UCAN been revoked, it MUST be a UCAN be either issued by a
+     * principal matching `with` field or depend on the delegation issued by
+     * the principal matching `with` field.
+     *
+     * Alternatively `with` field MAY match the `audience` of the this UCAN,
+     * which would imply that that delegate is revoking capabilities delegated
+     * to it. This allows delegate to proof that it is unable to invoke
```
travis, line 44:
```suggestion
     * to it. This allows delegate to prove that it is unable to invoke
```

travis: should this be "this allows the delegate to prove that it is **able** to invoke the delegated capabilities"? I'm not sure I follow how it would prove that it's unable to invoke the capabilities, or why it would do that...

travis (approved):
looks good! a couple questions that could probably be clarified with more comments, and one suggested tweak, but good to merge

packages/capabilities/src/ucan.js
Gozala: It is otherwise anyone could request a revocation. Put it differently that is how we infer context / scope CID, by walking up the chain and finding a proof that corresponds to `with`.

packages/capabilities/src/ucan.js
```diff
@@ -0,0 +1,52 @@
...
+   * delegation been revoked.
+   */
+  with: Schema.did(),
+  nb: Schema.struct({
+    /**
+     * Link of the UCAN been revoked, it MUST be a UCAN be either issued by a
+     * principal matching `with` field or depend on the delegation issued by
+     * the principal matching `with` field.
+     *
+     * Alternatively `with` field MAY match the `audience` of the this UCAN,
```
Gozala, line 42:
```suggestion
     * Alternatively `with` field MAY match the `audience` of the UCAN been revoked,
```

packages/capabilities/src/ucan.js
Gozala: No, maybe comment could be improved here, but e.g. consider following scenario. You `Alice` delegated capability to fire a nuke to `Bob`. `Bob` can revoke this delegated capability and use the revocation receipt to prove that he could not fire a nuke after the date of revocation.

This was brought up in the UAN WG group call as an important ability.

travis: ahh, I see, got it, thanks!

PR #996, opened 2023-10-24
fix: upgrade data segment v4

vasco-santos:
use new Piece CID from `data-segment` https://github.com/web3-storage/data-segment/pull/36


Gozala (approved):
Looks good to me

PR #997, opened 2023-10-24
fix: upgrade fr32-sha2-256-trunc-254-padded-binary-tree-multihash

vasco-santos:
Upgrades lib in `upload-client` to use PieceCidv2


packages/upload-client/src/index.js
```diff
@@ -13,7 +13,7 @@ export { Store, Upload, UnixFS, CAR }
 export * from './sharding.js'
 
 const CONCURRENT_REQUESTS = 3
-const PIECE_MULTIHASH_SIZE = PieceHasher.prefix.length + PieceHasher.size
+const PIECE_MULTIHASH_SIZE = PieceHasher.prefix.length + 1 + PieceHasher.size
```
Gozala, line 16: This is no longer correct as size will vary based on padding applied.

Gozala:
@vasco-santos I think error occurs because of the buffer overflow, that is supplied Uint8Array is smaller than amount of bytes been written into it. I'll test that hypothesis and will see if I can make tests case for fr32...hash to improve error messages.

Gozala (approved):

Gozala (approved):
Not sure why tests are failing, but changes look good to me.

vasco-santos:
> Not sure why tests are failing, but changes look good to me.

forgot to update lock file 😓

Issue #982, opened 2023-10-19
The Types Are Wrong!


travis:
Plugging a couple of our libraries into https://arethetypeswrong.github.io/ shows that our types could use some improvements:

https://arethetypeswrong.github.io/?p=%40web3-storage%2Fw3up-client%409.0.0
https://arethetypeswrong.github.io/?p=%40web3-storage%2Faccess%409.0.0

<img width="990" alt="Screenshot 2023-10-18 at 4 59 13 PM" src="https://github.com/web3-storage/w3up/assets/1113/e7b52d77-d65d-47cc-a486-0585b04ced66">
<img width="805" alt="Screenshot 2023-10-18 at 4 59 01 PM" src="https://github.com/web3-storage/w3up/assets/1113/cccc4488-c4d3-4a21-98de-e7725666c314">

Fortunately they also have a wonderful key for the errors they show, with a helpful page for fixing each of them:

https://github.com/arethetypeswrong/arethetypeswrong.github.io#readme

We should fix the types on the most forward-facing stuff first because I think this is actively hampering downstream development with our libraries, and it would be good to check all of our libraries soon.

```[tasklist]
### Tasks
- [ ] https://github.com/web3-storage/w3up/pull/1004
- [ ] https://github.com/web3-storage/w3up/issues/1013
- [ ] https://github.com/web3-storage/w3up/issues/1015
- [ ] https://github.com/web3-storage/w3up/issues/1016
- [ ] https://github.com/web3-storage/w3up/pull/1014
```

gobengo:
@travis will you plz review https://github.com/web3-storage/w3up/pull/1004 ? I think it may resolve this issue.

gobengo:
After the fix from #1004, which affected how npm/pnpm are configured in package.jsons in a way that could affect the release process of our packages, I wanted to do a test release of a package. I just did it with did-mailto and the release went ok and I confirmed the new release of did-mailto resolves arethetypesworking errors as I expected https://github.com/web3-storage/w3up/issues/1013#issuecomment-1779827607

I chose did-mailto because it's the simplest package and in case things went haywire, I think it has the least external dependers.

I'm more worried about the other 3 packages (*-client) that I touched in this PR. https://github.com/web3-storage/w3up/pull/1012/files
Next I'll test one of them

gobengo:
The original issue here shows arethetypesworking errors for the two packages of `@web3-storage/access` and `@web3-storage/w3up-client`.

[...]

PR #1004, opened 2023-10-24
fix: fix arethetypesworking errors in all packages

gobengo:
Motivation:
* https://github.com/web3-storage/w3up/issues/982

Learned While Devving
* #982 indicated 3 errors coming from atty. Here is how each error type was resolved
  1. `ESM (dynamic import only)` - this is more of a warning than an error. It's warning, rightly, that versions of node that only use CommonJS requires (e.g. `require("@web3-storage/did-mailto")`) won't be able to load these libraries. i.e. these libraries are "ESM-only" in that they only work on runtimes that can do ESM-style `import()`. By default, attw warns on this, but this PR configures attw to ignore instead (via .attw.json https://github.com/web3-storage/w3up/pull/1004/files#diff-a3be4137698f4bdc5777d3831fac4e67d11c102c5e57dd54906ff333cb551aff ).
  2.  [internal resolution error](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/InternalResolutionError.md)
      * Insight via https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/InternalResolutionError.md#declarations-and-javascript-match-but-are-not-checked-with-correct-settings
        > This problem can occur with tsc alone if the library author compiles with --moduleResolution node (also known as node10) or bundler, especially in combination with a package.json that has "type": "module" 
      * This attw error was resolved by this PR but this PR changing all packages `tsconfig.json` to use moduleResolution=NodeNext (not `Node`). To do that, many type imports had to change so the imported module identifier includes a file path suffix
  3. [no types](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/UntypedResolution.md) - Most of these were due to the relevant package having sub packabe submodule listed in package.json `exports` but without a corresponding js file existing (e.g. it was an old package submodule that has since been removed)

Also
* added an npm script `attw` to run attw for the package. Now can run attw for all packages like `pnpm -r attw` from monorepo root
* gitlab ci workflows for the two packages shown in #982 now run attw


packages/access-client/package.json
```diff
@@ -24,12 +25,12 @@
     "rc": "npm version prerelease --preid rc"
   },
   "exports": {
-    ".": "./src/index.js",
-    "./agent": "./src/agent.js",
-    "./drivers/*": "./src/drivers/*.js",
-    "./stores/*": "./src/stores/*.js",
-    "./types": "./src/types.js",
-    "./encoding": "./src/encoding.js"
+    ".": "./dist/src/index.js",
```
gobengo, line 28: changing where this points to resolved some atty errors. When we do it this way and let typescript also write the type declaration files with all its defaults next to the source files, then when another tsc in a depending lib finds all this config, it'll be able to find the src,types,etc again. I learned some of this from https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md#common-causes

packages/access-client/src/agent-data.js
```diff
@@ -6,16 +6,16 @@ import { CID } from 'multiformats'
 import { Access } from '@web3-storage/capabilities'
 import { isExpired } from './delegations.js'
 
-/** @typedef {import('./types').AgentDataModel} AgentDataModel */
+/** @typedef {import('./types.js').AgentDataModel} AgentDataModel */
```
gobengo, line 9: This is one of many small changes to ensure all imported module identifiers (e.g. type imports in comments like this) have a file suffix, which is required by `tsc` now that this PR enables `moduleResolution=NodeNext` for all tsc packages.

tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "allowJs": true,
     "checkJs": true,
     "target": "ES2022",
-    "module": "ES2022",
```
gobengo, line 9: This change inspired by
https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/64b1f489736bd48ebb4e3b72fb078b44437d59e0/docs/problems/InternalResolutionError.md
> This problem can occur with tsc alone if the library author compiles with --moduleResolution node (also known as node10) or bundler, especially in combination with a package.json that has "type": "module"

and

> Library authors should always compile/check with --module node16 --moduleResolution node16 or nodenext (note that module im

packages/upload-api/src/consumer/has.js
```diff
@@ -6,6 +6,7 @@ import { Consumer } from '@web3-storage/capabilities'
  * @param {API.ConsumerServiceContext} context
  */
 export const provide = (context) =>
+  // @ts-ignore
```
travis, line 9: could you add a comment here justifying the ignore?

packages/upload-client/test/unixfs.test.js
```diff
@@ -1,6 +1,7 @@
 import assert from 'assert'
 import { decode, NodeType } from '@ipld/unixfs'
 import { exporter } from 'ipfs-unixfs-exporter'
+// @ts-ignore
```
travis, line 4: here too plz

travis (approved):
Looks great! have a couple requests, and I'd feel a lot comfier if @alanshaw or @olizilla got a chance to take a look, but happy to see it merge as-is

packages/upload-client/test/unixfs.test.js
gobengo: updated https://github.com/web3-storage/w3up/pull/1004/commits/df5995bbf679029a9a9a4f5c52984a9eacf1c67e

packages/upload-api/src/consumer/has.js
gobengo: I removed it! good catch. My local checker was in a weird state across my editor and local terminal and ci, but I think it'll work now https://github.com/web3-storage/w3up/pull/1004/commits/e9882a4c4a7507dcf22ea31e090d84dee376b999

packages/access-client/package.json
alanshaw: Can you update the npm published files so we don't include src in the root?

alanshaw (approved):

packages/access-client/package.json
gobengo: @alanshaw https://github.com/web3-storage/w3up/pull/1012

Issue #959, opened 2023-10-09
Pricing capabilities


Gozala:
We want to have capabilities to switch between pricing plans. Here is the preliminary list

### Check current plan

```js
{
   "can": "plan/get",
   "with": "did:mailto:web.mail:alice"
}
```

```js
// receipt
{
   out:
     ok: {
        updatedAt: 1696865150828, // timestamp when plan was set
        product: "did:web:starter.w3up.storage"
     }
   }
}
```

### Update plan

```js
{
  "can": "plan/set"
  "with": "did:mailto:web.mail:alice"
  "nb": {
     "product": "did:web:lite.w3up.storage"
  }
}
```

```js
// receipt
{
   out:
     ok: {
        updatedAt: 1696865387813, // timestamp when plan was set
        product: "did:web:lite.w3up.storage"
     }
   }
}
```

### List plans (user can choose from)

```js
{
  "can": "plan/list",
  "with": "did:mailto:web.mail:alice"
}
```

```js
{
  out: {
    ok: {
      "did:web:starter.w3up.storage": {
          ???
      },
      "did:web:lite.w3up.storage": {
         ???
      },
      "did:web:expert.w3up.storage": {
          ??
      }
    }
  }
}
```

Gozala:
And here is the my sketch of dynamo table

```js
export const customerTableProps = {
  fields: {
    /**
     * CID of the UCAN invocation that set it to the current value.
     */
    cause: 'string',
    /**
     * DID of the user account e.g. `did:mailto:agent`.
     */
    customer: 'string',
    /**
     * Opaque identifier representing an account in the payment system
     * e.g. Stripe customer ID (stripe:cus_9s6XKzkNRiz8i3)
     */
    account: 'string',
    /**
     * Unique identifier of the product a.k.a tier.
     */
    product: 'string',
    insertedAt: 'string', // `2022-12-24T...`
    updatedAt: 'string', // `2022-12-24T...`
  },
  primaryIndex: { partitionKey: 'customer' },
}
```

travis:
>         updatedAt: 1696865150828, // timestamp when plan was set

We have generally used ISO8601 date strings in capabilities, so I think it makes sense to stick with them. 

Otherwise this all looks great!

PR #1005, opened 2023-10-24
feat: implement `plan/get` capability

travis:
In console, we need a way to tell if a user has a subscription. Implement the `plan/get` capability from https://github.com/web3-storage/w3up/issues/959 to enable that.

Includes the core capability definitions plus `upload-api` handlers and `access-client` helper functions.


packages/access-client/test/agent-use-cases.test.js
```diff
@@ -188,3 +192,50 @@ describe('authorizeWaitAndClaim', async function () {
...
+  await Promise.all([
+    ...await createAuthorization({ account: accountWithAPlan, agent: agent.issuer, service: server.id }),
+    ...await createAuthorization({ account: accountWithoutAPlan, agent: agent.issuer, service: server.id })
+  ].map(proof => agent.addProof(proof)))
+
+  it("should succeed for accounts with plans", async function () {
+    assert((await getAccountPlan(agent, accountWithAPlan)).ok)
+  })
+
+  it("should fail for accounts without a plan", async function () {
```
gobengo, line 238: could be nice to assert the error here. i.e. if I am a client and I try to plan/get, and there is an error, how do I know whether the error is a sign that my account has no plan (but a UI could still help user set one) vs a sign of something more wrong in the system (and e.g. a UI should show a 'come back later' state if the error is unknown).

gobengo: e.g. I think this would be a PlanNotFound ?

packages/capabilities/test/capabilities/plan.test.js
```diff
@@ -0,0 +1,96 @@
...
+    const result = await access(await auth.delegate(), {
+      capability: Plan.get,
+      principal: Verifier,
+      authority: service,
+      validateAuthorization,
+    })
+    assert.equal(result.error?.message.includes('not authorized'), true)
+  })
+
+  it('can delegate plan/add', async function () {
```
gobengo, line 68:
```suggestion
  it('can delegate plan/get', async function () {
```

packages/upload-api/test/handlers/plan.js
gobengo, line 1: Is this file really testing anything 'plan' related? I expected it to be based on the filename being `plan.js`

packages/upload-api/test/handlers/plan.spec.js
gobengo, line 1: Is this rename intentional? The file contents don't mention plans

travis: oh jeez, copy/paste fail, thx!

packages/capabilities/src/types.ts
```diff
@@ -514,6 +515,21 @@ export type AggregateAccept = InferInvokedCapability<
   typeof DealerCaps.aggregateAccept
 >
 export type DealInfo = InferInvokedCapability<typeof DealTrackerCaps.dealInfo>
+
+// Plan
+
+export type PlanGet = InferInvokedCapability<typeof PlanCaps.get>
+export interface PlanGetSuccess {
+  updatedAt: string
```
gobengo, line 523: What kind of string? Reading the code it looks like this will usually be an ISO8601 datetime string.
* I think that's a great choice
* notably, it is different than the number used in https://github.com/web3-storage/w3up/issues/959 . But again, I prefer a string to a number here
* string/number aside, here in the `PlanGetSuccess` typescript type, IMO it's worth adding a comment explaining constraints like ISO8601 string
  * may also be testing this in relevant tests that receive a PlanGetSuccess, i.e. can the PlanGetSuccess#updatedAt value be parsed as ISO8601 to a realistic datetime

packages/upload-api/test/handlers/plan.spec.js
travis: ok this is fixed - new test added

packages/access-client/test/agent-use-cases.test.js
travis: great idea - done

packages/capabilities/src/types.ts
travis: ah good call, agreed! I've left a comment on #959 to this effect, and have created a new `string` alias called `ISO8601Date` so that this is clearer at least. also added a test to the handler that verifies we keep dates in 8601 format throughout the stack.

gobengo (approved):
Looks like recent commits incorporated all my first wave of feedback

PR #1006, opened 2023-10-25
chore: make linting consistent across projects

travis:
in some cases we weren't running prettier, and in some cases we weren't linting - try to make things more consistent across packages


gobengo:
I changed the linting a bit in https://github.com/web3-storage/w3up/pull/1058

PR #1038, opened 2023-10-30
fix: aggregate accept should rely on deal tracker response to issue receipt

vasco-santos:
Otherwise, we need to have deal in the table

### DIFF
--- a/packages/filecoin-api/test/services/dealer.js
+++ b/packages/filecoin-api/test/services/dealer.js
@@ -1,5 +1,7 @@
 import { Dealer } from '@web3-storage/capabilities'
 import * as Signer from '@ucanto/principal/ed25519'
+import * as Server from '@ucanto/server'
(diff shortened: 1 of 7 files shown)


packages/filecoin-api/test/services/dealer.js
```diff
@@ -264,12 +294,12 @@ async function getServiceContext() {
 
 /**
  * @param {API.Test<DealerApi.ServiceContext>} testFn
- * @param {(context: DealerApi.ServiceContext) => DealerApi.ServiceContext} mockContextFunction
+ * @param {(context: DealerApi.ServiceContext) => Promise<DealerApi.ServiceContext>} mockContextFunction
  */
 function wichMockableContext(testFn, mockContextFunction) {
```
alanshaw, line 299:
```suggestion
function withMockableContext(testFn, mockContextFunction) {
```

alanshaw (approved):

PR #113, opened 2022-10-18
chore(main): release access-api 4.0.0

hugomrdias:
:robot: I have created a release *beep* *boop*
---

### ⚠ BREAKING CHANGES

* upgrade access-api @ucanto/* and @ipld/dag-ucan major versions ([#246](https://github.com/web3-storage/w3protocol/issues/246))
* access-client store decoupling ([#228](https://github.com/web3-storage/w3protocol/issues/228))
* upgrade to `@ucanto/{interface,principal}`@^4.0.0 ([#238](https://github.com/web3-storage/w3protocol/issues/238))
* follow up on the capabilities extract ([#239](https://github.com/web3-storage/w3protocol/issues/239))
* doc capabilities & make requierd nb non-optionals ([#159](https://github.com/web3-storage/w3protocol/issues/159))
* Remove 0.8 caps and add account delegation to the service ([#123](https://github.com/web3-storage/w3protocol/issues/123))
* bump to 0.9 ([#116](https://github.com/web3-storage/w3protocol/issues/116))

### Features

* [#153](https://github.com/web3-storage/w3protocol/issues/153) ([#177](https://github.com/web3-storage/w3protocol/issues/177)) ([d6d448c](https://github.com/web3-storage/w3protocol/commit/d6d448c16f188398c30f2d1b83f69e1d7becd450))
* access-api uses DID env variable when building its ucanto server id ([#275](https://github.com/web3-storage/w3protocol/issues/275)) ([311da78](https://github.com/web3-storage/w3protocol/commit/311da786b8f2fdf5ac585170e746a5a7b04ec57b))
* **access-client:** cli and recover ([#207](https://github.com/web3-storage/w3protocol/issues/207)) ([adb3a8d](https://github.com/web3-storage/w3protocol/commit/adb3a8d61d42b31f106e86b95faa3e442f5dc2c7))
* account recover with email ([#149](https://github.com/web3-storage/w3protocol/issues/149)) ([6c659ba](https://github.com/web3-storage/w3protocol/commit/6c659ba68d23c3448d5150bc76f1ddcb91ae18d8))
* bump to 0.9 ([#116](https://github.com/web3-storage/w3protocol/issues/116)) ([3e0b63f](https://github.com/web3-storage/w3protocol/commit/3e0b63f38aace3a86655a1aa40e529c1501dc136))
[...]

PR #1047, opened 2023-10-31
feat: access agent proofs method would fail to return some session proofs

gobengo:
Motivation:
* fix w3cli bug I witnessed in my local setup https://github.com/web3-storage/w3cli/issues/113

Context
* Why does this fix it?
  * previously, it was possible for calls to `Agent#registerSpace` to fail due to invalid proofs, even if the underlying AgentData had the right proof. This was because `registerSpace` ends up calling `invokeAndExecute(Provider.add)`, which calls `invoke` to build the invocation, which calls `proofs` to get the proofs for the invocation. And `proofs` could fail to include a sessionProof that was appropriate for the invocation audience of the invocation built by `Agent#invoke`. Now, `Agent#proofs` returns all the sessionProofs that might be relevant.
* explanation of underlying cause of that w3cli issue 113 https://github.com/web3-storage/w3up/pull/1047#discussion_r1377918605

### DIFF
--- a/packages/access-client/src/agent-data.js
+++ b/packages/access-client/src/agent-data.js
@@ -217,9 +217,6 @@ export function matchSessionProof(ucan, options) {
   const matchesRequiredProof =
     !options.attestedProof ||
     options.attestedProof.toString() === cap.nb.proof.toString()
-  if (!isSessionProof(ucan)) {
-    return false
-  }
   if (isExpiredButNotAllowed) {
     return false
   }
--- a/packages/access-client/src/agent-data.js
+++ b/packages/access-client/src/agent-data.js
@@ -194,36 +194,3 @@ export function getSessionProofs(data, issuer) {
   }
   return proofs
 }
-
-/**
- * Given a UCAN, return whether it is a session proof matching some options
- *
- * @param {Ucanto.Delegation} ucan
- * @param {object} options
- * @param {Ucanto.DID} [options.issuer] - issuer of session proof
- * @param {import('@ucanto/interface').UCANLink} [options.attestedProof] - CID of proof that the session proof attests to
- * @param {boolean} [options.allowExpired]
- * @returns {boolean} whether the ucan matches the options
- */
-export function matchSessionProof(ucan, options) {
-  if (!isSessionProof(ucan)) {
-    return false
-  }
-  const cap = ucan.capabilities[0]
-  const matchesRequiredIssuer =
-    options.issuer === undefined || options.issuer === ucan.issuer.did()
-  const isExpiredButNotAllowed = !options.allowExpired && isExpired(ucan)
-  const matchesRequiredProof =
-    !options.attestedProof ||
-    options.attestedProof.toString() === cap.nb.proof.toString()
-  if (isExpiredButNotAllowed) {
-    return false
-  }
-  if (!matchesRequiredIssuer) {
-    return false
-  }
-  if (!matchesRequiredProof) {
-    return false
-  }
-  return true
-}
--- a/packages/access-client/src/agent-data.js
+++ b/packages/access-client/src/agent-data.js
@@ -64,7 +64,7 @@ export class AgentData {
       dels.set(key, {
         delegation: importDAG(
           value.delegation.map((d) => ({
-            cid: CID.parse(d.cid),
+            cid: CID.parse(d.cid).toV1(),
             bytes: d.bytes,
           }))
         ),
@@ -168,13 +168,19 @@ export const isSessionProof = (delegation) =>
   delegation.capabilities.some((cap) => isSessionCapability(cap))
 
 /**
+ * @typedef {string} SessionProofAuthorizationCid - the nb.proof CID of the ucan/attest in the session proof
+ * @typedef {Ucanto.DID} SessionProofIssuer - issuer of ucan/attest session proof
+ * @typedef {Record<SessionProofAuthorizationCid, Record<SessionProofIssuer, [Ucanto.Delegation, ...Ucanto.Delegation[]]>>} SessionProofIndexedByAuthorizationAndIssuer
+ */
+
+/**
  * Get a map from CIDs to the session proofs that reference them
  *
  * @param {AgentData} data
- * @returns {Record<string, [Ucanto.Delegation, ...Ucanto.Delegation[]]>}
+ * @returns {SessionProofIndexedByAuthorizationAndIssuer}
  */
 export function getSessionProofs(data) {
-  /** @type {Record<string, [Ucanto.Delegation, ...Ucanto.Delegation[]]>} */
+  /** @type {SessionProofIndexedByAuthorizationAndIssuer} */
   const proofs = {}
   for (const { delegation } of data.delegations.values()) {
     if (isSessionProof(delegation)) {
@@ -182,8 +188,11 @@ export function getSessionProofs(data) {
       if (cap && !isExpired(delegation)) {
         const proof = cap.nb.proof
         if (proof) {
-          proofs[proof.toString()] = proofs[proof.toString()] ?? []
-          proofs[proof.toString()].push(delegation)
+          const proofCid = proof.toString()
+          const issuerDid = delegation.issuer.did()
+          proofs[proofCid] = proofs[proofCid] ?? {}
+          proofs[proofCid][issuerDid] = proofs[proofCid][issuerDid] ?? []
+          proofs[proofCid][issuerDid].push(delegation)
         }
       }
     }
--- a/packages/access-client/src/agent-data.js
+++ b/packages/access-client/src/agent-data.js
@@ -171,23 +171,19 @@ export const isSessionProof = (delegation) =>
  * Get a map from CIDs to the session proofs that reference them
  *
  * @param {AgentData} data
- * @param {Ucanto.DID|undefined} [issuer] - if provided, will only return session proofs issued by this id
- * @returns {Record<string, Ucanto.Delegation>}
+ * @returns {Record<string, [Ucanto.Delegation, ...Ucanto.Delegation[]]>}
  */
-export function getSessionProofs(data, issuer) {
-  /** @type {Record<string, Ucanto.Delegation>} */
+export function getSessionProofs(data) {
+  /** @type {Record<string, [Ucanto.Delegation, ...Ucanto.Delegation[]]>} */
   const proofs = {}
   for (const { delegation } of data.delegations.values()) {
-    if (issuer !== undefined && issuer !== delegation.issuer.did()) {
-      // delegation doesn't match issuer
-      continue
-    }
     if (isSessionProof(delegation)) {
       const cap = delegation.capabilities[0]
       if (cap && !isExpired(delegation)) {
         const proof = cap.nb.proof
         if (proof) {
-          proofs[proof.toString()] = delegation
+          proofs[proof.toString()] = proofs[proof.toString()] ?? []
+          proofs[proof.toString()].push(delegation)
         }
       }
     }
--- a/packages/access-client/src/agent-data.js
+++ b/packages/access-client/src/agent-data.js
@@ -171,7 +171,7 @@ export const isSessionProof = (delegation) =>
  * Get a map from CIDs to the session proofs that reference them
  *
  * @param {AgentData} data
- * @param {Ucanto.DID} [issuer] - if provided, will only return session proofs issued by this id
+ * @param {Ucanto.DID|undefined} [issuer] - if provided, will only return session proofs issued by this id
  * @returns {Record<string, Ucanto.Delegation>}
  */
 export function getSessionProofs(data, issuer) {
--- a/packages/access-client/src/agent-data.js
+++ b/packages/access-client/src/agent-data.js
@@ -189,3 +189,26 @@ export function getSessionProofs(data) {
   }
   return proofs
 }
+
+/**
+ * Given a UCAN, return whether it is a session proof matching some options
+ *
+ * @param {Ucanto.Delegation} ucan
+ * @param {object} options
+ * @param {Ucanto.DID} [options.issuer] - issuer of session proof
+ * @param {import('@ucanto/interface').UCANLink} [options.attestedProof] - CID of proof that the session proof attests to
+ * @param {boolean} [options.allowExpired]
+ * @returns {boolean} whether the ucan matches the options
+ */
+export function matchSessionProof(ucan, options) {
+  if ( ! isSessionProof(ucan)) { return false; }
+  const cap = ucan.capabilities[0]
+  const matchesRequiredIssuer = (options.issuer === undefined) || options.issuer === ucan.issuer.did()
+  const isExpiredButNotAllowed = ( ! options.allowExpired) && isExpired(ucan)
+  const matchesRequiredProof = ( ! options.attestedProof) || (options.attestedProof.toString() === cap.nb.proof.toString())
+  if ( ! isSessionProof(ucan)) { return false; }
+  if (isExpiredButNotAllowed) { return false; }
+  if ( ! matchesRequiredIssuer) { return false; }
+  if ( ! matchesRequiredProof) { return false; }
+  return true
+}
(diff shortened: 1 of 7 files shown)


packages/access-client/src/agent-data.js
```diff
@@ -189,3 +195,39 @@ export function getSessionProofs(data) {
...
+  }
+  const cap = ucan.capabilities[0]
+  const matchesRequiredIssuer =
+    options.issuer === undefined || options.issuer === ucan.issuer.did()
+  const isExpiredButNotAllowed = !options.allowExpired && isExpired(ucan)
+  const matchesRequiredProof =
+    !options.attestedProof ||
+    options.attestedProof.toString() === cap.nb.proof.toString()
+  if (!isSessionProof(ucan)) {
+    return false
```
alanshaw, line 221: This is repeated above...

packages/access-client/src/agent-data.js
```diff
@@ -189,3 +195,39 @@ export function getSessionProofs(data) {
...
+ * Given a UCAN, return whether it is a session proof matching some options
+ *
+ * @param {Ucanto.Delegation} ucan
+ * @param {object} options
+ * @param {Ucanto.DID} [options.issuer] - issuer of session proof
+ * @param {import('@ucanto/interface').UCANLink} [options.attestedProof] - CID of proof that the session proof attests to
+ * @param {boolean} [options.allowExpired]
+ * @returns {boolean} whether the ucan matches the options
+ */
+export function matchSessionProof(ucan, options) {
```
alanshaw, line 209: Where is this used?

packages/access-client/src/agent-data.js
```diff
@@ -171,12 +171,18 @@ export const isSessionProof = (delegation) =>
...
  * @returns {Record<string, Ucanto.Delegation>}
  */
-export function getSessionProofs(data) {
+export function getSessionProofs(data, issuer) {
   /** @type {Record<string, Ucanto.Delegation>} */
   const proofs = {}
   for (const { delegation } of data.delegations.values()) {
+    if (issuer !== undefined && issuer !== delegation.issuer.did()) {
+      // delegation doesn't match issuer
+      // eslint-disable-next-line no-continue
```
alanshaw, line 183: add to config in package.json?

packages/access-client/src/agent.js
```diff
@@ -175,21 +176,24 @@ export class Agent {
...
+      if (isTooEarly(value.delegation)) {
+        continue
+      }
+
+      if (Array.isArray(caps) && caps.length > 0) {
+        // caps param is provided. Ensure that the delegations we're looping over can delegate to these caps
+        for (const cap of _caps) {
+          if (canDelegateCapability(value.delegation, cap)) {
+            values.push(value)
+            _caps.delete(cap)
```
alanshaw: Can we make it so that this includes all delegations that can delegate this capability, not just the first?

From your investigation, this seemed to be the issue right?

alanshaw (commented):
I don't really understand how this fixes the issue.

packages/access-client/src/agent-data.js
gobengo: Originally I used this in the first thing I tried to get things to work, which was adding some session proof awareness to the agent's `.#delegations()` method (called by `.proofs()` method). Then I realized that was going to be a bit redundant because the `.proofs()` method impl calls `getSessionProofs`, so I refactored a bit to add the issuer-checking logic to `getSesssionProofs`.

This feels like a function we'll eventually want around for some advanced proof-finding scenarios, but if if I don't need it to get the tests passing, I can remove it and keep it in my notes

packages/access-client/src/agent-data.js
gobengo: good idea! did it in commit 7248d915fa53d7389f09ae4504b39122bf589dee

packages/access-client/src/agent-data.js
```diff
@@ -171,12 +171,17 @@ export const isSessionProof = (delegation) =>
  * Get a map from CIDs to the session proofs that reference them
  *
  * @param {AgentData} data
+ * @param {Ucanto.DID} [issuer] - if provided, will only return session proofs issued by this id
  * @returns {Record<string, Ucanto.Delegation>}
  */
-export function getSessionProofs(data) {
+export function getSessionProofs(data, issuer) {
```
gobengo, line 177: Part of what was causing the bug is that this function ends up building a map of CIDs to proofs. The CID it uses is the nb.proof from a ucan/attest. The agent might actually have multiple ucan/attest ucans with the same nb.proof CID, but with different issuers. Previously only one of these delegations would end up in the returned map from this function, and the caller had no way of getting a session proof appropriate for the audience of their request.

packages/access-client/src/agent.js
gobengo: First, fyi I just now undid my changes to this `#delegations()` fn and reverted it back to how it was. While I was debugging this yesterday and trying to recreate as a test case and then try ways of making it pass, I had originally tried adding some smarts to this function to make it `invocationAudience` and session proof aware. But then I realized that would be doing some work twice given that the whole way the `proofs()` method works is calling `this.#delegations` but then also tacking on sessionProofs that are relevant to the result. So then I moved the invocation audience aware logic out of here and a layer higher into the `proofs()` method and its usage of `getSessionProofs`.

> Can we make it so that this includes all delegations that can delegate this capability, not just the first?

I'm not against that, but it wasn't required to fix the bug that I was seeing in my setup and in the test, and this is the logic that's been here for awhile, so I'm hesitant to change it if I don't have a test case that illustrates why.

> From your investigation, this seemed to be the issue right?

Yesterday I realized that in the situation I personally ran into in [here](https://github.com/web3-storage/w3cli/issues/113) and then I think captured in test case, that wasn't actually the underlying issue.

The underlying issue was more directly that in the caller of this function, `Agent#proofs`, even if the right proof was returned, it would then use `getSessionProofs` to add a relevant session proof to the invocation, but that could return a session proof that had an issuer that was unknown to the invocation target. So then when the invocation target (e.g. `did:web:staging.web3.storage`) got a session proof issued by some unknown id (e.g. `did:web:web3.storage`), it would lead to that [error](https://github.com/web3-storage/w3cli/issues/113)

So once I realized that was the root cause of the issue for w3cli 113, I could fix things not by touching this function and instead by making `proofs()` method have an optional `invocationTarget` param and then passing that to `getSessionProofs` so that getSessionProofs could return the appropriate session proof in this weird case where there are multiple session proof issuers (which is unusual for end-user, but not unusual for a w3infra dev)

packages/access-client/src/agent.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-continue */
```
travis, line 1: is this needed here? I see the continue in `agent-data.js` and the global disabling of `no-continue` but don't see a continue in this file...

packages/access-client/src/agent.js
```diff
@@ -254,17 +255,17 @@ export class Agent {
    * proofs matching the passed capabilities require it.
    *
    * @param {import('@ucanto/interface').Capability[]} [caps] - Capabilities to filter by. Empty or undefined caps with return all the proofs.
+   * @param {Ucanto.DID} [invocationAudience] - audience of invocation these proofs will be bundled with.
    */
-  proofs(caps) {
+  proofs(caps, invocationAudience) {
```
travis, line 260: this is a bit confusing to me - we might be getting proofs for a delegation that we want to issue to someone else, in which case there won't be an invocation at all - I'd at least expect this to be an optional parameter, but at a higher level it just feels confusing to be talking about invocations here at all...

travis (changes requested):
One bit of cleanup, but I do think the `proofs` method signature needs some thought - happy to chat in higher bandwidth whenever's convenient for you!

packages/access-client/src/agent.js
gobengo: good catch. I removed it now that the rule is not enabled more globally

packages/access-client/src/agent.js
gobengo: We spoke after this comment. I mentioned I had been trying to write this as an optional parameter. I've since made commits to make it more obvious that undefined is allowed, and based on @travis suggestion I also changed this second positional param to be an options object with options.sessionProofIssuer filter that can be used when `Agent.invoke` calls `Agent.proofs`

packages/access-client/src/agent-data.js
gobengo: Thanks. I removed the repeat.

packages/access-client/src/agent-data.js
gobengo: I removed it

gobengo:
I met with @Gozala about this and he agreed it makes sense. In the medium term we both don't love the `Agent#proofs` method API as is, and think maybe `Agent#invoke` should use something else to find the proofs for an invocation, but adding this option is a good stopgap to fix the bug in a backward-compatible way.

travis (approved):
looks good, thanks!

packages/upload-api/src/access/confirm.js
```diff
@@ -111,6 +111,11 @@ export async function createSessionProofs({
     capabilities,
     expiration,
     proofs: delegationProofs,
+    facts: [
+      {
+        service: service.did(),
+      }
+    ]
```
Gozala, line 118: I realized after we spoke that it would be a better idea to include a link to the invocation that caused it specifically `access/confirm` capability we received here, which is created at 

https://github.com/web3-storage/w3up/blob/b232aa784c178cc6877982d7d9dfe4da06a06203/packages/upload-api/src/access/authorize.js#L51-L71

In other words I think it will be something like


```suggestion
    facts: [{ cause: invocation.cid }]
```

It also would make sense to add a link to `access/authorize` to a `access/confirm` we create, but that could also be a separate pr.

packages/access-client/src/agent-data.js
Gozala: I think this is fairly low level API so I suggest a breaking change that will change return type to 

```js
Record<Ucanto.DID, Record<string, ...Ucanto.Delegation>>
```

That way things will work correctly even if you do not pass the issuer. Furthermore I do have use case where I query `proofs` and I do not know the issuer of attestations so getting all the proofs would be desired there.

packages/upload-api/src/access/confirm.js
gobengo: I added a fact like the one you suggest

packages/access-client/src/agent.js
gobengo: I ended up finding a test that failed unless I removed the `caps.delete()` so i removed it https://github.com/web3-storage/w3up/compare/cb90340...79a544dda9c382a73b7dd6383729c9b09b72834b#diff-d14594f7f9441ea59499c7e8f4d6ebb32ccb406de725d9c89cb3e53f1d63e37cR394

packages/access-client/src/agent-data.js
gobengo: I did something like that here https://github.com/web3-storage/w3up/pull/1047/commits/a9aedf67d51b66aeb9a3653237565f67565ade61

packages/access-client/src/agent.js
gobengo: Later I ended up just removing this new param alltogether. Now there is no API change that allows for filtering the proofs from the `proofs` method to only get proofs related to building a certain invocation. The tradeoff is that `proofs()` may return more than is strictly needed, but @travis explained to me in 1:1 that that is a good way to err for now (and as long as we have tests for tricky scenarios, we can always refactor to optimize invocation/proofs size later without regressions)

packages/upload-api/src/access/confirm.js
```diff
@@ -68,6 +68,12 @@ export async function confirm({ capability, invocation }, ctx) {
     // implement sudo access anyway.
     delegationProofs: delegationsResult.ok,
     expiration: Infinity,
+    attestationFacts: [
+      {
+        // include this so that the attestation CID is unique across confirmations
+        cause: invocation.cid,
```
gobengo, line 74: @Gozala encouraged me to add this to affect future session proofs

packages/upload-api/src/access/confirm.js
```diff
@@ -68,6 +68,12 @@ export async function confirm({ capability, invocation }, ctx) {
...
+      }
+    ]
```
Gozala, line 76: Can you this just be a top level please ? And you could also include it in both delegations created

```suggestion
    // include this so that the attestation CID is unique across confirmations
    cause: invocation.cid,
```

packages/access-client/src/agent-data.js
```diff
@@ -171,18 +171,19 @@ export const isSessionProof = (delegation) =>
  * Get a map from CIDs to the session proofs that reference them
  *
  * @param {AgentData} data
- * @returns {Record<string, Ucanto.Delegation>}
+ * @returns {Record<string, [Ucanto.Delegation, ...Ucanto.Delegation[]]>}
  */
 export function getSessionProofs(data) {
-  /** @type {Record<string, Ucanto.Delegation>} */
+  /** @type {Record<string, [Ucanto.Delegation, ...Ucanto.Delegation[]]>} */
```
Gozala, line 177: This is ok but I would prefer `Record<Ucanto.DID, Record<string, ...Ucanto.Delegation>>` per https://github.com/web3-storage/w3up/pull/1047#discussion_r1378213049 because it makes it a lot easier to pick the ones you care and throw away others as opposed to having to go and group them in another loop.

Not a big deal, but unless there is compelling reason to prefer this I would suggest reverse.

packages/upload-api/src/access/confirm.js
Gozala: For the context we have another issue with the authorization (that this would allow us to resolve) where authorization succeeds before user clicks the email, that is because following function

https://github.com/web3-storage/w3up/blob/b232aa784c178cc6877982d7d9dfe4da06a06203/packages/access-client/src/agent-use-cases.js#L208C23-L213

uses not so great heuristics to determine when authorization has been completed

https://github.com/web3-storage/w3up/blob/b232aa784c178cc6877982d7d9dfe4da06a06203/packages/access-client/src/agent-use-cases.js#L97-L99


We could try to improve heuristics, but ideally we would have a way of just seeing if received delegation corresponds to the authorization invocation, which is why I'd like to capture these details somewhere.

Gozala (approved):
Approved with some suggestions

packages/access-client/src/agent-data.js
gobengo: @Gozala roger. I will change to that.

I assume the keys of the inner record (typed as `string`) would be session proof issuer? and keys of outer record would be the attested to proof cid?

packages/upload-api/src/access/confirm.js
gobengo: > ideally we would have a way of just seeing if received delegation corresponds to the authorization invocation,

huge +1. I was suggesting something similar to @travis the other day, but I didn't think to just use the CIDs of the precursor invocations as facts. love that idea

packages/access-client/src/agent-data.js
gobengo: did it https://github.com/web3-storage/w3up/pull/1047/commits/914405d21f57cb7c0969c767930e47c0dfa06e46

gobengo:
I was about to merge this, then did one last bit of testing in my local setup to make sure all this would still fix the situation for #113, but noticed that when I tried, I got a new 500 error from the backend. Two days ago I had tested and it worked. I think the regression happened because yesterday I took out the `Agent#proofs` `option.sessionProofIssuer` filter. This led to a simpler API, but at the cost of sometimes sending more sessionProofs than are needed in the invocation. I figured it would be harmless. But actually it was not harmless. Sending all those session proofs incl from unknown (to that service) issuers resulted in the upload-api lambda timing out for 10 seconds. I looked and was unable to find logs/sentry info that would hint at why, but my guess is that it was in the proof checking.

So I re-introduced `options.sessionProofIssuer` to `Agent#proofs` and re-added using it when `Agent#invoke` calls `Agent#proofs` to get proofs for the invocation. Once I made this change so ensure that `Agent#invoke` invocation doesn't include proof for sessions issued by issuers other than the invocation audience, then things work again.

PR #1052, opened 2023-10-31
feat!: add storefront filecoin api to upload api

vasco-santos:
Adds `filecoin/*` to the `upload-api`, so that web3.storage implements storefront capabilities

This will enable us to integrate the new API with w3infra, so that clients can execute `filecoin/offer` to web3.storage.

### DIFF
--- a/.github/workflows/upload-api.yml
+++ b/.github/workflows/upload-api.yml
@@ -41,6 +41,11 @@ jobs:
       - name: Prepare
         run: pnpm install
 
+      - name: Build
+        run: |
+          pnpm install
+          pnpm run --if-present build
+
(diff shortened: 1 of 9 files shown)


.github/workflows/upload-api.yml
```diff
@@ -41,6 +41,11 @@ jobs:
       - name: Prepare
         run: pnpm install
 
+      - name: Build
+        run: |
+          pnpm install
+          pnpm run --if-present build
+
```
vasco-santos, line 48: `filecoin-client` needs to be built :)

alanshaw (approved):

PR #1058, opened 2023-11-01
chore: no longer depends on hd-scripts, packages use/configure eslint directly, fixes warnings from npm lint script

gobengo:
Motivation:
* fix warning in console when running lint script while devving in access-client: https://github.com/web3-storage/w3up/issues/1053

Tactics:
* new package `packages/eslint-config-w3up`
* all packages use package.json to set `eslintConfig.extends` to `@web3-storage/eslint-config-w3up` and have a devDep on that package within the monorepo. No monorepo subpackages use any eslint rules other than those in eslint-config-w3up. i.e. all w3up packages use the same eslint config (this wasn't actually the case before, since different w3up packages were depending on different versions of hd-scripts)
* no packages depend on hd-scripts any more


packages/access-client/package.json
```diff
@@ -95,18 +99,32 @@
...
     "rules": {
-      "unicorn/prefer-number-properties": "off",
-      "unicorn/prefer-export-from": "off",
-      "unicorn/expiring-todo-comments": "off",
-      "unicorn/no-array-reduce": "off",
-      "unicorn/explicit-length-check": "off",
+      "@typescript-eslint/no-unused-vars": "off",
+      "@typescript-eslint/ban-ts-comment": "off",
       "no-continue": "off",
+      "jsdoc/check-values": "off",
```
gobengo, line 113: I copied these rules over from hd-scripts so this PR doesn't have to make lots of changes to the src, but over time I'd prefer to just use the default "plugin:jsdoc/recommended" rules.

travis (approved):
it looks like there are a couple typechecking warnings in packages/w3up-client/test/helpers/car.js that are probably worth addressing (see the bottom of the "Files changed" tab on this PR) but otherwise this looks good to go!

Issue #945, opened 2023-10-05
Client should setup space recovery when it creates one


Gozala:
Right now our clients create a space and delegate all capabilities to the agent, which later delegates them to the account.

This has two the problem that access to space can be permanently lost if agent gets compromised.

What we should do instead is on space creation setup a recovery plan. I think we can offer them to provide an account email so that it has full privileges for the account.

We can also follow the established cow path in the space and also ask user to write down set of phrases from which space key can be derived.

By doing both we will ensure that user will never loose access to their space.

Gozala:
So here is how things work today

```
w3 space create
did:key:z6MkqneEnjm67escdX9mXRniyh3habma5CvDMNRXpz78wvKy

w3 space register gozala@...
⁂ space registered to gozala@...

w3 up /Users/gozala/Downloads/w3-aggregation.svg
  1 file 0.2MB
⁂ Stored 1 file
⁂ https://w3s.link/ipfs/bafybeig6klboqjlqq55hbtx3xvsqqjvnopcluwj6jhn5ctqtgcet45qwtu
```

Here is how I suggest they work in the future

```
w3 space create

did:key:z6MkqneEnjm67escdX9mXRniyh3habma5CvDMNRXpz78wvKy

Here is your unique paper key, which will allow you to recover access to the created space on a different device in case access to all other devices are compromised. This is the only time you'll see this so be sure to write it down and then hit enter

seven physical orange dash island village scatter together recycle echo ladder cash response become aim desert desk advice axis gown unfold reduce plate rifle

Just to be sure you got it right, please enter or paste the paper you have saved and then hit enter

We recommend delegating full access to this space to your email address, which will allow you to gain access to the space using simple email authorization and without a paper key. If you exit now recovery email will not be setup and you will need to use paper key to authorize on a new device. Please enter email address you wish to authorize, or hit enter to use gozala@....com

New space is created but does not yet have any storage capacity, this can be arranged by running `w3 space provision` command.
...

**Comment by alanshaw** [comment]
LGTM, although as discussed before please ensure parameters exist for non-interactive use cases.

How do we use paper key to authorize on a new device?

**Comment by Gozala** [comment]
> How do we use paper key to authorize on a new device?

```sh
space recover did:key:zMySpace
⁂ Please enter or paste the paper key displayed during space creation and press Enter key
ℹ️ We will not store this key, but we will use it to delegate access to w3 on this device

# User pastes the key

⁂ Space has been added locally
```

[...]

PR #1059, opened 2023-11-03
feat!: authorization and registration flows

Gozala:
Implements flows described in #945

### Accounts

This pull request starts differentiating between two kinds of authorizations

1. Space authorization, allowing access to a space.
2. Account authorization, allowing account management and space provisioning.

Most core functionality has been added in `Account` module which exposes following functionality

- `Account.login(client, email)` - Request account level access though `access/authorize` and awaits until access is granted (through email link) or request expires.

   It returns an `Account` object which is just a view over UCAN delegations and an agent. That is convenience wrapper allowing us to add e.g. `.provision(space)` method (for provisioning spaces) and `.save()` method that will save account proofs into agent store.

- `Account.list()` - Lists all account (sessions) client has in storage. It essentially goes over proofs and wraps ones that are "account authorization"s into `Account` objects.

Above functionality should help us answer question whether user has already logged-in or not ? Previously we had no good way to answer such questions.

### Access

This fixes issues around authorization flows, where requests would succeed but access was not gained. Issue was caused by imperfect heuristics that tried to match received delegation with a requested authorization.

To address this problem `access/authorize` and `access/confirm` capabilities and their handler had been extended to capture request CIDs. This allows client polling for delegations to detect whether received delegation is corresponds to the request more precisely.

I have added new `Access` module that attempts to do the client side work without much configurability. 

- `Access.request` - will send `access/authorize` request and return `PendingAccessRequest` object, which in turn could be used to poll for the response via `.claim()` method. `.claim()` internally uses `.poll()` until it gets the delegation, session expires or request is aborted.

- `Access.claim` - sends `access/claim` request and returns `GrantedAccess` object that wraps received delegations. It also (just like `Account`) has `.save()` method which can be used to save delegations into client's store.

### Space

I have added new `Space` module that has has `Space.generate()` function to generate new spaces, `Space.fromDelegation`, `Space.fromMnemonic` and more. General idea here also is that `client.createSpace()` will give you `OwnedSpace` object that has bunch of methods like `.toMnemonic`, `.createAuthorization`, `.createRecovery` etc...

Notion of not registered spaces had been removed. If you create a space you have an object, you can save it in your client using `client.addSpace(await space. createAuthorization(agent))`.

### Attestation

I have moved `ucan/attest` capability definition from `access` to `ucan` as former place was confusing and made no sense.

---

The [account.test.js](https://github.com/web3-storage/w3up/pull/1059/files#diff-9a9133d8e343b56d7ef7bd18d01356b8ca7c630143b4687a51f683a5ebe1528f) attempts to capture authorization / provisioning flows from the event diagram.

There is one unfortunate problem right now, that is new spaces added will not show in other devices unless you re-authorize. Fixing that requires https://github.com/web3-storage/ucanto/pull/326 and some more work here. I think it's best to do it as a followup.

---

I had to rebase because things upstream have changed, unfortunately I failed to do it cleanly so there is lot of strange changes which I don't exactly know how they happened. I'll try to remove undo them tomorrow


packages/access-client/src/agent.js
```diff
@@ -425,21 +452,31 @@ export class Agent {
...
+  //  * @param {string} email
+  //  * @param {object} [opts]
+  //  * @param {AbortSignal} [opts.signal]
+  //  * @param {Ucanto.DID<'key'>} [opts.space] - space to register
+  //  * @param {Ucanto.DID<'web'>} [opts.provider] - provider to register - defaults to this.connection.id
+  //  */
+  // async registerSpace(email, opts = {}) {
+  //   return await addProviderAndDelegateToAccount(this, this.#data, email, opts)
+  // }
+
```
Gozala, line 479:
```suggestion
```

packages/access-client/src/agent.js
```diff
@@ -317,65 +333,76 @@ export class Agent {
...
+  //   await this.#data.addSpace(signer.did(), meta, proof)
+
+  //   return {
+  //     did: signer.did(),
+  //     meta,
+  //     proof,
+  //   }
+  // }
+
```
Gozala, line 368:
```suggestion
```

packages/upload-api/test/storage/car-store-bucket.js
Gozala, line 1: Had to change so it could be run in browsers

packages/access-client/src/access.js
```diff
@@ -0,0 +1,298 @@
...
+
+/**
+ * Checks if the given delegation is caused by the passed `request` for access.
+ *
+ * @param {API.Delegation} delegation
+ * @param {object} selector
+ * @param {API.Link} selector.request
+ * @returns
+ */
+
```
gobengo, line 260: extra newline

packages/access-client/src/agent.js
```diff
@@ -153,25 +152,39 @@ export class Agent {
...
+    return await this.addProofs([delegation])
+  }
+
+  /**
+   * Adds set of proofs to the agent store. Throws if the proofs are invalid.
+   * does not match the agent DID.
+   *
+   * @param {Iterable<API.Delegation>} delegations
+   */
+  async addProofs(delegations) {
```
gobengo, line 167: Can `validate` throw? If so, what happens if I pass 100 delegations, 49 are valid and then the 50th isn't, and I pass em all here?

My understanding is not only might the promise returned by this function throw (good), but when it does so, it would have already added the first 49 valid proofs to `this.#data`. i.e. there was basically partial application of `addProofs(delegations)` but without the caller knowing which part worked and which didn't.

Maybe it would be better to validate the whole batch at once in parallel before doing any `addDelegation`?

```
await Promise.all(delegations.map(proof => validate(proof, /*...*/)))
for (const proof of delegations) { this.#data.addDelegation(proof, { audience: this.meta })) }
```

packages/access-client/src/space.js
```diff
@@ -0,0 +1,236 @@
...
+  return new OwnedSpace({ signer, name })
+}
+
+/**
+ * @param {object} space
+ * @param {ED25519.EdSigner} space.signer
+ */
+export const toMnemonic = ({ signer }) => {
+  /** @type {Uint8Array} */
+  // @ts-expect-error - Field is defined but not in the interface
```
gobengo, line 45: that's actually a little terrifying :P
usually I'd pass a `signer` instead of a `secret` and expect the `signer` is hiding any secrets and only eposing the ability to sign. good to realize that `signer` is not actually hiding that prop

packages/access-client/src/space.js
```diff
@@ -0,0 +1,236 @@
...
+    facts: [{ space: { name } }],
+  })
+}
+
+// Default authorization session is valid for 1 year
+export const SESSION_LIFETIME = 60 * 60 * 24 * 365
+/**
+ * Creates (UCAN) delegation that gives specified `agent` an access to
+ * specified ability (passed as `access.can` field) on the this space.
+ * Optionally, you can specify `access.expiration` field to set the
```
gobengo, line 74: "to set the..."

packages/access-client/test/agent.test.js
```diff
@@ -23,31 +23,45 @@ describe('Agent', function () {
...
     const delegations = agent.proofs()
 
-    assert.equal(space.proof.cid, delegations[0].cid)
+    assert.equal(authorization.cid, delegations[0].cid)
   })
 
   it('should set current space', async function () {
     const agent = await Agent.create()
     const space = await agent.createSpace('test')
+    const authorization = await space.createAuthorization(agent, {
```
gobengo, line 46: I wouldn't mind if we left it so that `Agent#createSpace` automatically sets up authz, importing the space, and setting current, which would be a lot more back compat with what we've had. Not sure what's best here, but I'm worried that changing the behavior of `createSpace` too much might lead to headaches for end-users. I'd be inclined to try to keep all the current side effects for the same method (if easy), and only introduce new APIs/sideEffects in new methods that end-users can opt into over time.

packages/capabilities/src/access.js
```diff
@@ -83,6 +83,11 @@ export const confirm = capability({
   can: 'access/confirm',
   with: DID,
   nb: Schema.struct({
+    /**
+     * Link to the `access/authorize` request that this delegation was created
+     * for.
+     */
+    cause: Schema.link({ version: 1 }),
```
gobengo, line 90: should it be optional?

there are many confirms in the db already that may not parse unless its optional?

packages/upload-api/src/access/authorize.js
```diff
@@ -81,7 +89,20 @@ export const authorize = async ({ capability }, ctx) => {
     url,
   })
 
-  return {
-    ok: {},
+  const ok = Server.ok({
+    // let client know when the confirmation will expire
+    expiration: confirmation.expiration * 1000,
```
gobengo, line 94: can we make this an ISO8601 datetime string of exactly when expiration will happen instead of a relative expiry length.?

packages/access-client/src/access.js
```diff
@@ -0,0 +1,298 @@
...
+ * Represents a pending access request. It can be used to poll for the requested
+ * delegation.
+ */
+class PendingAccessRequest {
+  /**
+   * @param {object} source
+   * @param {API.Agent} source.agent
+   * @param {API.DID} source.audience
+   * @param {API.ProviderDID} source.provider
+   * @param {number} source.expiration
```
gobengo, line 133: what kind of `number`? e.g. is it number of second or milliseconds since epoch? something else?
(should it be a `Date`?)

packages/upload-api/src/access/authorize.js
gobengo: I think right now this `expiration` is number of milliseconds since epoch. would much prefer something more standard

packages/upload-api/test/access-client-agent.js
```diff
@@ -154,31 +161,36 @@ export const test = {
...
-    const spaceName = `space-test-${Math.random().toString().slice(2)}`
-    const spaceCreation = await agent.createSpace(spaceName)
-    await agent.setCurrentSpace(spaceCreation.did)
-
-    // 'register space' - i.e. add a storage provider as an account
-    await agent.registerSpace(accountEmail, {
-      provider: /** @type {API.DID<'web'>} */ (agent.connection.id.did()),
-    })
-  },
+  // 'can registerSpace': async (assert, context) => {
```
gobengo, line 164: @Gozala these are still commented

packages/w3up-client/test/result.test.js
```diff
@@ -0,0 +1,24 @@
+import * as Result from '../src/result.js'
+import assert from 'assert'
+
+describe('Result', () => {
+  it('expect throws on error', async () => {
+    assert.throws(() => Result.expect({ error: new Error('Boom') }), /Boom/)
```
gobengo, line 6: I had to go read the src when i saw this in use before to figure out what it's doing.

Maybe `Result.ok` or `Result.assert` would be more readable to me.

`Result.expect` hinted you were doing some kind of expectation about the result, but didn't tell me much about exactly what was being expected.

kinda like how some assertion libraries do `expect(target).toBeTruthy` not just `expect(target)`

gobengo (changes requested):
Looks good to me so far. I left a few comments above.

It's definitely worth changing at least the part I commented on where there are still some `registerSpace` stuff commented out.

packages/access-client/src/agent.js
Gozala: I honestly don't see value in doing validation here in first place. Even if the proofs are still valid they might expire in a minute. Furthermore `this.removeExpiredDelegations()` also goes and remove expired proofs.

I suppose we will throw errors if audience isn't the agent, but then again not sure why do we care about it.

packages/access-client/src/space.js
Gozala: Secret is just byte range within the signer byte array it's something like `signer.subarray(3, 3 + 32)` if I'm not wrong

Gozala: See https://github.com/web3-storage/ucanto/blob/43ea497ca2200f9e1e15dbbdd033ec24895edb7d/packages/principal/src/ed25519/signer.js#L155-L164

packages/access-client/test/agent.test.js
Gozala: I think that was a really bad design choice because it effectively creates a space that is not usable and throws away the key so it never can be made usable. New API intentionally does not save anything to the agent so you have to arrange recovery beforehand.

I think doing breaking changes before we actually launch is probably better than shipping a footgun for the users, especially given that only users right now are us

image

And jchris who's been running into this problem and keeps asking us to fix it.

---

If you insist I will create another method and leave `createSpace` alone, but I think it would be a mistake given that our API is not yet stable, has issues and it's essentially use fixing our apps.

packages/capabilities/src/access.js
Gozala: I don't think we should. We don't actually store confirms in DB just send them to user email and they expire in 15 minutes. So practically only way for someone to run into the problem would be if they tried to authorize just before we deploy this and then they click the link after we deployed it.

Given such a low probability and a low impact on not yet officially launched product, I don't think carrying the burden is worth it. If we were in the production I would have had a different opinion.

packages/upload-api/src/access/authorize.js
Gozala: Can you give some rational why choosing datetime string would be a better choice ? I chose to use same name and format as UCAN (and ucan receipts) and would find having it in different format in different places inconsistent and more strange.

That said `* 1000` was a mistake and I should really do that and turn it into `Date` instance in the `PendingRequest` just like you've suggested.

packages/upload-api/test/access-client-agent.js
```diff
@@ -154,31 +161,36 @@ export const test = {
...
+  //     expiration: Infinity,
+  //   })
+  //   await agent.importSpaceFromDelegation(spaceAuth)
+  //   await agent.setCurrentSpace(spaceCreation.did())
+
+  //   // 'register space' - i.e. add a storage provider as an account
+  //   await agent.registerSpace(accountEmail, {
+  //     provider: /** @type {API.DID<'web'>} */ (agent.connection.id.did()),
+  //   })
+  // },
```
Gozala, line 193:
```suggestion
```

packages/w3up-client/src/client.js
```diff
@@ -143,30 +141,29 @@ export class Client extends Base {
...
+  //  * @param {object} [options]
+  //  * @param {import('./types.js').DID<'web'>} [options.provider]
+  //  * @param {AbortSignal} [options.signal]
+  //  */
+  // async registerSpace(email, options = {}) {
+  //   options.provider =
+  //     options.provider ??
+  //     /** @type {import('./types.js').DID<'web'>} */ (this.defaultProvider())
+  //   await this._agent.registerSpace(email, options)
+  // }
```
Gozala, line 166:
```suggestion
```

packages/w3up-client/test/result.test.js
Gozala: Sorry I should have added more comments it's basically [Result.expect from Rust](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect).

[`Result.unwrap` is another](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap) if you like that name better, but it does not take second param to throw.

packages/access-client/src/agent.js
Gozala: I have decided to remove the validation step for the following reasons

1. Time bound checks is redundant because we remove expired delegations right after we end the loop.
2. Audience check is not a good idea e.g. because we would would also want to store delegations delegated to the accounts (this agent is logged-in with). Furthermore `agent.proofs` also filters out delegations where agent is not an audience, which suggests that there is in fact assumption that audience may not be this agent.
3. If we rotate agent key we may also end up with delegations in the store where audience is not the agent, so it's best to not to have code assume that all proofs are delegated to the agent.

packages/upload-api/src/access/authorize.js
gobengo: > I chose to use same name and format as UCAN (and ucan receipts) and would find having it in different format in different places inconsistent and more strange.

that's a good reason to do it that way then, consistent with other stuff.

packages/access-client/src/access.js
```diff
@@ -0,0 +1,335 @@
...
+
+  get provider() {
+    return this.model.provider
+  }
+
+  /**
+   *
+   * @returns {Promise<API.Result<API.Delegation[], API.InvocationError|API.AccessClaimFailure|RequestExpired>>}
+   */
+  async poll() {
```
alanshaw, line 166: How should this class be used? Should I call `poll` or `claim` on instances I receive? The docs above say I can use this to poll for the requested delegation but I think I would need to call `claim` not `poll` for that behaviour?

It seems back to front to me that `poll` sends a single request, but `claim` sends multiple requests. Perhaps `poll` should be a private method?

packages/access-client/src/access.js
```diff
@@ -0,0 +1,335 @@
...
+  for (const [can, details] of entries) {
+    if (details) {
+      abilities.push({ can })
+    }
+  }
+  return abilities
+}
+
+/**
+ * Set of capabilities required for by the agent to manage a space.
```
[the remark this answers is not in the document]
alanshaw: > to manage a space.

_All_ spaces for an account?

packages/access-client/src/agent.js
```diff
@@ -317,65 +318,49 @@ export class Agent {
...
+  async importSpaceFromDelegation(delegation, { name = '' } = {}) {
+    const space =
+      name === ''
+        ? Space.fromDelegation(delegation)
+        : Space.fromDelegation(delegation).withName(name)
 
-    this.#data.spaces.set(did, meta)
+    if (space.name === '') {
+      throw new Error(
+        'Space has no name please pass `option.name` to specify it'
```
alanshaw, line 342:
```suggestion
        'Space has no name, please pass a `name` option to specify it'
```

packages/access-client/src/space.js
```diff
@@ -0,0 +1,264 @@
...
+    access: Access.accountAccess,
+    expiration: Infinity,
+  })
+
+// Default authorization session is valid for 1 year
+export const SESSION_LIFETIME = 60 * 60 * 24 * 365
+
+/**
+ * Creates (UCAN) delegation that gives specified `agent` an access to
+ * specified ability (passed as `access.can` field) on the this space.
```
alanshaw, line 76:
```suggestion
 * specified ability (passed as `access.can` field) on this space.
```

alanshaw (approved):
❤️ This is great, it fits really nicely and seems pretty easy to understand. I've left some non-blocking suggestions.

We need a docs update for `w3up-client` but we can handle separately...

packages/access-client/src/access.js
```diff
@@ -0,0 +1,335 @@
...
+  get request() {
+    return this.model.request
+  }
+
+  get provider() {
+    return this.model.provider
+  }
+
+  /**
+   *
```
Gozala, line 163:
```suggestion
   * Low level method and most likely you want to use `.claim` instead. This method will poll 
   * fetch delegations **just once** and will return proofs matching to this request. Please note
   * that there may not be any matches in which case result will be `{ ok: [] }`.
   *
   * If you do want to continuously poll until request is approved or expired, you should use
   * `.claim` method instead.
   *
```

packages/access-client/src/access.js
```diff
@@ -0,0 +1,335 @@
...
+        : {
+            ok: result.ok.proofs.filter((proof) =>
+              isRequestedAccess(proof, this.model)
+            ),
+          }
+    }
+  }
+
+  /**
+   * @param {object} options
```
Gozala, line 184:
```suggestion
  /**
   * Continuously polls delegations until this request is approved or expired. Returns
   * a `GrantedAccess` object (view over the delegations) that can be used in the
   * invocations or can be saved in the agent (store) using `.save()` method.
   *
   * @param {object} options
```

packages/access-client/src/access.js
```diff
@@ -0,0 +1,335 @@
...
+ * @returns
+ */
+const isRequestedAccess = (delegation, { request }) =>
+  // `access/confirm` handler adds facts to the delegation issued by the account
+  // so that principal requesting access can identify correct delegation when
+  // access is granted.
+  delegation.facts.some((fact) => `${fact['access/request']}` === `${request}`)
+
+/**
+ * Maps access object that follows new UCAN spec inspired by recap style layout
```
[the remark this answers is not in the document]
Gozala: I think this is it https://github.com/spruceid/EIPs/blob/master/EIPS/eip-5573.md, but it's not worth mentioning a recap at all, it's more of the latest UCAN specs https://github.com/ucan-wg/spec#251-examples

packages/access-client/src/access.js
```diff
@@ -0,0 +1,335 @@
...
+ * into legacy UCAN 0.9 format used by various w3up capabilities that predate
+ * the new format.
```
Gozala, line 300:
```suggestion
/**
 * Maps access object that uses UCAN 0.10 capabilities format as opposed
 * to legacy UCAN 0.9 format used by w3up  which predates new format.
```

packages/w3up-client/src/account.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+}
+
+class Account {
+  /**
+   * @typedef {object} AccountModel
+   * @property {API.DidMailto} source.id
+   * @property {API.Agent} source.agent
+   * @property {API.Delegation[]} source.proofs
```
Gozala, line 106:
```suggestion
   * @property {API.DidMailto} id
   * @property {API.Agent} agent
   * @property {API.Delegation[]} proofs
```

packages/access-client/src/access.js
```diff
@@ -0,0 +1,344 @@
...
+ * delegation.
+ */
+class PendingAccessRequest {
+  /**
+   * @typedef {object} PendingAccessRequestModel
+   * @property {API.Agent} source.agent - Agent handling interaction.
+   * @property {API.DID} source.audience - Principal requesting an access.
+   * @property {API.ProviderDID} source.provider - Provider handling request.
+   * @property {API.UTCUnixTimestamp} source.expiration - Seconds in UTC.
+   * @property {API.Link} source.request - Link to the `access/authorize` invocation.
```
Gozala, line 136:
```suggestion
   * @property {API.Agent} agent - Agent handling interaction.
   * @property {API.DID} audience - Principal requesting an access.
   * @property {API.ProviderDID} provider - Provider handling request.
   * @property {API.UTCUnixTimestamp} expiration - Seconds in UTC.
   * @property {API.Link} request - Link to the `access/authorize` invocation.
```

PR #1063, opened 2023-11-03
fix: aggregate offer invocation cid wrong

vasco-santos:
because not pieces cid block being used but buffer block, which was meaning a different CID being computed for invocation and receipt, as seen in:

- https://github.com/web3-storage/w3up/blob/main/packages/filecoin-client/src/dealer.js#L58
- https://github.com/web3-storage/w3up/blob/main/packages/filecoin-api/src/aggregator/service.js#L97

So, `AggregatorRecord` now has a property for buffer and one for pieces. `buffer` allows us to get to know the pieces inside the `buffer` while `pieces` is the direct encoding of `PieceLink[]` dag-cbor. We could just rely on the buffer, but for issuing all `aggregate/accept` receipts for each piece, we would need an extra S3 get to read the buffer and compute the dag cbor for the list of pieces inside the buffer

### DIFF
--- a/packages/filecoin-api/src/aggregator/service.js
+++ b/packages/filecoin-api/src/aggregator/service.js
@@ -77,6 +77,7 @@ export const pieceAccept = async ({ capability }, context) => {
     }
   }
 
+  // Get buffered pieces 
   const [{ aggregate, inclusion }] = getInclusionRes.ok
   const getAggregateRes = await context.aggregateStore.get({ aggregate })
   if (getAggregateRes.error) {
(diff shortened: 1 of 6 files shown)


packages/filecoin-api/src/aggregator/service.js
```diff
@@ -77,6 +77,7 @@ export const pieceAccept = async ({ capability }, context) => {
     }
   }
 
+  // Get buffered pieces
```
vasco-santos, line 80: line 88 `.pieces` now has pieces CID instead of buffer :)

PR #1065, opened 2023-11-03
fix: a problem with yarn installation.

LanYunDev:
Fix a problem with yarn installation.

### DIFF
--- a/packages/access-client/package.json
+++ b/packages/access-client/package.json
@@ -71,7 +71,7 @@
     "bigint-mod-arith": "^3.1.2",
     "conf": "11.0.2",
     "multiformats": "^12.1.2",
-    "one-webcrypto": "git://github.com/web3-storage/one-webcrypto",
+    "one-webcrypto": "git@github.com:web3-storage/one-webcrypto.git",
     "p-defer": "^4.0.0",
     "type-fest": "^3.3.0",
     "uint8arrays": "^4.0.6"


gobengo:
Hello @LanYunDev. Thanks for the PR. The change alone won't pass CI because it leads to a new `/pnpm-lock.yaml` that must be generated with `pnpm i`. I made this PR that has your change plus that. https://github.com/web3-storage/w3up/pull/1068

Can you share more about how to reproduce the error with yarn installation that you witnessed? e.g. in which directory did you run `yarn`, and with which version of `yarn`?

LanYunDev:
node --version
v21.1.0
yarn add @web3-storage/w3cli
yarn add v1.22.19
info No lockfile found.
[1/4] 🔍  Resolving packages...
⠁ (node:7652) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
[2/4] 🚚  Fetching packages...
error Command failed.
Exit code: 128
Command: git
Arguments: ls-remote --tags --heads git://github.com/web3-storage/one-webcrypto.git
Directory: /Users/lanyun/Downloads/1
Output:
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.

The above is the previous error, I just tested it and it seems to have been fixed.

PR #1066, opened 2023-11-03
fix: dealer offer store keys without space

vasco-santos:
Removes space making new names as `2023-09-23T15:49:50.437Z_bafkzcibbdyuvfiycundi3xfjuzpuilhsijzf7ryp63tlsczf464zlwieujdqg.json`

### DIFF
--- a/packages/filecoin-api/test/events/dealer.js
+++ b/packages/filecoin-api/test/events/dealer.js
@@ -72,7 +72,7 @@ export const test = {
     const offerStoreGetAfterRename1 = await context.offerStore.get(
       `${new Date(
         aggregateRecord.insertedAt
-      ).toISOString()} ${aggregateRecord.aggregate.toString()}.json`
+      ).toISOString()}_${aggregateRecord.aggregate.toString()}.json`
     )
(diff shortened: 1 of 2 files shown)


packages/filecoin-api/test/events/dealer.js
```diff
@@ -72,7 +72,7 @@ export const test = {
     const offerStoreGetAfterRename1 = await context.offerStore.get(
       `${new Date(
         aggregateRecord.insertedAt
-      ).toISOString()} ${aggregateRecord.aggregate.toString()}.json`
+      ).toISOString()}${aggregateRecord.aggregate.toString()}.json`
```
alanshaw, line 75:
```suggestion
      ).toISOString()}_${aggregateRecord.aggregate.toString()}.json`
```

alanshaw (changes requested):

packages/filecoin-api/test/events/dealer.js
vasco-santos: yes just fixed 😓

PR #1068, opened 2023-11-03
fix: change access-client/package.json one-webcrypto git url to use different syntax that might work better with yarn

gobengo:
per https://github.com/web3-storage/w3up/pull/1065


gobengo:
this doesn't seem needed based on https://github.com/web3-storage/w3up/pull/1065#issuecomment-1793340023

PR #1078, opened 2023-11-07
feat!: collapse all clients into one

Gozala:
Collapses all clients into a single w3up-client

For backwards compatibility access-client and upload-client simply re-export their former modules which they import from the w3up-client.

I think we can release these versions on npm and then deprecate them and remove from the repo.


packages/filecoin-client/package.json
```diff
@@ -24,26 +24,25 @@
...
+      "types": "./dist/src/dealer.d.ts",
+      "import": "./src/dealer.js"
+    },
+    "./storefront": {
+      "types": "./dist/src/storefront.d.ts",
+      "import": "./src/storefront.js"
+    },
+    "./types": {
+      "types": "./dist/src/types.d.ts",
+      "import": "./src/types.js"
```
Gozala, line 45: @vasco-santos is this change ok ? Not sure why we painted to `dist` but that caused TS to fail because it was resolving to files that were not there yet.

PR #1079, opened 2023-11-07
feat: add usage/report capability

alanshaw:
Adds a capability allowing agents to request a usage report for a given period.

The report includes the size of the space at the start of the period and at the end of the period, and additionally the events that caused the space to change size. This should allow us to display current storage usage for a space as well as graph space size change over time.

### DIFF
--- a/packages/capabilities/src/types.ts
+++ b/packages/capabilities/src/types.ts
@@ -113,7 +113,7 @@ export type UsageReportSuccess = Record<ProviderDID, UsageData>
 export type UsageReportFailure = Ucanto.Failure
 
 export interface UsageData {
-  /** Space the report concerns. */
+  /** Provider the report concerns, e.g. `did:web:web3.storage` */
   provider: ProviderDID
   /** Space the report concerns. */
   space: SpaceDID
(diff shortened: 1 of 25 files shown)


packages/capabilities/src/types.ts
```diff
@@ -104,6 +105,43 @@ export interface DelegationNotFound extends Ucanto.Failure {
...
 
+// Usage
+
+export type Usage = InferInvokedCapability<typeof UsageCaps.usage>
+export type UsageReport = InferInvokedCapability<typeof UsageCaps.report>
+export type UsageReportSuccess = Record<ProviderDID, UsageData>
+export type UsageReportFailure = Ucanto.Failure
+
+export interface UsageData {
+  /** Space the report concerns. */
```
gobengo, line 116:
```suggestion
  /** Provider the report concerns. */
```

gobengo (approved):

packages/capabilities/src/usage.js
```diff
@@ -0,0 +1,42 @@
+import { capability, ok, Schema } from '@ucanto/validator'
+import { and, equal, equalWith, SpaceDID } from './utils.js'
+
+/**
+ * Capability can only be delegated (but not invoked) allowing audience to
+ * derived any `usage/` prefixed capability for the (memory) space identified
+ * by DID in the `with` field.
```
vasco-santos, line 7:
```suggestion
 * Capability can only be delegated (but not invoked) allowing audience to
 * be derived any `usage/` prefixed capability for the (memory) space identified
 * by DID in the `with` field.
```

vasco-santos (approved):
LGTM!

PR #1087, opened 2023-11-07
feat: optionally require plans for provisioning

travis:
We need this to ship billing, to stop users without a payment provider plan from provisioning new spaces.

I've added a flag to the `ProviderServiceContext` that lets us disable this requirement - this is extremely useful in testing and will be helpful as we roll this feature out - ie, w3infra can disable this requirement until we are ready to launch.

### DIFF
--- a/packages/upload-api/src/provider-add.js
+++ b/packages/upload-api/src/provider-add.js
@@ -17,7 +17,7 @@ export const provide = (ctx) =>
  */
 export const add = async (
   { capability, invocation },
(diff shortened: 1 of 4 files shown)


alanshaw (approved):
Personally, I'd name this var something more like `allowMissingPaymentPlan`.

travis:
ah I like it - I'd like `false` and `undefined` here to be equivalent so I'm going to go with `requirePaymentPlan`

PR #1091, opened 2023-11-08
feat: filecoin info

vasco-santos:
Goal of the PR is to make filecoin deals information available for end users via client (and later on CLI + console). For this:
- adds `filecoin/info` capability to `capabilities`, `filecoin-api` and `filecoin-client`
  - user can give the `piece` that was computed while uploading the file, or that was published via content claims
- exposes via `w3up-client` some filecoin sugar `*.filecoin.offer` and `*.filecoin.info` 

Note that `filecoin.info` only returns effective deals. Alternatively, user could ask directly dealTracker, even though user would need to know the aggregate where their piece was included. So, better make storefront facilitate user life with this shortcut

Would be great to actually give more information of the on going state, but I am failing to find a nice way of giving this info here. However, probably should be a separate capability (e.g. `track`, so that we avoid reading from so many stores and follow the chain, unless user specifically looks for it)

Note: I will add this to spec later on

### DIFF
--- a/packages/capabilities/src/filecoin/index.js
+++ b/packages/capabilities/src/filecoin/index.js
@@ -16,4 +16,5 @@ export {
   filecoinOffer as offer,
   filecoinSubmit as submit,
   filecoinAccept as accept,
+  filecoinInfo as info,
 } from './storefront.js'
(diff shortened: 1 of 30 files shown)


alanshaw (changes requested):
A lot of this is changes from linter...please do as separate PR next time 🙏

alanshaw (approved):
LGTM

PR #1124, opened 2023-11-14
feat: createSpace sets it as current

olizilla:
createSpace should set the new space as the current space by default.

`skipSetCurrent` is provided to allow the caller to opt-out.

License: MIT

### DIFF
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -179,9 +179,16 @@ export class Client extends Base {
    * Create a new space with a given name.
    *
    * @param {string} name
+   * @param {object} [options]
+   * @param {boolean} [options.skipSetCurrent]
    */
-  async createSpace(name) {
-    return await this._agent.createSpace(name)
+  async createSpace(name, options = {}) {
+    const space = await this._agent.createSpace(name)
+    if (options.skipSetCurrent) {
+      return space
+    }
+    await this.setCurrentSpace(space.did())
+    return space
   }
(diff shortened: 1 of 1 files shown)


alanshaw (approved):

packages/w3up-client/src/client.js
```diff
@@ -179,9 +179,16 @@ export class Client extends Base {
    * Create a new space with a given name.
    *
    * @param {string} name
+   * @param {object} [options]
+   * @param {boolean} [options.skipSetCurrent]
    */
-  async createSpace(name) {
-    return await this._agent.createSpace(name)
+  async createSpace(name, options = {}) {
+    const space = await this._agent.createSpace(name)
```
Gozala, line 186: Created space is no longer been automatically saved or made current, that is because unless you provision it with the account (that you have setup a billing for) it will not be usable.

For this reason API had been changed to something along the following lines

```js
const account = await client.login(email)
const space = await client.createSpace()
await account.provision(space)
await space.save() // if you want to save it
```

This also means in certain cases you can create spaces and even do some uploads without having to persist it across sessions.

Maybe new API has flaws and we probably should discuss how to improve it, but setting the space as default on creation does not seem like a good idea IMO given the rational above.

P.S.: Saving the space will make it current if you don't have one set already

Gozala (changes requested):
I wrote a comment explaining rational behind new interface and why I think making it default like this would be problematic.

PR #1126, opened 2023-11-14
chore: add minimal w3up upload example

olizilla:
trying to find a nice minimal example that folks can run and re-run to try thing out.

pulled from https://github.com/web3-storage/www/pull/22

License: MIT


vasco-santos (approved):
thanks for putting this together 🎉 just needs file to be included into tsconfig and should have CI happy

PR #1140, opened 2023-11-15
ci: .github/workflows/release.yml `docs` job runs when any package is released via release-please

gobengo:
Motivation:
* I think we want to run the docs release more frequently, especially if `w3up-client` is released


gobengo:
discovered while testing https://github.com/web3-storage/w3up/issues/1139 but this alone doesn't fully fix that

PR #1159, opened 2023-11-17
ci: w3up-client workflow runs when access-client changes

gobengo:
I think this makes sense because when code in access-client changes, that's going to affect `w3up-client` since `w3up-client` has a workspace-semver dep on access-client.

e.g. I think I would reasonably expect this commit to result in a release-please PR for w3up-client but it didn't. https://github.com/web3-storage/w3up/pull/1157


gobengo:
actually nvm I think it's fine

PR #1161, opened 2023-11-20
fix: package metadata

olizilla:
the most critical here is the w3up-client metadata points to the old archived repo, which is confusing.

the rest of the changes are for consistency.

License: MIT


alanshaw (approved):
Might want to `fix:` in the squashed commit message for release please to pick it up.

PR #1170, opened 2023-11-20
docs: split readme js example in two

olizilla:
split out uploading from creating a space, and update some api calls.

License: MIT


alanshaw (approved):
LGTM

PR #1171, opened 2023-11-20
feat!: account plan subscriptions and space usage API sugar

alanshaw:
This PR allows you to get a list of subscriptions for your account plan, and also get the current per space usage (I've not exposed the full report yet since we don't have any UI that would use it).

e.g.

Getting subscriptions list:

```js
const account = Object.values(client.accounts())[0]
const subs = await account.plan.subscriptions()
for (const sub of subs.ok) {
  console.log(`ID: ${sub.subscription}`)
  console.log(`Consumers: ${sub.consumers.join(', ')}`)
  console.log(`Provider: ${sub.provider}`)
}
```

Getting usage:

```js
const space = client.spaces()[0]
const usage = await space.usage.get()
console.log(`${space.did()}: ${usage.ok} bytes`)
```

So, you no longer have to invoke capabilities directly like this:
https://github.com/web3-storage/w3cli/blob/ca21f6736fb8c2180d3634f40931b91e4f0bb964/index.js#L553-L571

### DIFF
--- a/packages/w3up-client/src/space.js
+++ b/packages/w3up-client/src/space.js
@@ -1,19 +1,23 @@
 export * from '@web3-storage/access/space'
+import * as Usage from './capability/usage.js'
+import * as API from './types.js'
+
+/**
+ * @typedef {object} Model
+ * @property {API.SpaceDID} id
+ * @property {{name?:string}} [meta]
+ * @property {API.Agent} agent
+ */
 
 export class Space {
-  /** @type {import('./types.js').DID} */
-  #did
-
-  /** @type {Record<string, any>} */
-  #meta
+  #model
 
   /**
-   * @param {import('./types.js').DID} did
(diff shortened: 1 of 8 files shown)


packages/w3up-client/src/space.js
```diff
@@ -1,47 +1,99 @@
...
-  #meta
+export class Space {
+  #model
 
   /**
-   * @param {import('./types.js').DID} did
-   * @param {Record<string, any>} meta
+   * @param {Model} model
    */
-  constructor(did, meta = {}) {
```
gobengo, line 14: changes API. Should we either:
* support the old constructor signature too
* make this a major version bump
* note the breaking change but dont do more than minor (as is)

Any is fine, just noting the API change here

gobengo (approved):

travis (approved):

PR #1173, opened 2023-11-21
feat: upload-client uploadDirectory, by default, sorts the provided files by file name to help the user call us in a way that is deterministic and minimizes cost

gobengo:
Motivation:
* https://github.com/web3-storage/w3up/issues/1172
  * "Stabilize file ordering in uploadDirectory" in https://github.com/web3-storage/secrets/issues/28
    * this stabilizes it by sorting the files param unless an option is passed telling us not to


packages/upload-client/src/index.js
```diff
@@ -63,13 +67,16 @@ export async function uploadFile(conf, file, options = {}) {
  * has the capability to perform the action.
  *
  * The issuer needs the `store/add` and `upload/add` delegated capability.
- * @param {import('./types.js').FileLike[]} files File data.
+ * @param {Iterable<FileLike> & { sorted?: boolean }} files  Files that should be in the directory.
+ * To ensure determinism in the IPLD encoding, by default these files MUST be sorted by file.name or this function will return a rejected promise .
+ * To explicitly upload with in an indeterminate way, pass `files` with `files.sorted === false`.
```
gobengo, line 72:
```suggestion
 * To explicitly upload unsorted files in an indeterminate way, pass `files` with `files.sorted === false`.
```

packages/upload-client/src/index.js
```diff
@@ -63,13 +67,16 @@ export async function uploadFile(conf, file, options = {}) {
...
  *
  * The issuer needs the `store/add` and `upload/add` delegated capability.
- * @param {import('./types.js').FileLike[]} files File data.
+ * @param {Iterable<FileLike> & { sorted?: boolean }} files  Files that should be in the directory.
+ * To ensure determinism in the IPLD encoding, by default these files MUST be sorted by file.name or this function will return a rejected promise .
+ * To explicitly upload unsorted files in an indeterminate way, pass `files` with `files.sorted === false`.
  * @param {import('./types.js').UploadDirectoryOptions} [options]
  */
 export async function uploadDirectory(conf, files, options = {}) {
+  const entries = 'sorted' in files ? files : requireSortedFiles(files)
```
alanshaw, line 76: If I pass files with `sorted: true` there is nothing that checks the ordering.

packages/upload-client/src/index.js
```diff
@@ -63,13 +67,16 @@ export async function uploadFile(conf, file, options = {}) {
  * has the capability to perform the action.
  *
  * The issuer needs the `store/add` and `upload/add` delegated capability.
- * @param {import('./types.js').FileLike[]} files File data.
+ * @param {Iterable<FileLike> & { sorted?: boolean }} files  Files that should be in the directory.
```
alanshaw, line 70: Pass `sorted` as an option? It's a bit odd to me that this option is passed as a property on the input files.

Do you think we should use "order" instead of "sorted"? They are needed a particular order, whether they arrived at that order because we sorted them doesn't matter.

As a suggestion for an option name: `customOrder: true`?

packages/upload-client/src/sharding.js
```diff
@@ -84,3 +88,78 @@ export class ShardingStream extends TransformStream {
...
+    const { comparator, iterable } = this
+    return function* () {
+      let prev = null
+      for (const cur of iterable) {
+        if (prev && comparator(prev, cur) === 1) {
+          throw Object.assign(
+            new Error(
+              `sequentially iterated items were not sorted as expected`
+            ),
+            { unsorted: [prev, cur], code: 'SORTED_EXPECTATION_UNMET' }
```
alanshaw, line 118: Again, I'd call this "order".
```suggestion
            { unordered: [prev, cur], code: 'ERR_INVALID_FILE_ORDER' }
```

packages/upload-client/src/index.js
alanshaw, line 70: I'm not 100% sure it's desirable to allow streaming the listing. If the files are not verified to be ordered correctly at the start, then we run the risk of uploading 1 or more shards before failing.

...this is ignoring the fact that `UnixFS.createDirectoryEncoderStream(files)` iterates over the files immediately.

alanshaw (changes requested):
Ok, this is _not_ the PR I was expecting! I was expecting a tweak to automatically sort the files list and perhaps an option to disable that behaviour.

So something like this: https://github.com/web3-storage/files-from-path/pull/27

My assumption was that the majority of users would want this to happen automatically and by pushing that responsibility onto users we're forcing a breaking change as well as adding an additional code task for them to sort the files themselves.

I've made some suggestions in the PR but on balance I think we should more preferably do the above (i.e. same as `files-from-path`) instead.

packages/upload-client/src/sharding.js
```diff
@@ -84,3 +88,78 @@ export class ShardingStream extends TransformStream {
...
+          )
+        }
+        yield cur
+        prev = cur
+      }
+    }.bind(this)()
+  }
```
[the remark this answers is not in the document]
gobengo: thanks 🤦

packages/upload-client/src/sharding.js
gobengo: good suggestion, thank you

packages/upload-client/src/index.js
gobengo: This was on purpose. I was trying to make the API basically:
* if you pass it something ordered by name, it works
* if you pass it something unordered by name and with no indication you've sorted it how you want, throw an error once we find it is unordered
* if you pass it something unordered by name and with an explicit indication you've already ordered/sorted it intentionally (in some way but not necessarily by file name), allow the potentially nondeterministic ordering

So my intention was that passing `sorted: true` would be that the caller is indicating "I already sorted this in a way that I want, so don't try to check the ordering or order it on my behalf".
`sorted: false` would also be like "It's not sorted but I'm intentionally marking it as not sorted because I know what I'm doing and want to upload a directory of unsorted things"

But what you've described as what you expected is also reasonable so I will make it work that way

packages/upload-client/src/index.js
gobengo: > Do you think we should use "order" instead of "sorted"? They are needed a particular order, whether they arrived at that order because we sorted them doesn't matter.

Useful distinction. Thank you. I updated to prefer this notion of 'order' where appropriate

> As a suggestion for an option name: customOrder: true?

👍

gobengo:
> My assumption was that the majority of users would want this to happen automatically and by pushing that responsibility onto users we're forcing a breaking change as well as adding an additional code task for them to sort the files themselves.

@alanshaw cool. I updated to make this match the expectations you shared.

packages/upload-client/src/index.js
```diff
@@ -63,13 +63,17 @@ export async function uploadFile(conf, file, options = {}) {
  * has the capability to perform the action.
  *
  * The issuer needs the `store/add` and `upload/add` delegated capability.
- * @param {import('./types.js').FileLike[]} files File data.
+ * @param {import('./types.js').FileLike[]} files  Files that should be in the directory.
+ * To ensure determinism in the IPLD encoding, by default these files MUST be sorted by file.name or this function will return a rejected promise .
+ * To explicitly upload unsorted files in an indeterminate way, pass `files` with `files.sorted === false`.
```
alanshaw, line 68:
```suggestion
 * To ensure determinism in the IPLD encoding, files are automatically sorted by `file.name`. 
 * To retain the order of the files as passed in the array, set `customOrder` option to `true`.
```

packages/upload-client/src/sharding.js
```diff
@@ -84,3 +88,30 @@ export class ShardingStream extends TransformStream {
...
 }
+
+/**
+ * Default comparator for FileLikes. Sorts by file name in ascending order.
+ *
+ * @param {FileLike} a
+ * @param {FileLike} b
+ * @param {(file: FileLike) => string} getSortKey
+ */
+export const defaultFileComparator = (a, b, getSortKey = (a) => a.name) => {
```
alanshaw, line 99: Minor, but `getSortKey` is actually returning a value not a key.

packages/upload-client/src/types.ts
```diff
@@ -258,7 +258,10 @@ export interface UploadOptions
 export interface UploadDirectoryOptions
   extends UploadOptions,
     UnixFSDirectoryEncoderOptions,
-    UploadProgressTrackable {}
+    UploadProgressTrackable {
+  // whether the directory files have already been ordered in a custom way. indicates that the upload must not use a different order than the one provided.
```
alanshaw, line 262: This should be `/** ... */` style comment in order to expose this helpful info to consumers.

alanshaw (approved):
❤️ Some minor feedback but nothing blocking.

packages/upload-client/src/index.js
alanshaw: ACK makes sence, I think it was just the 3 possible values boolean that got me - null/undefined and true and false. Where true/false actually means the same thing in terms of what _we_ do with the files.

packages/upload-client/src/index.js
gobengo: ah thx!!

packages/upload-client/src/sharding.js
gobengo: ah I was using `key` in the sense of this https://docs.python.org/3/howto/sorting.html#key-functions
but I think you mean it like the 'key/value' of the File object, which is probably a more reasonable interpretation in this context than what I meant. I will rename it to `getSortValue` because that seems at least slightly less likely to cause confusion than what I did

PR #1174, opened 2023-11-21
docs: simplify the readme js code example

olizilla:
its better if we direct folks to the docs or the workshopper rather than try and explain it all here.

License: MIT


README.md
```diff
@@ -40,44 +40,31 @@ Run `w3 --help` or have a look at https://github.com/web3-storage/w3cli to find
...
-// give you space a nickname to remember it by
-const space = await client.createSpace('my space')
-
-console.log(space.name()) // 'my space'
-console.log(space.did())  // 'did:key:xyz...'
-```
-
 See https://web3.storage/docs/w3up-client for a guide to using the js client for the first time.
 
+For an interactive command line adventure into the using w3up check out `learnyouw3up` here https://github.com/web3-storage/learnyouw3up
```
vasco-santos, line 66: my main concern is that we need to make sure to keep this updated, which feels more difficult compared to within the README. 
I would personally prefer to have examples folder, and tests for them so that they keep updated when PRs land

olizilla: i will remove the link to learnyouw3up when it is not longer maintained. Right now it's our best documentation.

PR #1177, opened 2023-11-21
fix: don't error when we can't figure out a name for a space

travis:
@juliangruber reported an issue with our documented space creation flow when the exported space access delegation doesn't specify a name and the client doesn't specify one manually:

https://github.com/web3-storage/w3up/issues/1175

I think it's reasonable to require a name for spaces at some level, but I think the access-client API is too low-level for this.


travis:
if we do want to update the docs rather than patch as I'm proposing here, we'll need to update https://github.com/web3-storage/w3up/blob/main/packages/w3up-client/src/client.js#L227 to accept a name that can be passed down through

vasco-santos (approved):
Given we are trying to move forward with store less possible metadata as we could move that to a different layer with Pail, I would move this way. But curious to know other opinions

alanshaw:
Yeah this shouldn't be mandatory.

PR #1178, opened 2023-11-24
feat: add store.get and upload.get to clients

alanshaw:
This was implemented in `capabilities` and `upload-api` but for some reason was never hooked in the clients.


vasco-santos (approved):
lgtm

PR #1184, opened 2023-11-27
fix: reset function not working

zhanleewo:
The async function returned by `withObjectStore` was not called.


alanshaw:
Resolved with a test in https://github.com/web3-storage/w3up/pull/1199

Thanks @zhanleewo!

PR #1188, opened 2023-11-27
feat: aggregator keeping oldest piece ts

vasco-santos:
Conveniently store the timestamp of the oldest piece in table, so that when monitoring aggregates time to deal for alerts, we do not need to read buffer, decode it, and find out it every time a CRON runs.


alanshaw (approved):
I think this is fine but I'd perhaps call it something like `minPieceAge`...

vasco-santos:
Age would be more a difference from now to a given date, and not the given date itself. But I can rename from `oldestPieceInsertedAt` to `minPieceInsertedAt`

PR #1213, opened 2023-12-04
feat!: return allocated bytes in `store/add` receipt

alanshaw:
This PR adds a new field to the `store/add` success result: `allocated: number` - the total bytes allocated in the space to accommodate the stored item, it **may be zero if the item is _already_ stored in _this_ space**.

This allows us to accurately report and bill for items stored in a space. Currently, `store/add` when the item is already stored in the space will cause a new diff to be created, effecitvely counting the same shard multiple times. This could happen because of accidentally uploading the same item, or because of a retry of a big item where some shards were successful.

The following additional changes enable this functionality:

* `StoreTable` and `UploadTable` methods now have return types that are `Result<O, X>` - this allows us to specify and communicate 2 errors `RecordNotFound` and `RecordKeyConflict` (explained in the following bullets).
* In the context of these 2 tables the semantics of `insert` have changed to be more in line with postgres/other DBs - "insert" means add to the DB or fail (with `RecordKeyConflict`) if an item with the same key already exists.
    * We can satisfy this constraint in dynamodb with [condition expressions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html#Expressions.ConditionExpressions.PreventingOverwrites) or [`ReturnValues`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_PutItem.html#DDB-PutItem-request-ReturnValues).
* In the `UploadTable` I've renamed `insert` to `upsert` since the behaviour in use here is more akin to "update or insert" - aka ["upsert"](https://en.wiktionary.org/wiki/upsert).
* `remove` and `get` now fail with `RecordNotFound` if the item to delete is not available.
    * Again in dynamodb we can satisfy this constraint with condition expressions or `ReturnValues`

### DIFF
--- a/packages/capabilities/src/types.ts
+++ b/packages/capabilities/src/types.ts
@@ -513,14 +513,7 @@ export type UploadAddSuccess = Omit<UploadListItem, 'insertedAt' | 'updatedAt'>
 
 export type UploadGetSuccess = UploadListItem
 
-export type UploadRemoveSuccess = UploadDidRemove | UploadDidNotRemove
-
-export interface UploadDidRemove extends UploadAddSuccess {}
-
-export interface UploadDidNotRemove {
-  root?: undefined
-  shards?: undefined
-}
+export type UploadRemoveSuccess = UploadAddSuccess
 
 export interface UploadListSuccess extends ListResponse<UploadListItem> {}
 
--- a/packages/capabilities/src/types.ts
+++ b/packages/capabilities/src/types.ts
@@ -444,7 +444,7 @@ export type StoreAddSuccessStatusDone = 'done'
 
 export interface StoreAddSuccessResult {
   /**
-   * Status of the item to store. A "done" status incidactes that it is not
+   * Status of the item to store. A "done" status indicates that it is not
    * necessary to upload the item. An "upload" status indicates that the item
    * should be uploaded to the provided URL.
    */
--- a/packages/capabilities/src/types.ts
+++ b/packages/capabilities/src/types.ts
@@ -438,18 +438,34 @@ export type StoreRemove = InferInvokedCapability<typeof StoreCaps.remove>
 export type StoreList = InferInvokedCapability<typeof StoreCaps.list>
 
 export type StoreAddSuccess = StoreAddSuccessDone | StoreAddSuccessUpload
-export interface StoreAddSuccessDone {
-  status: 'done'
+
+export type StoreAddSuccessStatusUpload = 'upload'
+export type StoreAddSuccessStatusDone = 'done'
+
+export interface StoreAddSuccessResult {
+  /**
+   * Status of the item to store. A "done" status incidactes that it is not
+   * necessary to upload the item. An "upload" status indicates that the item
+   * should be uploaded to the provided URL.
+   */
+  status: StoreAddSuccessStatusUpload|StoreAddSuccessStatusDone
+  /**
+   * Total bytes allocated in the space to accommodate this stored item.
+   * May be zero if the item is _already_ stored in _this_ space.
+   */
+  allocated: number
+  /** DID of the space this item will be stored in. */
   with: DID
+  /** CID of the item. */
   link: UnknownLink
-  url?: undefined
-  headers?: undefined
 }
 
-export interface StoreAddSuccessUpload {
-  status: 'upload'
-  with: DID
-  link: UnknownLink
+export interface StoreAddSuccessDone extends StoreAddSuccessResult {
+  status: StoreAddSuccessStatusDone
+}
+
+export interface StoreAddSuccessUpload extends StoreAddSuccessResult {
+  status: StoreAddSuccessStatusUpload
   url: ToString<URL>
   headers: Record<string, string>
 }
--- a/packages/capabilities/src/types.ts
+++ b/packages/capabilities/src/types.ts
@@ -448,7 +448,7 @@ export interface StoreAddSuccessResult {
    * necessary to upload the item. An "upload" status indicates that the item
    * should be uploaded to the provided URL.
    */
-  status: StoreAddSuccessStatusUpload|StoreAddSuccessStatusDone
+  status: StoreAddSuccessStatusUpload | StoreAddSuccessStatusDone
   /**
    * Total bytes allocated in the space to accommodate this stored item.
    * May be zero if the item is _already_ stored in _this_ space.
--- a/packages/upload-api/src/store/get.js
+++ b/packages/upload-api/src/store/get.js
@@ -1,6 +1,7 @@
 import * as Server from '@ucanto/server'
 import * as Store from '@web3-storage/capabilities/store'
 import * as API from '../types.js'
+import { StoreItemNotFound } from './lib.js'
 
 /**
  * @param {API.StoreServiceContext} context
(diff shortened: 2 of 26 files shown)


vasco-santos (approved):
LGTM
small things inline

packages/capabilities/src/types.ts
```diff
@@ -438,18 +438,34 @@ export type StoreRemove = InferInvokedCapability<typeof StoreCaps.remove>
...
+   * Status of the item to store. A "done" status indicates that it is not
+   * necessary to upload the item. An "upload" status indicates that the item
+   * should be uploaded to the provided URL.
+   */
+  status: StoreAddSuccessStatusUpload | StoreAddSuccessStatusDone
+  /**
+   * Total bytes allocated in the space to accommodate this stored item.
+   * May be zero if the item is _already_ stored in _this_ space.
+   */
+  allocated: number
```
gobengo, line 456: We may want to advertise how much we can ensure consistency of this, which I imagine might be 'not very much'.

e.g. let's say cid A is not stored in the space. Two `store/add` invocations get created with different CIDs (e.g. different nonce/expirty) to add cid A to the space. Then those two invocations are submitted to us at the exact same second. iiuc it may be the case that both of those responses come back with `allocated: size(A)` (because we dont have transactions, and in this scenario both invocations calls `carStoreBucket.has(link)` [here](https://github.com/web3-storage/w3up/blob/880f0013049862658e07038016b647c323920f16/packages/upload-api/src/store/add.js#L37) would have returned false.

my main concern
* someone might use this `allocated` response value in some business logic without being aware of the eventual consistency lack-of-transactionality in generating this value. i.e. this could be nonzero even if technically the a different `store/add` invocation added the CID first (in the race condition describes above).
* in that race condition, the same CID may have multiple store/add receipts whose results have `allocated` set to the size of the referent of the CID.

If we want to return this information in the result of the write operation (store/add), we probably need something like [transactions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html) to do it in a way that isn't subject to these race conditions.
But if possible it may be nice to avoid returning this value in the result of the write operation (because it almost always increases the cost of the write operation to add return values that. must also be consistent)

alternatively we ignore the above concerns but maybe it's a good idea to at least put in the jsdoc here that there a nonzero response does not necessarily indicate that the `store/add` is the authoritative addition of the CID to the space.

alanshaw: Great point, in hindsight this is maybe not the right solution because _we_ want to use it in _our_ business logic...

vasco-santos: good point, so maybe we want the insert later on to have a result of being conflict or not. Only one Put should happen in dynamo successfully

alanshaw: Ok, on reflection I think we _can_ make `allocated` in the response be authoratitive about the addition of the CID to the space. This is what we can do:

* We don't need `TransactWriteItems` since we're not writing multiple items!
* We can use conditional expressions to ensure writes fail:
    * Change store/add `PutItem` so that it fails if [not a new record](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html#Expressions.ConditionExpressions.PreventingOverwrites)
    * Change store/remove `DeleteItem` so that it fails if record does not exist

alanshaw:
@gobengo please for another review 🙏 

A few changes since you last looked:

* `StoreTable` and `UploadTable` methods now have return types that are `Result<O, X>` - this allows us to specify and communicate 2 errors `RecordNotFound` and `RecordKeyConflict` (explained in the following bullets).
* In the context of these 2 tables the semantics of `insert` have changed to be more in line with postgres/other DBs - "insert" means add to the DB or fail (with `RecordKeyConflict`) if an item with the same key already exists.
    * We can satisfy this constraint in dynamodb with [condition expressions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html#Expressions.ConditionExpressions.PreventingOverwrites) or [`ReturnValues`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_PutItem.html#DDB-PutItem-request-ReturnValues).
* In the `UploadTable` I've renamed `insert` to `upsert` since the behaviour in use here is more akin to "insert or update" - aka ["upsert"](https://en.wiktionary.org/wiki/upsert).
* `remove` and `get` now fail with `RecordNotFound` if the item to delete is not available.
    * Again in dynamodb we can satisfy this constraint with condition expressions or `ReturnValues`

packages/upload-api/src/store/get.js
```diff
@@ -14,16 +15,9 @@ export function storeGetProvider(context) {
...
-        error: {
-          name: 'StoreItemNotFound',
-          message: 'Store item not found',
-        },
-      }
-    }
-    return {
-      ok: res,
+    if (res.error && res.error.name === 'RecordNotFound') {
+      return Server.error(new StoreItemNotFound(space, link))
```
vasco-santos, line 19: Shouldn't this be wrapped into `error` like before?

`context.storeTable.get(space, link)` returns a Result<A, B>, but `ucanto` error is not such format https://github.com/web3-storage/ucanto/blob/main/packages/server/src/handler.js#L268 ?

Therefore, this is inconsistent with return for res, as well as other files like `list.js`

packages/upload-api/src/store/remove.js
```diff
@@ -11,41 +12,13 @@ export function storeRemoveProvider(context) {
     const { link } = capability.nb
     const space = Server.DID.parse(capability.with).did()
 
-    const item = await context.storeTable.get(space, link)
-    if (!item) {
-      return Server.error(new StoreItemNotFound(space, link))
+    const res = await context.storeTable.remove(space, link)
+    if (res.error && res.error.name === 'RecordNotFound') {
+      if (res.error.name === 'RecordNotFound') {
+        return Server.error(new StoreItemNotFound(space, link))
```
vasco-santos, line 18: same as above?

packages/upload-api/src/store/get.js
vasco-santos: so it is not consistent with server.fail https://github.com/web3-storage/ucanto/blob/main/packages/core/src/result.js#L58 too

packages/upload-api/src/upload/get.js
```diff
@@ -14,16 +15,9 @@ export function uploadGetProvider(context) {
...
-        error: {
-          name: 'UploadNotFound',
-          message: 'Upload not found',
-        },
-      }
-    }
-    return {
-      ok: res,
+    if (res.error && res.error.name === 'RecordNotFound') {
+      return Server.error(new UploadNotFound(space, root))
```
vasco-santos, line 19: error format?

packages/upload-api/src/upload/remove.js
```diff
@@ -11,10 +12,13 @@ export function uploadRemoveProvider(context) {
...
     const space = Server.DID.parse(capability.with).did()
 
-    const result = await context.uploadTable.remove(space, root)
-
-    return {
-      ok: result || {},
+    const res = await context.uploadTable.remove(space, root)
+    if (res.error) {
+      if (res.error.name === 'RecordNotFound') {
+        return Server.error(new UploadNotFound(space, root))
```
vasco-santos, line 18: error format

vasco-santos (changes requested):
I think there is some weirdness on the return types of errors? I don't understand why ts is not complaining

vasco-santos (approved):

packages/upload-api/src/store/get.js
alanshaw: `Server.error(new Error('boom'))` => `{ error: new Error('boom') }`

gobengo (approved):

ALIK202 (commented):
0xBa5c04Cdf4771ac6Da3f48fc368314a00DA064B7

PR #1219, opened 2023-12-06
fix: use an ArrayBuffer for delegation bits in AgentData

travis:
we noticed that Safari users couldn't log in:

https://github.com/web3-storage/console/issues/46

I traced this down to a seemingly unreported WebKit bug where IndexedDB throws an error when saving Uint8Arrays in an object that uses inline-keys. I don't know these two things don't seem related, but there does seem to be some similar precedent:

https://github.com/dexie/Dexie.js/issues/656#issuecomment-391038490

This change tweaks `AgentData`'s export format to use `ArrayBuffer`s rather than Uint8Arrays, which should solve the issue. I was able to handle coercing back into Uint8Arrays pretty elegantly, and introduced a new `AgentDataImport` type to keep things typesafe.

### DIFF
--- a/packages/access-client/src/agent-data.js
+++ b/packages/access-client/src/agent-data.js
@@ -8,6 +8,18 @@ import { isExpired } from './delegations.js'
 
 /** @typedef {import('./types.js').AgentDataModel} AgentDataModel */
 
+/**
+ * Convert a Uint8Array to an ArrayBuffer, taking into account
+ * that we may be looking at a "data view".
+ * thanks, https://stackoverflow.com/a/54646864
+ * 
+ * @param {Uint8Array} array 
+ * @returns ArrayBuffer
+ */
+function uint8ArrayToArrayBuffer(array) {
+  return array.buffer.slice(array.byteOffset, array.byteLength + array.byteOffset)
+}
+
 /** @implements {AgentDataModel} */
 export class AgentData {
   /** @type {(data: import('./types.js').AgentDataExport) => Promise<void> | void} */
@@ -53,7 +65,7 @@ export class AgentData {
   /**
    * Instantiate AgentData from previously exported data.
    *
-   * @param {import('./types.js').AgentDataExport} raw
+   * @param {import('./types.js').AgentDataImport} raw
    * @param {import('./types.js').AgentDataOptions} [options]
    */
   static fromExport(raw, options) {
@@ -65,7 +77,7 @@ export class AgentData {
         delegation: importDAG(
           value.delegation.map((d) => ({
             cid: CID.parse(d.cid).toV1(),
-            bytes: d.bytes,
+            bytes: new Uint8Array(d.bytes),
           }))
         ),
         meta: value.meta,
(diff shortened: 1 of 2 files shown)


gobengo (approved):

packages/access-client/src/agent-data.js
```diff
@@ -65,7 +80,7 @@ export class AgentData {
         delegation: importDAG(
           value.delegation.map((d) => ({
             cid: CID.parse(d.cid).toV1(),
-            bytes: d.bytes,
+            bytes: new Uint8Array(d.bytes),
```
alanshaw, line 83: Avoid copy if already a `Uint8Array`?

```suggestion
            bytes: d.bytes instanceof Uint8Array ? d.bytes : new Uint8Array(d.bytes),
```

alanshaw (commented):
Great detective work!

I'm hesitant to inflict this conversion on other browsers because one of them has a bug.

Personally I'd do some feature detection in `access-client/src/drivers/indexeddb.js` and export a custom driver for safari that uses "out of line" keys:

```js
export const IndexedDBDriver = isSafari
  ? OutOfLineKeyDriver 
  : InLineKeyDriver
```

Then Safari uses a different driver, existing stores continue to work, and there's no extra type needed.

You alternatively could export both and push the feature detection into the `w3up-client`...

alanshaw:
Just FYI I explicitly didn't request changes here it's just a suggestion 😁

Just a few notes after reading your comments:

We don't deal directly with ArrayBuffers anywhere else in the IPLD stack including Ucanto - whenever we have byte views they are always Uint8Arrays. So this would be a little different to this convention.

I'm 🤷 on supporting it forever - we very rarely change these and even if Safari fix the issue we don't need to migrate...

I think it's really the additional type that bothers me the most. If you manually constructed an `AgentDataImport` object there's no way to know if it's compatible with your driver...

travis:
ok cool! I've implemented your suggestions and added a test to ensure we're not doing any extra copying. I've also gotten rid of the `AgentDataImport` type.

packages/access-client/src/utils/buffers.js
```diff
@@ -0,0 +1,21 @@
...
+ * that we may be looking at a "data view".
+ * thanks, https://stackoverflow.com/a/54646864
+ * 
+ * If we aren't looking at a data view, simply returns the underlying ArrayBuffer
+ * directly.
+ *
+ * @param {Uint8Array} array
+ * @returns ArrayBuffer
+ */
+export function uint8ArrayToArrayBuffer(array) {
```
gobengo, line 12: btw I don't understand why this can't just return `array.buffer`

If the point is explicitly to copy to a new array, then perhaps the jsdoc should say that (and maybe the function name should have 'copy' in it?)

gobengo: like maybe this could be called `copyTypedArray`

Also if the point is to accomodate a `DataView` maybe the type of the `array` param should be `DataView`?

gobengo: travis corrected me that `Uint8Array#buffer` may actually refer to a much wider buffer

PR #1220, opened 2023-12-06
feat: upload-client funcs have options.piece to opt in to piece link calculation

gobengo:
Motivation:
* @Gozala suggested it to help debug
* allow callers to opt into piece generation, and it's disabled by default (until we can ensure it'll work well for huge files)
  * we may not keep this, but it could be helpful to allow `w3 up` to be used to upload huge files without the piece link calculation (in case it is erroring or taking too much memory)

### DIFF
--- a/packages/upload-client/src/types.ts
+++ b/packages/upload-client/src/types.ts
@@ -253,8 +253,8 @@ export interface UploadOptions
     ShardStoringOptions,
     UploadProgressTrackable {
   onShardStored?: (meta: CARMetadata) => void
-  /** when true, uploading will skip calculation of filecoin piece link */
-  skipPieceLink?: boolean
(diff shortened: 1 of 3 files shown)


packages/upload-client/src/types.ts
```diff
@@ -253,6 +253,7 @@ export interface UploadOptions
     ShardStoringOptions,
     UploadProgressTrackable {
   onShardStored?: (meta: CARMetadata) => void
+  skipPieceLink?: boolean
```
gobengo, line 256:
```suggestion
  /** when true, uploading will skip calculation of filecoin piece link */
  skipPieceLink?: boolean
```

Gozala (approved):

PR #1222, opened 2023-12-06
uploadBlockStream uses less streams and parallel

gobengo:


tsconfig.json
```diff
@@ -24,7 +24,7 @@
     // linter checks
     "noImplicitReturns": false,
     "noFallthroughCasesInSwitch": true,
-    "noUnusedLocals": true,
+    // "noUnusedLocals": true,
```
gobengo, line 27: can we not use this. It means thinks won't build just because you temporarily comment out some code

gobengo:
wow somehow this commit to fix tests also regresses the nicer memory usage. there must be some other subtle thing here I'm missing...
https://github.com/web3-storage/w3up/pull/1222/commits/dccd808a18f097301a36ada22ac158de2e8926d9

alanshaw:
It was the parralel transform stream buffering data. Resolved upstream in the module and updated here: https://github.com/web3-storage/w3up/pull/1227

gobengo:
@alanshaw Thanks for tracking it down! 🎉

PR #1227, opened 2023-12-07
fix: memory usage

alanshaw:
Previously `parallel-transform-web` was running up to N transform tasks concurrently, but while those tasks were running it buffered the rest of the stream data internally 🤦 .

This was resolved in https://github.com/alanshaw/parallel-transform-web/pull/1 so just pulling in the latest code here.

I was seeing memory usage in w3cli for a 12GB file grow to ~10GB but now it stays at around at round ~4GB (which is the same usage I see when I remove `parallel-transform-web` and replace it with a regular serial `TransformStream`).


vasco-santos (approved):
thanks ❤️

PR #1240, opened 2023-12-13
feat: dont use ipfs-utils fetch with upload progress by default 

gobengo:
because importing it in bundled setups breaks, e.g.
* https://github.com/web3-storage/w3up/issues/1183
* https://github.com/web3-storage/w3up/issues/828

I tested this change locally and confirmed that once it's made, upload-client can be relied on by a nextjs project and build successfully

### DIFF
--- a/packages/upload-client/src/fetch-with-upload-progress.js
+++ b/packages/upload-client/src/fetch-with-upload-progress.js
@@ -4,8 +4,5 @@ import ipfsUtilsFetch from 'ipfs-utils/src/http/fetch.js'
  * @type {import('./types.js').FetchWithUploadProgress}
  */
 export const fetchWithUploadProgress = (url, init) => {
-  if (init && 'readable' in init) {
-    throw new Error('init cannot be readable')
-  }
   return ipfsUtilsFetch.fetch(url, init)
 }
--- a/packages/upload-client/src/fetch-with-upload-progress.js
+++ b/packages/upload-client/src/fetch-with-upload-progress.js
@@ -1,0 +1,11 @@
+import ipfsUtilsFetch from 'ipfs-utils/src/http/fetch.js'
+
+/**
+ * @type {import('./types.js').FetchWithUploadProgress}
+ */
+export const fetchWithUploadProgress = (url, init) => {
+  if (init && 'readable' in init) {
+    throw new Error('init cannot be readable')
+  }
+  return ipfsUtilsFetch.fetch(url, init)
+}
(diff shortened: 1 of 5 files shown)


packages/upload-client/src/store.js
```diff
@@ -128,6 +127,16 @@ export async function add(
     }
   )
 
+  if (!fetchDidCallUploadProgressCb && options.onUploadProgress) {
+    // the fetch implementation didn't support onUploadProgress
+    const carBlob = new Blob([car])
+    options.onUploadProgress({
+      total: carBlob.size,
```
travis, line 134: are `total` and `loaded` here the same because this will be the only progress event sent?

travis (approved):

packages/upload-client/src/store.js
gobengo: yep! the only progress even for this car file, and all we know here is the fetch succeeded so it's 100% progress all at once

packages/upload-client/src/store.js
```diff
@@ -90,10 +88,9 @@ export async function add(
   const responseAddUpload = result.out.ok
 
   const fetchWithUploadProgress =
-    /** @type {(url: string, init?: import('./types.js').FetchOptions) => Promise<Response>} */ (
-      fetch
-    )
+    options.fetchWithUploadProgress || options.fetch || globalThis.fetch
```
alanshaw, line 91: Needs `globalThis.fetch.bind(globalThis)`

packages/upload-client/src/types.ts
```diff
@@ -45,6 +45,16 @@ import {
   UsageReportFailure,
 } from '@web3-storage/capabilities/types'
 
+type Override<T, R> = Omit<T, keyof R> & R
+
+type FetchOptions = Override<
+  IpfsUtilsFetchOptions,
+  {
+    // `fetch` is a browser API and browsers don't have `Readable`
+    body: Exclude<IpfsUtilsFetchOptions['body'], import('node:stream').Readable>
```
alanshaw, line 54: What does this do?

packages/upload-client/src/store.js
alanshaw: Just call it `customFetch` and type it as required?

packages/upload-client/src/fetch-with-upload-progress.js
```diff
@@ -0,0 +1,8 @@
+import ipfsUtilsFetch from 'ipfs-utils/src/http/fetch.js'
```
alanshaw, line 1: 🙏 Pull out as a separate library that solves the issue or send a PR.

packages/upload-client/src/types.ts
gobengo: it excludes `Readable` from the allowed body types that come from `ipfs-utils` because that's not what other `fetches` will expect

PR #1244, opened 2023-12-13
feat: expose OwnedSpace and SharedSpace from access-client

travis:
Not having these available as types is making it hard to build the web console space creation wizard.


Gozala (approved):
I would personally prefer defining interfaces that expose only necessary functionality as opposed to exporting classes themself. That said I I'm ok with exporting them as a temporary measure.

PR #1248, opened 2024-01-04
feat: add remove high level function to client

vasco-santos:
This PR adds high level function for removing to the client. It essentially follows what has been done in the CLI https://github.com/web3-storage/w3cli/blob/main/index.js#L189-L237

We currently provide this feature according to the docs https://web3.storage/docs/how-to/remove/#using-the-js-client-or-cli but it is not there.

It is annoying that we have docs for something that does not exist. Other option would be to change the docs to use low level capabilities, even though that means explaining more to the users what shards are, which I think is not the best direction. 

Two discussion points:
- Not performing this in an atomic operation may lead to inconsistencies, where for instance part of the shards could be removed but then other operations fail... This is also the current behaviour in the CLI. We could return the result of Promise.allSettled if that would be better.
  - but honestly, I think we should consider not facilitating a high level function to client as it is not an atomic op, may have mutations on partial failure, and as pointed out when part of multiple uploads may also be an issue.
- CLI and docs have defaults as implemented here, but I think it would make more sense to delete shards by default in this high level function, any thoughts?

Closes https://github.com/web3-storage/w3up/issues/1246

### DIFF
--- a/packages/w3up-client/README.md
+++ b/packages/w3up-client/README.md
@@ -603,7 +603,7 @@ function remove (
   options: {
     shards?: boolean
   } = {}
-): Promise<Delegation>
+): Promise<void>
 ```
 
 Removes association of a content CID with the space. Optionally, also removes association of CAR shards with space.
--- a/packages/w3up-client/README.md
+++ b/packages/w3up-client/README.md
@@ -608,7 +608,7 @@ function remove (
 
 Removes association of a content CID with the space. Optionally, also removes association of CAR shards with space.
 
-⚠️ Please be aware that all shards will be deleted even if there is another upload(s) that reference same shards, which in turn could corrupt those uploads.
+⚠️ If `shards` option is `true` all shards will be deleted even if there is another upload(s) that reference same shards, which in turn could corrupt those uploads.
 
 ### `getReceipt`
 
--- a/packages/w3up-client/README.md
+++ b/packages/w3up-client/README.md
@@ -394,6 +394,7 @@ sequenceDiagram
   - [`addProof`](#addproof)
   - [`delegations`](#delegations)
   - [`createDelegation`](#createdelegation)
+  - [`remove`](#remove)
   - [`capability.access.authorize`](#capabilityaccessauthorize)
   - [`capability.access.claim`](#capabilityaccessclaim)
   - [`capability.space.info`](#capabilityspaceinfo)
@@ -593,6 +594,21 @@ function createDelegation (
 ```
 
 Create a delegation to the passed audience for the given abilities with the _current_ space as the resource.
+
+### `remove`
+
+```ts
+function remove (
+  contentCID?: CID
+  options: {
+    shards?: boolean
+  } = {}
+): Promise<Delegation>
+```
+
+Removes association of a content CID with the space. Optionally, also removes association of CAR shards with space.
+
+⚠️ Please be aware that all shards will be deleted even if there is another upload(s) that reference same shards, which in turn could corrupt those uploads.
 
 ### `getReceipt`
 
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -324,20 +324,29 @@ export class Client extends Base {
    * @param {boolean} [options.shards]
    */
   async remove(contentCID, options = {}) {
-    // Remove association of content CID with selected space.
+    // Shortcut if there is no request to remove shards
+    if (!options.shards) {
+      // Remove association of content CID with selected space.
+      try {
+        await this.capability.upload.remove(contentCID)
+        /* c8 ignore start */
+      } catch (/** @type {any} */ err) {
+        throw new Error(
+          `remove failed for ${contentCID}: ${err.message ?? err}`
+        )
+      }
+      return
+    }
+
+    // Get shards associated with upload.
     let upload
     try {
-      upload = await this.capability.upload.remove(contentCID)
+      upload = await this.capability.upload.get(contentCID)
       /* c8 ignore start */
     } catch (/** @type {any} */ err) {
       throw new Error(`remove failed for ${contentCID}: ${err.message ?? err}`)
     }
     /* c8 ignore stop */
-
-    // Shortcut if there is no request to remove shards
-    if (!options.shards) {
-      return
-    }
 
     // Check if we have information about the shards to remove
     if (!upload.root) {
@@ -353,7 +362,30 @@ export class Client extends Base {
 
     // Remove shards
     await Promise.allSettled(
-      upload.shards.map((shard) => this.capability.store.remove(shard))
+      upload.shards.map(async (shard) => {
+        try {
+          await this.capability.store.remove(shard)
+        } catch (/** @type {any} */ error) {
+          /* c8 ignore start */
+          // If not found, we can tolerate error as it may be a consecutive call for deletion where first failed
+          if (error?.cause?.name !== 'StoreItemNotFound') {
+            throw new Error(
+              `remove failed for shard ${shard.link()} from ${contentCID}: ${
+                error?.message
+              }`
+            )
+          }
+          /* c8 ignore stop */
+        }
+      })
     )
+
+    // Remove association of content CID with selected space.
+    try {
+      await this.capability.upload.remove(contentCID)
+      /* c8 ignore start */
+    } catch (/** @type {any} */ err) {
+      throw new Error(`remove failed for ${contentCID}: ${err.message ?? err}`)
+    }
   }
 }
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -327,65 +327,34 @@ export class Client extends Base {
     // Shortcut if there is no request to remove shards
     if (!options.shards) {
       // Remove association of content CID with selected space.
-      try {
-        await this.capability.upload.remove(contentCID)
-        /* c8 ignore start */
-      } catch (/** @type {any} */ err) {
-        throw new Error(
-          `remove failed for ${contentCID}: ${err.message ?? err}`
-        )
-      }
+      await this.capability.upload.remove(contentCID)
       return
     }
 
     // Get shards associated with upload.
-    let upload
-    try {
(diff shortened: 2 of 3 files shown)


packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,46 @@ export class Client extends Base {
...
+   *
+   * @param {import('multiformats').UnknownLink} contentCID
+   * @param {object} [options]
+   * @param {boolean} [options.shards]
+   */
+  async remove(contentCID, options = {}) {
+    // Remove association of content CID with selected space.
+    let upload
+    try {
+      upload = await this.capability.upload.remove(contentCID)
```
Gozala, line 327: I would expect this to be run only after we remove all the shards instead of doing it before we remove the shards. That is because is one of the operations fails, caller will still have an opportunity to re-try. On the other hand if we remove the upload and fail to remove one of the shards, there will be no way for the user to retry or clear the shards.

If there is a good reason to do it this way can you please add code comments elaborating it. Otherwise I suggest to try removing all the shards and if they succeed only after remove the root.

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,46 @@ export class Client extends Base {
       proofs: options.proofs,
     })
   }
+
+  /**
+   * Removes association of a content CID with the space. Optionally, also removes
+   * association of CAR shards with space.
```
Gozala, line 317: For the very least I think we should document significant downside this function has.

```suggestion
   * association of CAR shards with space.
   *
   * ⚠️ Please be aware that all shards will be deleted even if there is another upload(s) that
   * reference same shards, which in turn could corrupt those uploads.
```

Gozala (changes requested):
I agree it makes more sense to have this functionality inside a as opposed to just CLI. I do worry however about the data corruption and would love us to do something on the backend to prevent that from happening e.g. instead of `store/remove` we could have `store/unlink` that could remove a link between upload and shard and delete actual CAR only if whet number of links is 0. That would require changing our DB so that shard to upload relations are stored separate from uploads, but that is something we have been considering already due to other issues we encountered with the current design (where vast number of shards can lead to record size limit).

In the immediate future we could probably just add big warning sign around this API and make it a user problem.

I have requested changes, but I do not think we need to re-review this. The change I'm referring to is in the inline comments tldr; lets remove shards first so that on failure user still can retry or if there is a reason not to lets document it in the code so it's clear to the code maintainer.

packages/w3up-client/src/client.js
vasco-santos: There is no real reason, I basically copied what was done in CLI given that functionality is there. I will see who did the change and ask for feedback to align on it

vasco-santos: Ok, so this code was added over a year ago https://github.com/web3-storage/w3cli/pull/20

I guess we would be ok with changing for suggested, but let me ask @olizilla if there is any reason on their head

vasco-santos: Honestly, I think we could consider not facilitating a high level function to client as it is not an atomic op, may have mutations on partial failure, and as pointed out when part of multiple uploads may also be an issue.

Leaving this for the advanced function prefixed by capabilities could be a way to deal with this, but we would need to explain users upload and store, and update the docs (and of course make CLI no delete like this, but also in a granular way)

Gozala:
This inspired me to write up following proposal https://github.com/web3-storage/w3up/issues/1249

alanshaw:
@vasco-santos can you fix this up as @Gozala requested and deploy? We'll redesign the relations later.

vasco-santos:
@alanshaw I am more tempted to move with my latest thoughts:

> https://github.com/web3-storage/w3up/pull/1248#discussion_r1442078525

Remove this function and the docs for it should be updated for low level capabilities until we figure this out. However, maybe CLI should also not enable this directly. What do you think?

packages/w3up-client/README.md
```diff
@@ -594,6 +595,21 @@ function createDelegation (
...
+  contentCID?: CID
+  options: {
+    shards?: boolean
+  } = {}
+): Promise<void>
+```
+
+Removes association of a content CID with the space. Optionally, also removes association of CAR shards with space.
+
+⚠️ Please be aware that all shards will be deleted even if there is another upload(s) that reference same shards, which in turn could corrupt those uploads.
```
alanshaw, line 611:
```suggestion
⚠️ If `shards` option is `true` all shards will be deleted even if there is another upload(s) that reference same shards, which in turn could corrupt those uploads.
```

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,81 @@ export class Client extends Base {
...
+   *
+   * ⚠️ Please be aware that all shards will be deleted even if there is another upload(s) that
```
alanshaw, line 319:
```suggestion
   * ⚠️ If `shards` option is `true` all shards will be deleted even if there is another upload(s) that
```

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,81 @@ export class Client extends Base {
...
+   * @param {object} [options]
+   * @param {boolean} [options.shards]
+   */
+  async remove(contentCID, options = {}) {
+    // Shortcut if there is no request to remove shards
+    if (!options.shards) {
+      // Remove association of content CID with selected space.
+      try {
+        await this.capability.upload.remove(contentCID)
+        /* c8 ignore start */
```
alanshaw, line 332: Where does it stop?

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,81 @@ export class Client extends Base {
...
+          }
+          /* c8 ignore stop */
+        }
+      })
+    )
+
+    // Remove association of content CID with selected space.
+    try {
+      await this.capability.upload.remove(contentCID)
+      /* c8 ignore start */
```
alanshaw, line 386: Where does it stop?

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,81 @@ export class Client extends Base {
...
+    } catch (/** @type {any} */ err) {
+      throw new Error(`remove failed for ${contentCID}: ${err.message ?? err}`)
+    }
```
alanshaw, line 389: Personally I'd wrap a conditional around the shard deleting bit and not repeat this try/catch twice in this same function.

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,81 @@ export class Client extends Base {
...
+          await this.capability.store.remove(shard)
+        } catch (/** @type {any} */ error) {
+          /* c8 ignore start */
+          // If not found, we can tolerate error as it may be a consecutive call for deletion where first failed
+          if (error?.cause?.name !== 'StoreItemNotFound') {
+            throw new Error(
+              `remove failed for shard ${shard.link()} from ${contentCID}: ${
+                error?.message
+              }`
+            )
```
alanshaw, line 376: 🙏 don't smother the stack and the cause.

```suggestion
            throw new Error(`failed to remove shard: ${shard}`, { cause: error })
```

IMHO, only catch errors and re-throw to add information that it is not possible for the caller to know already. This is the perfect case where this needs to happen because the caller won't be able to identify which shard failed to be removed. However we don't need to include `contentCID` in the error message, since that is what the caller asked to be removed. The caller should never rely on context they pass to the function to be part of the error message, since the function may throw for other reasons that weren't annotated this way. So it is on the caller to add their context to the error (or log it with appropriate context).

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,81 @@ export class Client extends Base {
...
+      return
+    }
+
+    // Get shards associated with upload.
+    let upload
+    try {
+      upload = await this.capability.upload.get(contentCID)
+      /* c8 ignore start */
+    } catch (/** @type {any} */ err) {
+      throw new Error(`remove failed for ${contentCID}: ${err.message ?? err}`)
```
alanshaw, line 347: I would just let this throw - we don't have anything to add here that the caller does not already know.

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,81 @@ export class Client extends Base {
...
+
+    // Check if we have information about the shards to remove
+    if (!upload.root) {
+      throw new Error(
+        `⁂ upload not found. could not determine shards to remove.`
+      )
+    }
+    /* c8 ignore start */
+    if (!upload.shards || !upload.shards.length) {
+      return
```
alanshaw, line 359: We still need to remove the upload.

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,81 @@ export class Client extends Base {
...
+    try {
+      upload = await this.capability.upload.get(contentCID)
+      /* c8 ignore start */
+    } catch (/** @type {any} */ err) {
+      throw new Error(`remove failed for ${contentCID}: ${err.message ?? err}`)
+    }
+    /* c8 ignore stop */
+
+    // Check if we have information about the shards to remove
+    if (!upload.root) {
```
alanshaw, line 352: Why would we not have `root`? That field is mandatory no?

packages/w3up-client/src/client.js
```diff
@@ -311,4 +311,81 @@ export class Client extends Base {
...
+        }
+      })
+    )
+
+    // Remove association of content CID with selected space.
+    try {
+      await this.capability.upload.remove(contentCID)
+      /* c8 ignore start */
+    } catch (/** @type {any} */ err) {
+      throw new Error(`remove failed for ${contentCID}: ${err.message ?? err}`)
```
alanshaw, line 388: Again, I would just let this throw - we don't have anything to add here that the caller does not already know.

alanshaw (changes requested):

packages/w3up-client/src/client.js
vasco-santos: copied from CLI code, bot looks like not needed. Going to remove

alanshaw (approved):
Thank you ❤️

PR #1252, opened 2024-01-08
feat: add a function to verify and return Abilities.

travis:
Given a list of strings representing capability names (`W3UpAbility`s), verify that all the strings are valid `W3UpAbility`s and return `W3UpAbility[]`.

`W3UpAbility[]` is still just a list of strings, but this helps us play nice with Typescript.

Inspired by https://github.com/web3-storage/w3up/issues/1250

### DIFF
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -6,6 +6,7 @@ import {
 import {
   Store as StoreCapabilities,
   Upload as UploadCapabilities,
+  abilitiesAsStrings,
 } from '@web3-storage/capabilities'
 import { CAR } from '@ucanto/transport'
 import { Base } from './base.js'
@@ -52,6 +53,27 @@ export class Client extends Base {
       usage: new UsageClient(agentData, options),
     }
     this.coupon = new CouponAPI(agentData, options)
+  }
+
+  /**
+   * Verify and return Abilities.
+   * 
+   * Given a list of strings representing capability names (Abilities), 
+   * verify that all the strings are valid Abilities and return Abilities[].
+   * 
+   * Abilities[] is still just a list of strings, but this helps us play
+   * nice with Typescript.
+   * 
+   * @param {string[]} abilities
+   * @returns {import('@web3-storage/access').Abilities[]}
+   */
+  static abilities(abilities) {
+    for (const ability of abilities) {
+      if (!abilitiesAsStrings.includes(/** @type {import('@web3-storage/access').Abilities} */(ability))) {
+        throw new Error(`${ability} is not a supported capability`)
+      }
+    }
+    return /** @type {import('@web3-storage/access').Abilities[]} */(abilities)
   }
 
   did() {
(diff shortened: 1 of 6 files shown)


packages/w3up-client/src/client.js
```diff
@@ -54,6 +55,31 @@ export class Client extends Base {
...
+   * Abilities[] is still just a list of strings, but this helps us play
+   * nice with Typescript.
+   *
+   * @param {string[]} abilities
+   * @returns {import('@web3-storage/access').Abilities[]}
+   */
+  static abilities(abilities) {
+    for (const ability of abilities) {
+      if (
+        !abilitiesAsStrings.includes(
```
alanshaw, line 73: Might be slightly better to put in a set and use `.has(ability)`...

packages/w3up-client/src/client.js
```diff
@@ -54,6 +55,31 @@ export class Client extends Base {
...
+          /** @type {import('@web3-storage/access').Abilities} */ (ability)
```
alanshaw, line 74: Side note: I'm finding it super confusing that the type of a single ability is called "Abilities"...

packages/w3up-client/test/client.test.js
```diff
@@ -480,4 +480,17 @@ describe('Client', () => {
...
   })
+
+  describe('abilities', () => {
+    it('should return the passed argument if all abilities are valid', async () => {
+      const abilities = ['store/add', 'upload/add']
+      assert.equal(Client.abilities(abilities), abilities)
+    })
+
+    it('should throw an error if one of the abilities is not supported', async () => {
+      assert.throws(() => {
```
alanshaw, line 491: what does it throws? - maybe add an assertion on the message.

packages/w3up-client/src/client.js
```diff
@@ -54,6 +55,31 @@ export class Client extends Base {
...
+   * Given a list of strings representing capability names (Abilities),
+   * verify that all the strings are valid Abilities and return Abilities[].
+   *
+   * Abilities[] is still just a list of strings, but this helps us play
+   * nice with Typescript.
+   *
+   * @param {string[]} abilities
+   * @returns {import('@web3-storage/access').Abilities[]}
+   */
+  static abilities(abilities) {
```
alanshaw, line 70: I'd export this as a function instead: `asAbilities(abilities: string[]): Abilities[]`

packages/w3up-client/src/client.js
travis: Yea hard agree - I've expanded the scope of this PR a bit to `s/Abilities/Ability` in the relevant places. Keeping the old names around for now but marking them deprecated to give folks a chance to update.

travis:
Thanks @alanshaw! I've reworked this a bit per your suggestions.

alanshaw (approved):
I think this is fine to merge but I'd probably call it `Ability` or `ServiceAbility` or `KnownAbility` over `W3UpAbility` - putting product name in the type makes me feel uneasy. I'm guessing you didn't just use `Ability` because it conflicts with ucanto type...

travis:
Sure! I figured it was the repo/package name rather than the product name, but I don't have strong feelings - will call it `ServiceAbility`

PR #1253, opened 2024-01-08
feat: introduce capability for changing billing plan

travis:
We'd like to allow users to update their billing plans from console - introduce a new capability to enable this.


gobengo (approved):
lgtm

PR #1258, opened 2024-01-12
feat: change `plan/update` to `plan/set` and use existing `PlansStorage#set` to implement an invocation handler

travis:
We already have a method on the `PlansStorage` interface that has the semantics we want - rename `plan/update` to `plan/set` and add a handler implementation that hands off to the `PlanStorage#set` method.


alanshaw (approved):
LGTM

PR #1259, opened 2024-01-12
fix: avoid duplicates on aggregator buffer concat

vasco-santos:
While we know that some queues/streams we use are "at least 1 delivery", we have barely seen any duplicate (see https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1704731857750279).

However, when we had a considerable amount of failures and retries, looks like the "at least 1 delivery" primitive got more significant and we actually got some repeated items. This tries to decrease likelihood of repeated items. Of course that, if they actually make their way into different aggregates this won't be effective, but same piece being added in same aggregate should not happen anymore.

### DIFF
--- a/packages/filecoin-api/src/aggregator/buffer-reducing.js
+++ b/packages/filecoin-api/src/aggregator/buffer-reducing.js
@@ -232,9 +232,15 @@ export async function getBufferedPieces(bufferPieces, bufferStore) {
   let bufferedPieces = []
   for (const b of getBufferRes) {
     if (b.error) return b
-    bufferedPieces = bufferedPieces.concat(b.ok.buffer.pieces || [])
+    for (const piece of b.ok.buffer.pieces) {
+      const isDuplicate = bufferedPieces.some(
+        existingPiece => existingPiece.piece.equals(piece.piece)
+      )
+      if (!isDuplicate) {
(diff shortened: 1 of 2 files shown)


packages/filecoin-api/src/aggregator/buffer-reducing.js
```diff
@@ -232,9 +232,15 @@ export async function getBufferedPieces(bufferPieces, bufferStore) {
   let bufferedPieces = []
   for (const b of getBufferRes) {
     if (b.error) return b
-    bufferedPieces = bufferedPieces.concat(b.ok.buffer.pieces || [])
+    for (const piece of b.ok.buffer.pieces) {
+      const isDuplicate = bufferedPieces.some(
+        existingPiece => existingPiece.piece.equals(piece.piece)
```
alanshaw, line 237: Maybe add to a `Set` and then use `.has`? - I think that might be quicker/less work - `.equals` does full buffer comparison each time.

alanshaw (approved):

packages/filecoin-api/src/aggregator/buffer-reducing.js
vasco-santos: good call, I think it should be faster indeed

PR #1264, opened 2024-01-17
feat: strictly require nodejs version

alanshaw:
This PR adds `engines` and `engineStrict` props to the `package.json` ensuring users are running the lower bound of what we're currently testing on in CI (v18) when installing.


vasco-santos (approved):
thanks! it is also already documented in README, so we should be able to go

Gozala (approved):

PR #1266, opened 2024-01-17
feat: re-export Store implementations from w3up-client

travis:
It's useful to be able to use these in code that doesn't want a dependency on `access-client`.


alanshaw (commented):
You need to add an export to package.json right?

travis:
I definitely do 😅 thx!

alanshaw (approved):

PR #1276, opened 2024-01-24
fix: allow import memory store in the browser

alanshaw:
`stores.js` was importing code that uses nodejs modules and is not bundleable....

refs https://github.com/web3-storage/w3up/issues/1274#issuecomment-1907561703


travis (approved):
ugh, thanks!!

PR #1278, opened 2024-01-24
feat: add `initialize` method to `PlansStorage`

travis:
We need a way to initialize an account's "plan" with information about the product the user has selected and an billing ID used by the billing system.

Add this and more tests that demonstrate how it is expected to be used.


alanshaw (approved):

travis:
made a small tweak to preserve backwards compatibility - gonna merge but happy to address any concerns about this post-merge! https://github.com/web3-storage/w3up/pull/1278/commits/ce2ff3286949cb297c904ac6964886ab27bf92ac

PR #1279, opened 2024-01-25
feat: store/add handler includes method: "PUT" in upload response

gobengo:
Rationale:
* this makes the response better describe to the client the action that needs to be taken. I was [doing this](https://github.com/web3-storage/migrate-to-w3up/pull/3/commits/380c25e39c28534badd814b7927cd5b0632f7ac7#diff-ce8b41637e82da88d0592206f82d41cd876392ff3d2c63a738b07deded27d2c7R99) and had to hardcode PUT after trying POST.
* if server instructs client on method, it increases the amount of upload targets this protocol can serve, not just those accepting PUT requests, but any http method (which in general can be any string)

### DIFF
--- a/packages/capabilities/src/types.ts
+++ b/packages/capabilities/src/types.ts
@@ -468,6 +468,11 @@ export interface StoreAddSuccessUpload extends StoreAddSuccessResult {
   status: StoreAddSuccessStatusUpload
   url: ToString<URL>
   headers: Record<string, string>
+  /**
+   * http method to use in request to url.
+   * If undefined, try "PUT".
+   */
+  method?: string
 }
 
 export interface StoreRemoveSuccess {
--- a/packages/upload-api/src/store/add.js
+++ b/packages/upload-api/src/store/add.js
@@ -71,7 +71,7 @@ export function storeAddProvider(context) {
       }
     }
 
-    const { url, headers } = await carStoreBucket.createUploadUrl(link, size)
+    const { url, headers, method = 'PUT' } = await carStoreBucket.createUploadUrl(link, size)
     return {
       ok: {
         status: 'upload',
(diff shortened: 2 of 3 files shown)


packages/capabilities/src/types.ts
```diff
@@ -468,6 +468,11 @@ export interface StoreAddSuccessUpload extends StoreAddSuccessResult {
   status: StoreAddSuccessStatusUpload
   url: ToString<URL>
   headers: Record<string, string>
+  /**
+   * http method to use in request to url.
+   * If undefined, try "PUT".
+   */
+  method?: string
```
alanshaw, line 475: We should be explicit.

```suggestion
  /**
   * HTTP method to for the request to `url`.
   */
  method: 'PUT'
```

alanshaw (changes requested):
Needs a PR to the specs repo 🙏

packages/capabilities/src/types.ts
gobengo: I would like the type to allow the server to instruct the client to send any http method, not just 'PUT'. It seems like typing this as only 'PUT' would move away from that goal

gobengo: > We should be explicit.

what's bad about it being explicitly any syntactically valid http method (i.e. something that metches the type of `fetch` `options.method`) vs being explicitly PUT?

alanshaw: Ah, I meant explicit like not optional :) 

I added PUT because that's all we use right now...but any HTTP method I guess is fine for the type...PUT|POST - are there others it could be?

Eianaftyio:
.patch

Eianaftyio:
.patch

alanshaw (approved):

Eianaftyio:
.diff

Eianaftyio (approved):

alanshaw:
Closing as we moved to `blob/add`.

PR #1280, opened 2024-01-26
fix: one more tweak to the `PlanStorage` interface

travis:
Introduce `UnexpectedError` to allow implementations to return errors that weren't anticipated in the interface definition.

More information about this problem here:

https://github.com/web3-storage/RFC/pull/7#issuecomment-1911245054

Also in this PR:

1) export plan.js as a module so callers can use error classes defined there 
2) use `prettier` to make things prettier

### DIFF
--- a/packages/capabilities/src/types.ts
+++ b/packages/capabilities/src/types.ts
@@ -55,6 +55,7 @@ export type SpaceDID = DID<'key'>
  */
 export interface UnexpectedError extends Ucanto.Failure {
   name: 'UnexpectedError'
+  cause: unknown
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/capabilities/src/types.ts
```diff
@@ -49,6 +49,14 @@ export type CARLink = Link<unknown, typeof CAR.codec.code>
 export type AccountDID = DID<'mailto'>
 export type SpaceDID = DID<'key'>
 
+/**
+ * Error for cases where an interface implementation needs to return an
+ * error that isn't defined explicitly in the interface.
+ */
+export interface UnexpectedError extends Ucanto.Failure {
+  name: 'UnexpectedError'
```
Gozala, line 57:
```suggestion
  name: 'UnexpectedError'
  cause: unknown
```

I would make a slot for the caught error whatever it may be.

Gozala (approved):
I'm not sure `UnexpectedError` is the right name, but I don't have a better one so this is good enough

alanshaw (approved):

PR #1295, opened 2024-01-31
fix: upload client does not need cors mode on presigned url put

vasco-santos:
Apparently some fetch implementations may not have "mode", which is the case for Cloudflare Workers for instance. CF worker deploying upload-client is now failing because this "mode" is set. You can also read more about the general problem in https://github.com/aws/aws-sdk-js-v3/issues/1165

In summary, "mode" may not be implemented in fetch, but the default appears to be cors anyway so this could be deleted without loss of functionality: https://developer.mozilla.org/en-US/docs/Web/API/Request/mode#default_mode.

Related to https://github.com/web3-storage/w3up/issues/1273

### DIFF
--- a/packages/upload-client/src/store.js
+++ b/packages/upload-client/src/store.js
@@ -98,7 +98,6 @@ export async function add(
       try {
         const res = await fetchWithUploadProgress(responseAddUpload.url, {
           method: 'PUT',
-          mode: 'cors',
           body: car,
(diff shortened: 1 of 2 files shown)


alanshaw (approved):

alanshaw:
Maybe move it to web3-storage org so you get some help with maintenance 😊

PR #1301, opened 2024-02-06
feat: add support to prepend pieces while buffering to aggregate

vasco-santos:
We may need to at the implementation level inject pieces in the start of the Aggregate to overcome requirements of Boost when using PoDSI-formatted aggregate, where first bytes can be problematic for SP to grab in the start of the aggregate

This allows implementer to easily pass an array within the lambda context to run, and easily drop it when needed without being hardcoded in the actual lib :)

### DIFF
--- a/packages/filecoin-api/src/aggregator/buffer-reducing.js
+++ b/packages/filecoin-api/src/aggregator/buffer-reducing.js
@@ -159,12 +159,8 @@ export async function handleBufferReducingWithoutAggregate({
  * @param {BufferedPiece[]} [config.prependBufferedPieces]
  */
 export function aggregatePieces(bufferedPieces, config) {
-  const transformedBufferedPieces = [
-    ...(config.prependBufferedPieces || []),
-    ...bufferedPieces
-  ]
   // Guarantee buffered pieces total size is bigger than the minimum utilization
-  const bufferUtilizationSize = transformedBufferedPieces.reduce((total, p) => {
+  const bufferUtilizationSize = bufferedPieces.reduce((total, p) => {
     const piece = Piece.fromLink(p.piece)
     total += piece.size
     return total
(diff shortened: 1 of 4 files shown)


packages/filecoin-api/src/aggregator/buffer-reducing.js
```diff
@@ -182,7 +187,7 @@ export function aggregatePieces(bufferedPieces, sizes) {
   /** @type {BufferedPiece[]} */
   const remainingBufferedPieces = []
 
-  for (const bufferedPiece of bufferedPieces) {
+  for (const bufferedPiece of transformedBufferedPieces) {
     const p = Piece.fromLink(bufferedPiece.piece)
     if (builder.estimate(p).error) {
       remainingBufferedPieces.push(bufferedPiece)
```
alanshaw, line 193: If you somehow configure big piece(s) to prepend that exceed `maxAggregateSize`, then `remainingBufferedPieces` might end up with a configured prepend piece 😱 .

Perhaps we should not concat `prependBufferedPieces` with `bufferedPieces`, and instead iterate over them first adding them to `addedBufferedPieces` _before_ iterating over `bufferedPieces`, and just throw if `estimate` errors since we cannot build an aggregate consisting of _only_ prepended pieces.

alanshaw (changes requested):
Interested how we configure this...and do we need to ensure `insertedAt` is a dynamic "now" date so it does not affect metrics (like the oldest piece)?

packages/filecoin-api/src/aggregator/buffer-reducing.js
vasco-santos: ok, that seems reasonable considering other could implement this.

alanshaw (approved):

PR #1305, opened 2024-02-07
feat: non-parallel piece hashing and CAR upload

alanshaw:
I have a gut feeling that synchronous piece hashing is causing store requests to timeout, specifically waiting for headers in node.js. I think they are probably sent, but event loop is too busy to actually receive them and a timeout gets triggered.

This PR removes parallelism around generating piece hashes, sending `store/add` invocation and uploading a CAR.

refs https://github.com/web3-storage/w3up/issues/1290


vasco-santos (approved):
lgtm

gobengo (approved):

PR #1307, opened 2024-02-09
feat!: add inclusion claim to filecoin api

vasco-santos:
BREAKING CHANGE: aggregator events context need assert service

Closes: https://github.com/web3-storage/w3up/issues/1275


olizilla:
> Apologies for the noise from some formatting changes, but this was from running pnpm format which was needed to make the CI happy 🤷🏼‍♂️

in these cases it can help to have a commit in your branch that shows the clean diff with just the changes you want to make, and then a subsequent commit that applies just the formatting change that ci demands, so folks can review the meaningful stuff.

vasco-santos:
> in these cases it can help to have a commit in your branch that shows the clean diff with just the changes you want to make, and then a subsequent commit that applies just the formatting change that ci demands, so folks can review the meaningful stuff.

Typically I would agree with you, if it would have failed in CI and I would fix.

However, that is not really the case here. I always run format before the commit because my own changes would be problematic for it... I could have looked at what files were purely changed by the format (which I didn't, but also would not expect commited things to be bad...). So, the core changes from lint are actually in files I changed.

Anyway, this is not a big difference in the diff, so it would be needed considerable more time reverting specific changes from linting per file than eye balls ignoring the diff on these cases, which github already offers nice way to see that just formatted change. I think overall we have a problem with the linter and we should get rid of prettier because this only happens in this repo

packages/filecoin-api/src/aggregator/events.js
```diff
@@ -324,6 +325,42 @@ export const handleInclusionInsertToIssuePieceAccept = async (
...
+) => {
+  const claimResult = await Assert.inclusion
+    .invoke({
+      issuer: context.assertService.invocationConfig.issuer,
+      audience: context.assertService.invocationConfig.audience,
+      with: context.assertService.invocationConfig.with,
+      nb: {
+        content: record.aggregate,
+        includes: record.piece,
+        proof: record.proof,
```
alanshaw, line 346: What is proof here?

vasco-santos: proof is the CID of the dag-cbor block that we encode the proof on for storing it

vasco-santos: https://us-west-2.console.aws.amazon.com/dynamodbv2/home?region=us-west-2#item-explorer?operation=SCAN&table=prod-w3filecoin-aggregator-inclusion-store the inclusion store is storing it map, to the proof bytes

PR #1310, opened 2024-02-16
feat: store deliver and confirm on upload-api

vasco-santos:
This PR adds `store/deliver` and `store/confirm` capability and implements its handler in the service based on RFC https://github.com/web3-storage/RFC/pull/10 with the queue pattern.

This will allow us to add event hooks as follow up

Also note that client was not implemented due to the current large refactor in the works. So, I will add it once it lands


vasco-santos:
This PR is outdated in favour of blob

PR #1332, opened 2024-03-13
feat: api waits for trigger filecoin pipeline from the client

vasco-santos:
This PR hooks up possibility to compute Piece CID when submit message is received.

For extra context, today w3infra sets `skipFilecoinSubmitQueue` on `filecoin/offer` from users, resulting `filecoin/offer` in a noop that simply links receipts for effects of `filecoin/submit` and `filecoin/accept`. This is due to the API to not care about what was submit by client as it is calculated in bucket event anyway. However, we want to change this behaviour to have filecoin pipeline only triggered on `filecoin/offer`, putting it in the submit queue where validation will now happen.

We can summarise these changes in two units:
- `handleFilecoinSubmitMessage` now reads data from store, computes piece CID for the bytes and checks equality with provided Piece (i.e. bucket event code https://github.com/web3-storage/w3infra/blob/main/filecoin/index.js#L33 now lives in the handler
- `handlePieceInsertToEquivalencyClaim` now exposed in storefront events. This behaviour was [also triggered from bucket event](https://github.com/web3-storage/w3infra/blob/main/filecoin/index.js#L107) by the service. Now we make it a side effect of insertion to piece table, which happens on `handleFilecoinSubmitMessage`

Rollout plan:
- w3infra swaps `skipFilecoinSubmitQueue` to False when releasing with this. For the old bucket we invoke `filecoin/offer` from bucket event, given current client does not do it

Follow ups:
- [ ] `filecoin/offer` from service on old bucket event
- [ ] hook up finding where data is with datastore
- [ ] clean up `skipFilecoinSubmitQueue` option

### DIFF
--- a/packages/filecoin-api/src/storefront/api.ts
+++ b/packages/filecoin-api/src/storefront/api.ts
@@ -18,6 +18,7 @@ import {
 import {
   Store,
   UpdatableAndQueryableStore,
+  StreammableStore,
   Queue,
   ServiceConfig,
 } from '../types.js'
@@ -29,7 +30,7 @@ export type PieceStore = UpdatableAndQueryableStore<
 >
 export type FilecoinSubmitQueue = Queue<FilecoinSubmitMessage>
 export type PieceOfferQueue = Queue<PieceOfferMessage>
-export type DataStore = Store<UnknownLink, AsyncIterable<Uint8Array>>
+export type DataStore = StreammableStore<UnknownLink, Uint8Array>
 export type TaskStore = Store<UnknownLink, Invocation>
 export type ReceiptStore = Store<UnknownLink, Receipt>
 
@@ -123,8 +124,7 @@ export interface ClaimsClientContext {
    */
   claimsService: {
     invocationConfig: ClaimsInvocationConfig
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    connection: ConnectionView<any>
+    connection: ConnectionView<import('@web3-storage/content-claims/server/service/api').Service>
   }
 }
 
--- a/packages/filecoin-api/src/storefront/api.ts
+++ b/packages/filecoin-api/src/storefront/api.ts
@@ -5,6 +5,9 @@ import type {
   Receipt,
   Invocation,
   Failure,
+  DID,
+  Proof,
+  ConnectionView,
 } from '@ucanto/interface'
 import { PieceLink } from '@web3-storage/data-segment'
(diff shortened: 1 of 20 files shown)


packages/filecoin-api/src/storefront/events.js
```diff
@@ -34,7 +38,28 @@ export const handleFilecoinSubmitMessage = async (context, message) => {
     }
   }
 
-  // TODO: verify piece
+  // read and compute piece for content
+  // TODO: needs to be hooked with location claims
+  const contentGetRes = await context.dataStore.get(message.content)
```
vasco-santos, line 43: here we actually need to decide this interface. Being a datastore would be nice to allow implementers to use R2/S3 client kind of things, which would decrease costs and enhance performance. However, probably we do not want the datastore to be hooked with the claims itself, so probably datastore should in the future receive claims for the content?

packages/filecoin-api/src/storefront/api.ts
```diff
@@ -92,6 +98,36 @@ export interface StorefrontClientContext {
...
+}
+
+export interface ClaimsClientContext {
+  /**
+   * Claims own connection to issue claims.
+   */
+  claimsService: {
+    invocationConfig: ClaimsInvocationConfig
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    connection: ConnectionView<any>
```
alanshaw, line 127:
```suggestion
    connection: ConnectionView<import('@web3-storage/content-claims/server/service/api').Service>
```

packages/filecoin-api/src/storefront/events.js
alanshaw: Not sure I understand - the datastore can use content claims under the hood no?

alanshaw: `.get()` implies to me it's retrieving the whole data, perhaps consider calling this `.stream()`? (similar to https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream)

alanshaw (approved):

packages/filecoin-api/src/storefront/events.js
vasco-santos: yeah good point, I will change to stream. 

> Not sure I understand - the datastore can use content claims under the hood no?

yes it can. But probably we would like it to be within the protocol implementation rather than next to the infra wiring? Let's discuss it when we get there

PR #1333, opened 2024-03-20
fix: upload client should perform filecoin offer

vasco-santos:
In context of https://github.com/w3s-project/project-tracking/issues/25 and https://github.com/web3-storage/w3infra/issues/344 we set the client to perform filecoin/offer already so that new users, as well as users that upgrade in the meantime can already perform this, in order to better prepare us to get rid of bucket event

### DIFF
--- a/packages/upload-client/src/index.js
+++ b/packages/upload-client/src/index.js
@@ -138,7 +138,7 @@ async function uploadBlockStream(conf, blocks, options = {}) {
             {
               issuer: conf.issuer,
               audience: conf.audience,
-              // Resource of storefront is the issuer did
+              // Resource of invocation is the issuer did for being self issued
               with: conf.issuer.did(),
               proofs: conf.proofs,
             },
--- a/packages/upload-client/src/index.js
+++ b/packages/upload-client/src/index.js
@@ -148,7 +148,10 @@ async function uploadBlockStream(conf, blocks, options = {}) {
           )
 
           if (result.out.error) {
-            throw new Error('failed to offer piece for aggregation into filecoin deal', { cause: result.out.error })
+            throw new Error(
+              'failed to offer piece for aggregation into filecoin deal',
+              { cause: result.out.error }
+            )
           }
 
           const { version, roots, size } = car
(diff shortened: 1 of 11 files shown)


packages/upload-client/src/types.ts
```diff
@@ -93,7 +94,7 @@ export interface ProgressStatus extends XHRProgressStatus {
 
 export type ProgressFn = (status: ProgressStatus) => void
 
-export interface Service {
+export interface Service extends StorefrontService {
```
vasco-santos, line 97: adds storefront capabilities as implemented by `upload-api` per https://github.com/web3-storage/w3up/blob/main/packages/upload-api/src/types.ts#L164

packages/upload-client/src/index.js
```diff
@@ -129,7 +130,27 @@ async function uploadBlockStream(conf, blocks, options = {}) {
...
+
+          // Invoke store/add and write bytes to write target
           const cid = await Store.add(conf, bytes, options)
+          // Invoke filecoin/offer for data
+          const result = await Storefront.filecoinOffer(
+            {
+              issuer: conf.issuer,
+              audience: conf.audience,
+              // Resource of storefront is the issuer did
+              with: conf.issuer.did(),
```
alanshaw, line 142: Not sure I understand why this is the issuer?

https://github.com/web3-storage/w3up/blob/f5bac9d27610d976d1144ea12a5c6f9448e7974a/packages/capabilities/src/filecoin/storefront.js#L22-L25

lolz looks like I wrote that comment...still why?

packages/upload-client/src/index.js
```diff
@@ -129,7 +130,27 @@ async function uploadBlockStream(conf, blocks, options = {}) {
...
+              proofs: conf.proofs,
+            },
+            cid,
+            piece,
+            options
+          )
+
+          if (result.out.error) {
+            throw result.out.error
```
alanshaw, line 151: I think this is consistent with what we're doing elsewhere in the client:

```suggestion
            throw new Error('failed to offer piece for aggregation into filecoin deal', { cause: result.out.error })
```

alanshaw (approved):
LGTM!

packages/upload-client/src/index.js
vasco-santos: my comment is actually also wrong, the did must be the same as the issuer. Otherwise you would need to either have been delegated such capability or inline proofs given resource is not did of issuer.

Issue #1339, opened 2024-03-22
A stop gap solution for blob implementation until invocation spec


vasco-santos:
See `blob protocol` https://github.com/web3-storage/specs/pull/115

1. In the meantime I suggest we go with blob/* for new capabilities until invocation spec given the current implementation limitation with task/delegations. Just not sure if we also want a temporary prefix for service ones?
2. I think we have a similar pattern as w3-filecoin. Here is where my mind is:
  a. blob/add queues for allocation if content is not already allocated in given space (resource of invocation)
  b. blobAllocateQueue consumer reads blob requests for allocation and self invokes (by w3s) blob/allocate , which if space has enough bytes available will insert in an allocationStore  record with content+space.
3. We store in our store/table (counterpart of allocationStore) this [input](https://github.com/web3-storage/w3up/blob/main/packages/upload-api/src/types.ts#L513-L519). Looking at blob/allocate we are missing issuer and size in the args. We have the taskCid though, which would allow us to find out those two, however I would prefer to make them just part of the param instead. Would you think this is reasonable? I honestly do not know why we store the issuer and the size . Size may be nice to have to query all items in space and get sum of their sizes without HEAD request, and we actually return it in the proposed receipt. So I think that I would def add it
4. I assume we are opting on the implementation to invoke blob/add second time when bytes were written to presigned URL. Can you confirm my assumption?
6. car/multihash checks on migration path
  a. we currently have carpark-prod-0 in R2/S3. When we perform a store/add  we check bytes are there so that we ask users to write things.
  b. we now move to write things by multihash and I am wondering if you have any thoughts on deduping here? Could we simply test multihash with CAR codec and give location claim?
  c. actually same also happen on our storeTable today. We key it by the link…

Gozala:
> I assume we are opting on the implementation to invoke blob/add second time when bytes were written to presigned URL. Can you confirm my assumption?

In terms of implementing `/service/blob/accept` and specifically how does client signal that they've completed upload I suggest that we do whichever is easiest option from following two:

1. Make client sent second `blob/add` invocation after they've done upload so we can perform a lookup.
2. Add another temp capability with empty output either very specific like `blob/add/poll` or very general like `invocation/wake { task: CID } `.

Gozala:
[...]

PR #1340, opened 2024-03-22
feat: blob capability implementation

Gozala:
Just creating this PR so we could discuss implementation details for #1339

### DIFF
--- a/packages/upload-api/src/blob/add.js
+++ b/packages/upload-api/src/blob/add.js
@@ -1,0 +1,70 @@
+import * as Server from '@ucanto/server'
+import * as Blob from '@web3-storage/capabilities/blob'
+import * as API from '../types.js'
+
+/**
+ * @param {API.BlobServiceContext} context
+ * @returns {API.ServiceMethod<API.BlobAdd, API.BlobAddSuccess, API.BlobAddFailure>}
+ */
+export function blobAddProvider(context) {
+  return Server.provideAdvanced({
+    capability: Blob.add,
+    handler: async ({ capability }) => {
+      const { id, allocationStore, maxUploadSize } = context
+      const { content, size } = capability.nb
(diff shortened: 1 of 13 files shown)


packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,70 @@
...
+            issuer: id,
+            audience: id,
+            with: id.toDIDKey(),
+            nb: {
+              content,
+              size,
+            },
+            expiration: Infinity,
+          })
+          .delegate(),
```
Gozala, line 40: I was thinking this one could be actual invocation that we run from here, I do not believe we need to schedule it for later. In my mind it would basically do what we have here


https://github.com/web3-storage/w3up/blob/9aba8a155cca663387952fc9b11bc3748de8f2fc/packages/upload-api/src/store/add.js#L13-L74


Than we would do what you do here. Followed by logic that checks if we already have blob, if so we could also issue receipt for accept. If we do not have it yet then do exactly what you do here.

Gozala: I'm not sure if we have a good way to check if we already have a receipt for this task or not, but assuming we can do that I imagine we'd have a big if clause around all these so if we do have a receipt we just check if blob was received and if so issue receipt for the `accept` if we do not have receipt yet it's a first run so we do what was described prior

vasco-santos:
Oh I just now noticed this. I also created PR https://github.com/web3-storage/w3up/pull/1342 with more iterations. Thanks for the feedback, I will port there

packages/upload-api/src/blob/add.js
vasco-santos: I just noticed this and also created #1342 with this branch with more iterations. I ported these comments over on https://github.com/web3-storage/w3up/pull/1342#pullrequestreview-1957849688

PR #1342, opened 2024-03-25
feat: blob implementation

vasco-santos:
Adds implementation of `blob/*`, `web3.storage/*` and `ucan/conclude` handlers and capabilities.

Also upgrades to latest ucanto all over the place


packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,71 @@
...
+            audience: id,
+            with: id.toDIDKey(),
+            nb: {
+              blob,
+              cause: invocation.link(),
+              space
+            },
+            expiration: Infinity,
+          })
+          .delegate(),
```
vasco-santos, line 41: Grabbing comment from @Gozala on https://github.com/web3-storage/w3up/pull/1340#discussion_r1536145607

I was thinking this one could be actual invocation that we run from here, I do not believe we need to schedule it for later. In my mind it would basically do what we have here


https://github.com/web3-storage/w3up/blob/9aba8a155cca663387952fc9b11bc3748de8f2fc/packages/upload-api/src/store/add.js#L13-L74

Than we would do what you do here. Followed by logic that checks if we already have blob, if so we could also issue receipt for accept. If we do not have it yet then do exactly what you do here.

vasco-santos: Grabbing comment from @Gozala on https://github.com/web3-storage/w3up/pull/1340#discussion_r1536150410

I'm not sure if we have a good way to check if we already have a receipt for this task or not, but assuming we can do that I imagine we'd have a big if clause around all these so if we do have a receipt we just check if blob was received and if so issue receipt for the accept if we do not have receipt yet it's a first run so we do what was described prior

vasco-santos: @Gozala looks like we are actually more closely aligned than I thought based on previous [code snipped shared in spec](https://github.com/web3-storage/specs/pull/115#discussion_r1535858479). This was as I understood an example on more futuristic Invocation spec, and not what we can do today or we would need to block on accept.

1. Starting by the first comment as adopting same strategy we had today in `store/add` handler with invocation right away instead of scheduling it for later, YES we can do as before, but I want to give more details on why I opted for this implementation for your understanding.
  1.1. In the future, I expect us to have extra step that would come in place during the scheduling phase. So, imagine moving from a world that handler has full knowledge right away on where to write (S3, R2 configured/hard coded) to a world where we have a protocol where `blob/*` service interacts with a protocol for discovery of write targets based on a set of heuristics in order to find out a presigned URL for the blob. I would expect this to happen async and in the service schedule, and therefore opting to go this way already.  @Gozala what do you think?
2. Regarding second comment and lookup for receipts, I still want to explore a bit more, but probably worth deciding after we align on first point here

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @param {API.BlobServiceContext} context
+ * @returns {API.ServiceMethod<API.BlobAdd, API.BlobAddSuccess, API.BlobAddFailure>}
+ */
+export function blobAddProvider(context) {
+  return Server.provideAdvanced({
+    capability: Blob.add,
+    handler: async ({ capability, invocation }) => {
+      const { id, allocationStorage, maxUploadSize } = context
```
vasco-santos, line 13: context should have a scheduler

vasco-santos: scheduler impl should check task store before executing

packages/upload-api/src/blob/add.js
Gozala: Me and @vasco-santos talked about this sync, but posting here for future reference.

Generally speaking all tasks should be async by nature, and only reason we had to resort to workarounds in filecoin pipeline because our client expected to get receipt as response. It is probably best if we try to keep working around client implementation detail and just make async receipts possible.

Apart from the client there is also question of the how do we proceed with a handlers on server. I think it would make sense to create generic invocation queue which can be basis for the scheduler. More concretely invocation CARs should be written in S3 (ideally in target space) and invocation CID put into invocation queue. Queue consumer can then

1. Check if there is already a receipt for this invocation, if so ignore and move on.
2. Otherwise fetch invocation, decode and dispatch to an appropriate handler
3. Write receipt into S3 (ideally in target space)

Client will only have to learn to poll receipts if one was not included in the response.

⚠️ It is important to properly handle concurrency however. Specifically queue consumer may check that receipt for the invocation does not exist and decide to run the task at the same time same task may be already running concurrently which would create a race. Which is to say I we need to write something before we run task to prevent this.

vasco-santos: > ⚠️ It is important to properly handle concurrency however. Specifically queue consumer may check that receipt for the invocation does not exist and decide to run the task at the same time same task may be already running concurrently which would create a race. Which is to say I we need to write something before we run task to prevent this.

The task result should be the same right, so it is not critical to check this but an optimization right?

vasco-santos: Note that @Gozala recommended to not do the scheduling and actually call `allocate` within `blob/add` for the time being. We can complicate things when we need a more async write target selection

packages/upload-api/src/blob/allocate.js
```diff
@@ -0,0 +1,100 @@
...
+    return {
+      ok: {
+        size: blob.size,
+        address: {
+          url: createUploadUrl.ok.url.toString(),
+          headers: createUploadUrl.ok.headers,
+        },
+      },
+    }
+  })
```
Gozala, line 99: I would expect allocation to produce same result regardless of whether we have a blob or not, it's just if we do have a blob already we would succeed put without awaiting on user to perform upload and consequently succeed blob/add as well.

packages/upload-api/src/blob/lib.js
```diff
@@ -0,0 +1,56 @@
...
+
+  toJSON() {
+    return {
+      ...super.toJSON(),
+      space: this.space,
+    }
+  }
+}
+
+export const BlobItemSizeExceededName = 'BlobItemSizeExceeded'
```
Gozala, line 32: Perhaps `BlobExceedsSizeLimit` would be more clear.

packages/upload-api/src/blob/lib.js
```diff
@@ -0,0 +1,56 @@
...
+    super()
+    this.maxUploadSize = maxUploadSize
+  }
+
+  get name() {
+    return BlobItemSizeExceededName
+  }
+
+  describe() {
+    return `Maximum size exceeded: ${this.maxUploadSize}, split DAG into smaller shards.`
```
Gozala, line 47:
```suggestion
    return `Blob exceeded maximum size limit: ${this.maxUploadSize}, consider splitting it into blobs that fit limit.`
```

packages/upload-api/src/blob/list.js
```diff
@@ -0,0 +1,15 @@
+import * as Server from '@ucanto/server'
+import * as Blob from '@web3-storage/capabilities/blob'
+import * as API from '../types.js'
+
+/**
+ * @param {API.BlobServiceContext} context
+ * @returns {API.ServiceMethod<API.BlobList, API.BlobListSuccess, API.Failure>}
+ */
+export function blobListProvider(context) {
```
Gozala, line 9: Nit: I have avoided spec-ing blob/list because I'm not sure what should it return exactly. Broadly speaking I've been thinking that all our list operations should be client side and simply provide view over relevant receipts.

packages/upload-api/src/types/blob.ts
```diff
@@ -0,0 +1,75 @@
...
+  RecordKeyConflict,
+  RecordNotFound,
+  ListOptions,
+  ListResponse,
+} from '../types.js'
+import { Storage } from './storage.js'
+
+export type TasksStorage = Storage<UnknownLink, Invocation>
+
+export interface AllocationsStorage {
```
Gozala, line 24: Do we not need sizes anywhere here ?

Gozala (commented):
This is an amazing progress, I'm blown away. I did provide some feedback here and there only thing I'd call out is that we really should capture dependencies across tasks and we'd have to define those in capability schemas also otherwise they will be ignored during encode.

Other than that I was also thinking when writing spec that it would be a good idea to have separate internal invocations when space usage is increased / decreased so that billing could more simply tap into that without having to make guesses if `allocate` resulted in more space use or not. Relatedly there had been various other cases where we want to do batch operations that billing would have trouble with, however it could be also addressed if we had simple internal ops that increase / decreate usage. This is definitely something we need @alanshaw feedback on

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,269 @@
...
+  with: SpaceDID,
+  derives: equalWith,
+})
+
+/**
+ * `web3.storage/blob//allocate` capability can be invoked to create a memory
+ * address where blob content can be written via HTTP PUT request.
+ */
+export const allocate = capability({
+  can: 'web3.storage/blob/allocate',
```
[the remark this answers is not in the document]
vasco-santos: I wanted to put things already in the future place, but I realize that we may need to run both at same time, so I will do these chaanges

packages/capabilities/src/types.ts
```diff
@@ -439,6 +440,81 @@ export interface UploadNotFound extends Ucanto.Failure {
...
+
+// Blob add
+export interface BlobAddSuccess {
+  claim: {
+    'await/ok': Link
+  }
+}
+
+export interface BlobItemSizeExceeded extends Ucanto.Failure {
+  name: 'BlobItemSizeExceeded'
```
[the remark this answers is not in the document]
vasco-santos: the error you mean is from BlobAllocate

packages/capabilities/src/types.ts
```diff
@@ -439,6 +440,81 @@ export interface UploadNotFound extends Ucanto.Failure {
...
+}
+export type BlobAddFailure = BlobItemSizeExceeded | Ucanto.Failure
```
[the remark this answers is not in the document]
vasco-santos: We have a lot of other errors with the Store operations that also need to be included

packages/upload-api/src/blob/allocate.js
```diff
@@ -0,0 +1,100 @@
...
+    // Rate limiting validation
+    const rateLimitResult = await ensureRateLimitAbove(
+      context.rateLimitsStorage,
+      [space],
+      0
+    )
+    if (rateLimitResult.error) {
+      return {
+        error: {
+          name: 'InsufficientStorage',
```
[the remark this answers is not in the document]
vasco-santos: I copied this from `store/allocate`. Could you create an issue to change that in general? I wanted now to just be consistent

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,153 @@
...
+        const messageCar = await CAR.outbound.encode(message)
+        allocateUcanConcludefx = await UCAN.conclude.invoke({
+          issuer: id,
+          audience: id,
+          with: id.toDIDKey(),
+          nb: {
+            bytes: messageCar.body
+          },
+          expiration: Infinity,
+        }).delegate()
```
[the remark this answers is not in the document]
vasco-santos: So my idea on using the Message is that we can rely on the message version when we break things. Also attach would be tricky given there will be multiple blocks and we require block links to be in nb, we cannot just put the root and traverse :(

packages/upload-api/src/blob/allocate.js
vasco-santos: I am following the spec, which makes `address` optional for this reason? https://github.com/web3-storage/specs/blob/feat/http-put/w3-blob.md#allocate-blob-receipt-schema

In the meantime, latest implementation of blob dd should actually do what you say

packages/upload-api/src/types/blob.ts
vasco-santos: the `BlobAddInput` has the size

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,153 @@
...
+        space,
+        blob.content
+      )
+      let allocateUcanConcludefx
+      if (!allocatedExistsRes.ok) {
+        // Execute allocate invocation
+        const allocateRes = await blobAllocate.execute(getServiceConnection())
+        if (allocateRes.out.error) {
+          return {
+            error: allocateRes.out.error,
```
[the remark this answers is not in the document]
vasco-santos: not sure I understood this, but in the meantime I changed this quite a lot. Could u check again and give mroe feedback? 🙏🏼

packages/upload-api/src/blob/accept.js
```diff
@@ -0,0 +1,62 @@
...
+      const digest = new Digest(sha256.code, 32, blob.content, blob.content)
+      const content = createLink(CAR.code, digest)
+      const w3link = `https://w3s.link/ipfs/${content.toString()}`
+
+      // TODO: Set bucket name
+      // TODO: return content commitment
+      const locationClaim = await Assert.location
+        .invoke({
+          issuer: context.id,
+          // TODO: we need user agent DID
```
vasco-santos, line 37: @Gozala I realized now that we do not have a good way to track this... Probably `accept` should also have `cause` CID like `blob/allocate` so that we can know this info.

Any ideas?

Gozala: Yeah I don't mind sticking cause here, but we could also add `space` field like we have in `allocate`

packages/capabilities/src/http.js
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * DID of the (memory) space where Blob is intended to
+   * be stored.
+   */
+  with: SpaceDID,
+  nb: Schema.struct({
+    /**
+     * Blob to allocate on the space.
+     */
+    blob: blobStruct,
```
vasco-santos, line 31: @Gozala note that I may change this to get the size info from `content length` as you suggested

packages/upload-api/src/blob/accept.js
Gozala: I believe `accept` already has dependency on `allocate` via `ucan/await` syntax it just prefixed `_` implying that there is no need to pass an argument to a function. If we just want to capture link in the structure we probably don't need to do anything beyond that. However if we do want link to be passed into arguments then I do agree that we should add `cause`(al) link as you have suggested.

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,76 @@
...
+})
+
+/**
+ * Blob description for being ingested by the service.
+ */
+export const blobStruct = Schema.struct({
+  /**
+   * A multihash digest of the blob payload bytes, uniquely identifying blob.
+   */
+  content: Schema.bytes(),
```
Gozala, line 43:
```suggestion
  digest: Schema.bytes(),
```

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,76 @@
...
+   * be stored.
+   */
+  with: SpaceDID,
+  derives: equalWith,
+})
+
+/**
+ * Blob description for being ingested by the service.
+ */
+export const blobStruct = Schema.struct({
```
Gozala, line 39:
```suggestion
export const content = Schema.struct({
```

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * DID of the (memory) space where Blob is intended to
+   * be stored.
+   */
+  with: SpaceDID,
+  nb: Schema.struct({
+    /**
+     * Blob to allocate on the space.
+     */
+    blob: blobStruct,
```
Gozala, line 69:
```suggestion
    blob: content,
```

packages/capabilities/src/http.js
Gozala:
```suggestion
    content,
```

packages/capabilities/src/http.js
```diff
@@ -0,0 +1,55 @@
...
+    address: Schema.struct({
+      /**
+       * HTTP(S) location that can receive blob content via HTTP PUT request.
+       */
+      url: Schema.string(),
+      /**
+       * HTTP headers.
+       */
+      headers: Schema.unknown(),
+    }).optional(),
```
Gozala, line 44: It think we could change it to following

```suggestion
  nb: Schema.struct({
    /**
     * HTTP(S) location that can receive blob content via HTTP PUT request.
     */
    url: Schema.string(),
    /**
     * HTTP headers.
     */
    headers: Schema.dictionary({ value: Schema.string() }),
    /**
     * Body of the HTTP PUT request
     */
    body: content,
```

packages/upload-api/src/blob/add.js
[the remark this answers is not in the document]
Gozala: What I was trying to say is that I do not think we should just propagate error as is, instead we should wrap it into something like

```js
class AwaitError extends Ucanto.Failure {
   name = 'AwaitError'
   /**
    * @param {object} source
    * @param {string} source.at - argument path that referenced failed `await`
    * @param {[selector: string, task: API.Link]} source.reference - awaited reference that failed
    * @param {API.Failure} source.cause - error that caused referenced `await` to fail
    */
   constructor({ at, reference, cause }) {
        super()
        this.at = at
        this.reference = reference
        this.cause = cause
   }
   describe() {
      const [selector, task] = this.reference
      return `Awaited (${selector} ${task}) reference at ${at} has failed:\n${this.cause}`
   }
}
```

Then here I'd expect

```js
return { error: AwaitError({ cause: allocateRes.out.error, at, reference }) }
```

My other comment regarding spec was that I think exact shape and structure of `AwaitError` should probably described by the ucan invocation spec.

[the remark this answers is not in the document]
Gozala: Created issue for the spec https://github.com/ucan-wg/invocation/issues/34

packages/capabilities/src/blob.js
vasco-santos: Per the spec this is  `content` https://github.com/web3-storage/specs/blob/main/w3-blob.md#blob-content. Would you prefer to change it?

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,216 @@
...
+        // If this is a new allocation, `http/put` effect should be returned with address
+        blobAllocateOutAddress = allocateRes.out.ok.address
+        blobAllocateReceipt = allocateRes
+      }
+
+      // Create `blob/allocate` receipt invocation to inline as effect
+      const message = await Message.build({ receipts: [blobAllocateReceipt] })
+      const messageCar = await CAR.outbound.encode(message)
+      const bytes = new Uint8Array(messageCar.body)
+      const messageLink = await CAR.codec.link(bytes)
```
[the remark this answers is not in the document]
vasco-santos: I already raised this somewhere else, but can't find the comment now. We can only attach blocks that have a `link` in the nb field. Doing what you suggest below with iterating through receipt blocks leads to 
```
HandlerExecutionError: service handler {can: "blob/add"} error: given block with bafyreieaipb7idctqbfmgzpb4mfjy7msgmncng7qieiqssbshv5v5z7hci is not an attached link
```

Moreover, the message allows us to easily handle breaking changes. So, while I started to use it as a workaround for the above, I now actually give it the value of message version. What do you think?

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,216 @@
...
+          return receiptGet
+        }
+        blobAllocateReceipt = receiptGet.ok
+
+        // Check if despite allocated, the blob is still not stored
+        const blobHasRes = await context.blobsStorage.has(blob.content)
+        if (blobHasRes.error) {
+          return blobHasRes
+        } else if (!blobHasRes.ok) {
+          // @ts-expect-error receipt type is unknown
```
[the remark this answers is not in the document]
vasco-santos: No the logic here is correct. If we do not have a Blob stored (`!blobHasRes.ok`) we give the `http/put` effect to signal the client to send the bytes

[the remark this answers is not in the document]
vasco-santos: I added a comment there

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,216 @@
...
+        result
+          // 1. System allocated memory in user space for the blob.
+          .fork(allocatefx)
+          .fork(allocateUcanConcludefx)
+          // 3. System will attempt to accept uploaded content that matches blob
+          // multihash and size.
+          .join(acceptfx)
+      )
+    },
+  })
```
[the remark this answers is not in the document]
vasco-santos: there is also other unknown, the case where other user already allocated and wrote, but you did not.

For me until now, I always expected the `receipt.out` to need to be the same, but thought the effects could vary (like in this case)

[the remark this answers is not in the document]
vasco-santos: also we cannot attach blocks if there is no link for it in the `nb's`. I think we are getting extremely complex with this flow now and I would much rather simplify instead of keeping more and more complexity

packages/capabilities/src/blob.js
Gozala: Yeah I was thinking of changing it to align with [`Content-Digest` header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-digest-headers)

packages/upload-api/src/blob/add.js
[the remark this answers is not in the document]
Gozala: > Moreover, the message allows us to easily handle breaking changes

I think having two version of message one outer and one inner is probably going to be more pain than help, however that is just an intuition. I'm not opposed to it, but would prefer to avoid if we can.

Per our call today, I'm going to see if I can lift the limitations in ucanto to make this work

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,153 @@
...
+          space,
+        },
+        expiration: Infinity,
+      })
+      const blobPut = Blob.put.invoke({
+        issuer: putSubject,
+        audience: putSubject,
+        with: putSubject.toDIDKey(),
+        nb: {
+          content: blob.content,
```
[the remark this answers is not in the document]
Gozala: > @Gozala I thought this was future syntax that represents what we do today with effects. 

Awaits were part of the invocation pipeline, however they only supported `ok` / `error` branches. We have not implemented support for those and in new version we're making `await` more general to allow specifying deeper path. Since we have not implemented old version of await I suggest we go ahead with new syntax, but it's ok if you think we should do the old one instead.

> So, we should add this to the capability definitions right?

Correct we need to add it to the capability definition otherwise fields will be stripped out on encode.

> are they basicaly the same, shouldn't I just want the allocate?

I think you are right we already have dependency with `address` so we should not need another one in form of `_allocate`.

packages/capabilities/src/blob.js
Gozala: I was working on this https://github.com/web3-storage/specs/pull/118/files and forgot to mention

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,216 @@
...
+      const bytes = new Uint8Array(messageCar.body)
+      const messageLink = await CAR.codec.link(bytes)
+
+      const allocateUcanConcludefx = await UCAN.conclude
+        .invoke({
+          issuer: id,
+          audience: id,
+          with: id.toDIDKey(),
+          nb: {
+            message: messageLink,
```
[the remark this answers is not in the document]
Gozala: Related to https://github.com/web3-storage/ucanto/pull/351 you could also do `blobAllocateReceipt.root.cid` instead here. That said I do like the change in ucanto.

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * DID of the (memory) space where Blob is intended to
+   * be stored.
+   */
+  with: SpaceDID,
+  nb: Schema.struct({
+    /**
+     * Blob to allocate on the space.
+     */
+    blob: content,
```
alanshaw, line 69: Why not `nb: content`?

packages/capabilities/src/http.js
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * DID of the (memory) space where Blob is intended to
+   * be stored.
+   */
+  with: SpaceDID,
+  nb: Schema.struct({
+    /**
+     * BodyBlob to allocate on the space.
+     */
+    body: content,
```
alanshaw, line 31: Personally I'd just repeat the fields here so this doesn't depend on blob capability.

packages/capabilities/src/web3.storage/blob.js
```diff
@@ -0,0 +1,109 @@
...
+  with: SpaceDID,
+  derives: equalWith,
+})
+
+/**
+ * `web3.storage/blob//allocate` capability can be invoked to create a memory
+ * address where blob content can be written via HTTP PUT request.
+ */
+export const allocate = capability({
+  can: 'web3.storage/blob/allocate',
```
alanshaw, line 35: Why are these prefixed with `web3.storage/`?

packages/capabilities/src/types.ts
```diff
@@ -578,6 +655,15 @@ export type UCANRevokeFailure =
   | UnauthorizedRevocation
   | RevocationsStoreFailure
 
+/**
+ * Error is raised when receipt is received for unknown invocation
+ */
+export interface InvocationNotFoundForReceipt extends Ucanto.Failure {
+  name: 'InvocationNotFoundForReceipt'
```
alanshaw, line 662:
```suggestion
export interface ReceiptInvocationNotFound extends Ucanto.Failure {
  name: 'ReceiptInvocationNotFound'
```

packages/capabilities/src/types.ts
```diff
@@ -439,6 +442,77 @@ export interface UploadNotFound extends Ucanto.Failure {
...
+  address?: BlobAddress
+}
+
+export interface BlobAddress {
+  url: ToString<URL>
+  headers: Record<string, string>
+}
+
+export interface BlobItemNotFound extends Ucanto.Failure {
+  name: 'BlobItemNotFound'
```
alanshaw, line 496:
```suggestion
export interface BlobNotFound extends Ucanto.Failure {
  name: 'BlobNotFound'
```

packages/capabilities/src/types.ts
```diff
@@ -439,6 +442,77 @@ export interface UploadNotFound extends Ucanto.Failure {
...
+}
+
+export interface BlobNotAllocableToSpace extends Ucanto.Failure {
```
alanshaw, line 499: Can you leave a comment to explain what circumstances cause this error to happen?

packages/capabilities/src/web3.storage/blob.js
vasco-santos: https://github.com/web3-storage/w3up/issues/1339#issuecomment-2015560570

packages/capabilities/src/web3.storage/blob.js
```diff
@@ -0,0 +1,109 @@
...
+  with: Schema.did(),
+  nb: Schema.struct({
+    /**
+     * Blob to accept.
+     */
+    blob: content,
+    /**
+     * Expiration of location site.
+     */
+    exp: Schema.integer(),
```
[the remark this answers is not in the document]
vasco-santos: I do not understand exactly this question. This is related to the expiration that you want for the content claim you get back.
Also, probably better to not tie the implementation with this, this follows the spec, so I would recommend to discuss this at the spec level via issue/PR rather than here

packages/capabilities/src/http.js
vasco-santos: probably better to not tie the implementation with this, this follows the spec, so I would recommend to discuss this at the spec level via issue/PR rather than here... see https://github.com/web3-storage/specs/pull/118

packages/capabilities/src/blob.js
vasco-santos: probably better to not tie the implementation with this, this follows the spec, so I would recommend to discuss this at the spec level via issue/PR rather than here... see https://github.com/web3-storage/specs/pull/118

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * DID of the (memory) space where Blob is intended to
+   * be stored.
+   */
+  with: SpaceDID,
+  nb: Schema.struct({
+    /**
+     * Blob to allocate on the space.
+     */
+    blob: blobModel,
```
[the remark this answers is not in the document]
vasco-santos: probably better to not tie the implementation with this, this follows the spec, so I would recommend to discuss this at the spec level via issue/PR rather than here... see https://github.com/web3-storage/specs/pull/118

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,216 @@
...
+        const receiptGet = await context.receiptsStorage.get(allocatefx.link())
+        if (receiptGet.error) {
+          return receiptGet
+        }
+        blobAllocateReceipt = receiptGet.ok
+
+        // Check if despite allocated, the blob is still not stored
+        const blobHasRes = await context.blobsStorage.has(blob.content)
+        if (blobHasRes.error) {
+          return blobHasRes
```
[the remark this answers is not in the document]
vasco-santos: @Gozala any suggestion on how to track if expired? 

in the URL it is encoded a `X-Amz-Date=20240409T155511Z&X-Amz-Expires=86400`, but I would like to avoid requiring this to be tied to a provider. Also, I would like to try to only create presigned URL if it allocates, which also makes it difficult to just store the state in allocation table. I wonder then, if it could be an option to stick it in the receipt, but also annoying to not be tied to the presigned url encoding. Also AWS presigned URL do not allow us to specify an expiration date, but the duration after presigned URL is created.

What do you think?

packages/capabilities/src/blob.js
Gozala: I agree with @vasco-santos on discussing this in spec, however since initial spec was merged I'll respond here. I think we'd like to maintain some flexibility to extend capability over time, e.g. I anticipate that we may want to allow setting `expiry` and perhaps target `region` in the future. In this instance if we made `nb: content` it would made those extensions more difficult.

Relatedly I'm also exploring thread of making `content` into a multihash that captures payload size per https://github.com/multiformats/multihash/issues/163 which is also why we want to treat `content` as a unit as opposed to set of fields.

packages/capabilities/src/http.js
Gozala: Moved this to https://github.com/web3-storage/specs/pull/118/files#r1558328467

packages/capabilities/src/web3.storage/blob.js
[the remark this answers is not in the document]
Gozala: Idea here is that you tell service what do you want content location commitment expiration to be, which that will be encoded as expiry of the issued location claim. We should not infer desired expiry of commitment from the `blob/accept` invocation expiry because later tells time frame within which invocation can be exercised and not how long you want it's result to be valid for.

We could probably drop this for now as I think we decided that we default to non-expiring commitments. 

@vasco-santos we also should probably either make this field optional or allow non-expring commitments by allowing `null` here.

packages/capabilities/package.json
```diff
@@ -56,6 +56,10 @@
       "types": "./dist/src/filecoin/dealer.d.ts",
       "import": "./src/filecoin/dealer.js"
     },
+    "./web3.storage/blob": {
+      "types": "./dist/src/web3.storage/blob.d.ts",
+      "import": "./src/web3.storage/blob.js"
+    },
```
Gozala, line 62: We should probably expose other capabilities also

```suggestion
  },
  "./blob": {
      "types": "./dist/src/blob.d.ts",
      "import": "./src/blob.js"
    },
  "./http": {
      "types": "./dist/src/http.d.ts",
      "import": "./src/http.js"
    },
```

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Blob description for being ingested by the service.
+ */
+export const content = Schema.struct({
+  /**
+   * A multihash digest of the blob payload bytes, uniquely identifying blob.
+   */
+  digest: Schema.bytes(),
+  /**
+   * Size of the Blob file to be stored. Service will provision write target
```
Gozala, line 45:
```suggestion
   * Number of bytes contained by this blob. Service will provision write target
```

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,76 @@
...
+   * for this exact size. Attempt to write a larger Blob file will fail.
+   */
+  size: Schema.integer(),
+})
+
+/**
+ * `blob/add` capability allows agent to store a Blob into a (memory) space
+ * identified by did:key in the `with` field. Agent must precompute Blob locally
+ * and provide it's multihash and size using `nb.blob` field, allowing
```
Gozala, line 54:
```suggestion
 * identified by did:key in the `with` field. Agent should compute blob multihash
 * and size and provide it under `nb.blob` field, allowing
```

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,76 @@
...
+ * a service to provision a write location for the agent to PUT or POST desired
```
Gozala, line 55: Do we actually  support `POST` if so perhaps we should consider changing `http/put` to something like `/http/request` and stick `method: 'POST' | 'PUT'` field in there ?

packages/capabilities/src/blob.js
```diff
@@ -0,0 +1,76 @@
...
+export const add = capability({
+  can: 'blob/add',
+  /**
+   * DID of the (memory) space where Blob is intended to
+   * be stored.
+   */
+  with: SpaceDID,
+  nb: Schema.struct({
+    /**
+     * Blob to allocate on the space.
```
Gozala, line 67:
```suggestion
     * Blob to be added on the space.
```

packages/capabilities/src/http.js
```diff
@@ -0,0 +1,50 @@
...
+ * @module
+ */
+import { capability, Schema, ok } from '@ucanto/validator'
+import { content } from './blob.js'
+import { equal, equalBody, equalWith, SpaceDID, Await, and } from './utils.js'
+
+/**
+ * `http/put` capability invocation MAY be performed by any agent on behalf of the subject.
+ * The `blob/add` provider MUST add `/http/put` effect and capture private key of the
+ * `subject` in the `meta` field so that any agent could perform it.
```
Gozala, line 18: I suggest we remove reference to `blob/add` here as it introduces additional context which I'm not sure helps here.

```suggestion
 * `http/put` capability invocation MAY be performed by any authorized agent on behalf of the subject
 * as long as they have referenced `body` content to do so.
```

packages/capabilities/src/types.ts
```diff
@@ -439,6 +442,99 @@ export interface UploadNotFound extends Ucanto.Failure {
...
+  digest: BlobMultihash
+  size: number
+}
+
+// Blob add
+export interface BlobAddSuccess {
+  site: {
+    'ucan/await': ['.out.ok.site', Link]
+  }
+}
```
Gozala, line 466: I suggest to define a type for await as I think we'll keep seeing it more

```suggestion
export interface UCANAwait<Selector extends string = string, Task extends unknown = unknown> {
  'ucan/await': [Selector, Link<Task>]
}

// Blob add
export interface BlobAddSuccess {
  site: UCANAwait<'.out.ok.site'>
}
```

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,376 @@
...
+    audience: context.id,
+    with: context.id.did(),
+    nb: {
+      blob,
+      cause: cause,
+      space,
+    },
+    expiration: Infinity,
+  })
+  const allocatefx = await allocate.delegate()
```
Gozala, line 319: I don't think you need both `allocate` and `allocate.delegate()` because you could do `connection.execute(allocatefx)` instead of `allocate.execute(connection)`.

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,376 @@
...
+      // Schedule allocate
+      const scheduleAllocateRes = await scheduleAllocate({
+        context,
+        blob,
+        allocate: next.allocate,
+        allocatefx: next.allocatefx,
+      })
+      if (scheduleAllocateRes.error) {
+        return scheduleAllocateRes
+      }
```
Gozala, line 49: Nit: I think it would be easier to follow what's going on if instead you refactored code into following three functions

```ts
const accept = (context, { blob, space, cause }): { task: Invocation, receipt?: Receipt }
```

And same for the `put` and `allocate`. Specifically I'm suggesting to combine invocation creation and with what `schedule` functions do and return back invocation and receipt when possible.

packages/upload-api/src/blob/allocate.js
```diff
@@ -0,0 +1,106 @@
...
+      let size = blob.size
+
+      // Rate limiting validation
+      // TODO: we should not produce rate limit error but rather suspend / queue task to be run after enforcing a limit without erroring
+      const rateLimitResult = await ensureRateLimitAbove(
+        context.rateLimitsStorage,
+        [space],
+        0
+      )
+      if (rateLimitResult.error) {
```
Gozala, line 24: I find it very confusing that rateLimitResult provides some error but we throw it away and then say space is blocked. I'm not sure why do we conclude space is blocked or why do we discard error that rate limit check produced, should not that error be more accurate in describing what happened ?

packages/upload-api/src/blob/allocate.js
```diff
@@ -0,0 +1,106 @@
...
+        return {
+          error: {
+            name: 'RateLimited',
+            message: `${space} is blocked`,
+          },
+        }
+      }
+
+      // Has Storage provider validation
```
Gozala, line 33:
```suggestion
      // We check if space has storage provider associated. If it does not
      // we return `InsufficientStorage` error as storage capacity is considered
      // to be 0.
```

packages/upload-api/src/blob/allocate.js
```diff
@@ -0,0 +1,106 @@
...
+        return {
+          /** @type {API.AllocationError} */
+          error: {
+            name: 'InsufficientStorage',
+            message: `${space} has no storage provider`,
+          },
+        }
+      }
+
+      // Allocate in space, ignoring if already allocated
```
Gozala, line 48:
```suggestion
      // Allocate memory space for the blob. If memory for this blob is
      // already allocated, this allocates 0 bytes.
```

packages/upload-api/src/blob/allocate.js
```diff
@@ -0,0 +1,106 @@
...
+            error: allocationInsert.error,
+          }
+        }
+      }
+
+      // Check if blob already exists
+      const hasBlobStore = await context.blobsStorage.has(blob.digest)
+      if (hasBlobStore.error) {
+        return hasBlobStore
+      }
```
Gozala, line 71: I don't think we should be checking whether we have a blob during allocation, it should be irrelevant and something we check in the `blob/add` instead.

packages/upload-api/src/blob/allocate.js
```diff
@@ -0,0 +1,106 @@
...
+
+      // If blob is stored, we can just allocate it to the space with the allocated size
+      if (hasBlobStore.ok) {
+        return {
+          ok: { size },
```
Gozala, line 76: We should not return different result type even if we have blob, especially because `http/put` will depend on `.adddress.url` and `.address.headers` which will not exist in this code path.

We could either return existing allocation receipt or create new address despite having the blob. Since client will get all the receipts they will not re-upload and even if they do that will be a noop.

packages/upload-api/src/blob/lib.js
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * @param {import('@ucanto/interface').DID} [space]
+   */
+  constructor(space) {
+    super()
+    this.space = space
+  }
+
+  get name() {
+    return BlobItemNotFoundName
```
Gozala, line 14:
```suggestion
    return 'BlobItemNotFoundName'
```

packages/upload-api/src/blob/lib.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  describe() {
+    if (this.space) {
```
Gozala, line 18: I'm not sure why would we capture space given that we either have blob or not and space is irrelevant.

packages/upload-api/src/blob/lib.js
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * @param {Number} maxUploadSize
+   */
+  constructor(maxUploadSize) {
+    super()
+    this.maxUploadSize = maxUploadSize
+  }
+
+  get name() {
+    return BlobSizeOutsideOfSupportedRangeName
```
Gozala, line 44:
```suggestion
    return 'BlobSizeOutsideOfSupportedRangeName'
```

packages/upload-api/src/blob/lib.js
```diff
@@ -0,0 +1,85 @@
...
+    this.at = at
+    this.reference = reference
+    this.cause = cause
+  }
+  describe() {
+    const [selector, task] = this.reference
+    return `Awaited (${selector} ${task}) reference at ${this.at} has failed:\n${this.cause}`
+  }
+  get name() {
+    return AwaitErrorName
```
Gozala, line 78:
```suggestion
    return 'AwaitErrorName'
```

packages/upload-api/src/types/blob.ts
```diff
@@ -0,0 +1,74 @@
...
+  insert: (
+    item: BlobAddInput
+  ) => Promise<Result<BlobAddOutput, RecordKeyConflict>>
+  list: (
+    space: DID,
+    options?: ListOptions
+  ) => Promise<Result<ListResponse<BlobListItem>, Failure>>
+}
+
+export interface Blob {
```
Gozala, line 34: Don't call it `Blob` please as TS sometimes tries to unify it with builtin `Blob` interface, maybe call it `Content` or `BlobModel` instead.

packages/upload-api/src/types/blob.ts
```diff
@@ -0,0 +1,74 @@
...
+     * The number of seconds before the presigned URL expires
+     */
+    expiresIn: number
+  ) => Promise<
+    Result<
+      {
+        url: URL
+        headers: {
+          'x-amz-checksum-sha256': string
+          'content-length': string
```
Gozala, line 68:
```suggestion
          'x-amz-expires': string
          'content-length': string
```

Should we add this header also ?

packages/upload-api/src/ucan/conclude.js
```diff
@@ -0,0 +1,163 @@
...
+    const scheduleRes = await Promise.all(
+      httpPutTaskGetRes.ok.capabilities
+        // Go through invocation tasks and get all `http/put`
+        .filter((cap) => cap.can === HTTP.put.can)
+        // @ts-expect-error body exists in `http/put` but unknown type here
+        .map(async (/** @type {API.HTTPPut} */ cap) => {
+          // Get triggering task (blob/allocate) by checking blocking task from `url`
+          /** @type {API.UnknownLink} */
+          // @ts-expect-error ts does not know how to get this
+          const blobAllocateTaskCid = cap.nb.url['ucan/await'][1]
```
Gozala, line 50: I think this reads better

```suggestion
          const [,blobAllocateTaskCid] = cap.nb.url['ucan/await']
```

packages/upload-api/src/ucan/conclude.js
```diff
@@ -0,0 +1,163 @@
...
+                space: allocateCapability.nb.space,
+                _put: {
+                  'ucan/await': ['.out.ok', ranInvocation.link()],
+                },
+              },
+              expiration: Infinity,
+            })
+            .delegate()
+
+          return tasksScheduler.schedule(blobAccept)
```
Gozala, line 81: I have not read through the scheduler yet, but I do find it surprising that invocation is scheduled here. I would expect this invocation to be already scheduled / pending and here we would just wake it by `blobAllocateTaskCid`.

packages/upload-api/src/ucan/conclude.js
```diff
@@ -0,0 +1,163 @@
...
+        })
+    )
+
+    const scheduleErrors = scheduleRes.filter((res) => res.error)
+    if (scheduleErrors.length && scheduleErrors[0].error) {
+      return {
+        error: scheduleErrors[0].error,
+      }
+    }
```
Gozala, line 90: Seems like scheduler should not fail the receipt handler but the task been awaken instead. Only time I would expect receipt handler to fail is if:

1. We can not find the task receipt is for.
2. Task receipt has already been received and it is different from this one.

packages/upload-api/src/ucan/conclude.js
```diff
@@ -0,0 +1,163 @@
...
+      ok: { time: Date.now() },
+    }
+  })
+
+/**
+ * @param {import('multiformats').UnknownLink} cid
+ * @param {IterableIterator<import('@ucanto/interface').Block<unknown, number, number, 1>>} blocks
+ * @returns {Promise<import('@ucanto/interface').Result<Uint8Array, DecodeBlockOperationFailed>>}
+ */
+export const findBlock = async (cid, blocks) => {
```
Gozala, line 102: I can't see this function used anywhere, do we need it ?

packages/upload-api/src/ucan/conclude.js
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/**
+ * @param {API.Signer} id
+ * @param {API.Verifier} serviceDid
+ * @param {API.Receipt} receipt
+ */
+export function createConcludeInvocation(id, serviceDid, receipt) {
+  const receiptBlocks = []
+  for (const block of receipt.iterateIPLDBlocks()) {
```
Gozala, line 141: Might be a good idea to stick cids into second array to avoid second iteration in `facts`.

packages/upload-api/src/ucan/conclude.js
```diff
@@ -0,0 +1,163 @@
...
+      receipt: receipt.link(),
+    },
+    expiration: Infinity,
+    facts: [
+      {
+        ...receiptBlocks.map((b) => b.cid),
+      },
+    ],
+  })
+  for (const block of receipt.iterateIPLDBlocks()) {
```
Gozala, line 158: Better to use already collected array to avoid second traversal

```suggestion
  for (const block of receiptBlocks) {
```

packages/capabilities/package.json
vasco-santos: they are exposes because are in top level https://github.com/web3-storage/w3up/blob/feat/blob-implementation/packages/capabilities/package.json#L32

packages/capabilities/src/types.ts
```diff
@@ -439,6 +442,99 @@ export interface UploadNotFound extends Ucanto.Failure {
...
+export interface BlobListItem {
+  blob: BlobModel
+  insertedAt: ISO8601Date
+}
+
+// Blob allocate
+export interface BlobAllocateSuccess {
+  size: number
+  address?: BlobAddress
+}
```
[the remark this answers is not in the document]
vasco-santos: Can we please align this on the spec https://github.com/web3-storage/specs/blob/fix/align-blob-with-impl/w3-blob.md#allocate-blob ?

It is difficult to manage what is expected if we are always diverging from there. For a second invocation with different nonce, I think not creating a presigned url for something is there makes sense, but happy to reconsider

packages/capabilities/src/types.ts
```diff
@@ -439,6 +442,99 @@ export interface UploadNotFound extends Ucanto.Failure {
...
+
+export interface BlobAddress {
+  url: ToString<URL>
+  headers: Record<string, string>
+  expiresAt: ISO8601Date
```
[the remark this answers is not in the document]
vasco-santos: We talked about not making this provider specific yesterday and the limitations that aws gives. see https://github.com/web3-storage/w3up/pull/1342#discussion_r1557906738

packages/upload-api/src/blob/allocate.js
vasco-santos: if we have blob we would not want to create a presigned url / in the future not look for a node where to write this, while we want to allocate it to the user. What do you think?

packages/upload-api/src/ucan/conclude.js
vasco-santos: Ok, I think we are not aligned on what the schedule is. For me it is an async execution of the invocation out of this context. So schedule it to be run, can we move forward this way while we do not have the real scheduler?

packages/upload-api/src/ucan/conclude.js
vasco-santos: old thing, I will remove

packages/capabilities/src/blob.js
vasco-santos: I think is just put, will change

packages/upload-api/src/blob/allocate.js
vasco-santos: I feel we already talked about this. I am just keeping doing what `store` does and not trying to also figure out what this rate limit should do. What is your suggestion? I am honestly feeling it is better to just remove this for now

packages/upload-api/src/blob/allocate.js
vasco-santos: This follow the spec and as you say there will be a receipt in `http/put` that makes this not needed. Also cuts cost on not requesting a new presigned URL/figure out where to write on a decentralized write target.

Anyway, if we want to always return address like I mentioned above spec does not state that https://github.com/web3-storage/specs/blob/fix/align-blob-with-impl/w3-blob.md#allocate-blob

packages/upload-api/src/types/blob.ts
vasco-santos: it is not a date but number of ms to expire. Which is useless, also I wanted to avoid adding AWS specific things for the future decentralized write targets

packages/capabilities/src/web3.storage/blob.js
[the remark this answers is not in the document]
vasco-santos: Ok, I will change this to `ttl` and make it optional for now

packages/capabilities/src/types.ts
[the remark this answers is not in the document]
vasco-santos: I already raised this a week ago https://github.com/web3-storage/w3up/pull/1342#discussion_r1553944549

I would prefer to get this merged and list what we need to improve, rather than blocking iterative approach on this huge piece of work where there is always something else and details in spec never stable

packages/capabilities/src/types.ts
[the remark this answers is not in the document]
vasco-santos: I want to have a date here, otherwise I cannot infer if presigned URL is expired, and therefore allocation needs to run again. TTL is usually how long more available, rather then a date. So, we would be doing the same issue that `X-Amz-Expires` has. We could also have a receipt time to avoid this. Thoughts?

packages/capabilities/src/types.ts
```diff
@@ -439,6 +442,99 @@ export interface UploadNotFound extends Ucanto.Failure {
...
+  name: 'StorageOperationFailed'
+}
+
+export interface RecordNotFound extends Error {
+  name: 'RecordNotFound'
+}
+
+export interface EncodeRecordFailed extends Error {
+  name: 'EncodeRecordFailed'
+}
```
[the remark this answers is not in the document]
vasco-santos: These are errors related to the storage side of things (AllocationStorage and BlobStorage) and how `put` /`get` may fail.

packages/upload-api/src/blob/add.js
vasco-santos: I need the delegate to have a CID for it, the `allocate` does not give it to me

packages/upload-api/src/blob/allocate.js
vasco-santos: I quite adding the rate limit thing. We won't be able to block users (which I think is the reason for this) until we figure this out, but I don't want to block on this

packages/upload-api/src/blob/lib.js
vasco-santos: this is the name variable...

packages/upload-api/src/blob/lib.js
vasco-santos: this is the name variable... I like to use this as a const so that I easily can check in the tests

packages/upload-api/src/blob/lib.js
vasco-santos: this is the name variable... I like to use this as a const so that I easily can check in the tests

packages/upload-api/src/ucan/conclude.js
vasco-santos: yeah, this was my goal but forgot to change

packages/capabilities/src/types.ts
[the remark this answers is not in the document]
Gozala: 1. The whole rename thing was just a thought, I was not implying that we should do it, sorry for not been clear about this. I just wanted to know if you it made sense to you, if so we can go update spec and then implementation as you say here.
2. You are also correct to point out that `address?: BlobAddress` comes from spec. I was honestly surprised looking at the spec now, clearly I have overlooked the issue I have then pointed out in regards to await's. I'll fix the spec and we can change this in the impl as followup.

packages/capabilities/src/types.ts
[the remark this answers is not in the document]
Gozala: Per wikipedia about [TTL](https://en.wikipedia.org/wiki/Time_to_live) 
> Time to live (TTL) or hop limit is a mechanism which limits the lifespan or lifetime of data in a computer or network. TTL may be implemented as a [counter](https://en.wikipedia.org/wiki/Counter_(digital)) or [timestamp](https://en.wikipedia.org/wiki/Timestamp) attached to or embedded in the data.

I mostly have seen timestamps in TTL but in datagram it is indeed more of a lifetime left, so I think you are correct. Relatedly wiki mentions [Expires](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Responses) header and same named cookie so I think you are correct that TTL may not be a good name.


How about renaming it to `expires` just to align with HTTP header and a cookie ? I'm also ok with leaving it as is.

In regards to my comment of not replicating data already in headers, my primary motivation here is to avoid getting two out of sync, but after thinking more about it I think you are correct. We may choose to expire it sooner than what's in the presigned URL so perhaps having two out of sync is not so bad, although having our expiry higher than what's in headers still bad, but I think it's ok.

packages/upload-api/src/blob/accept.js
```diff
@@ -0,0 +1,62 @@
...
+      if (hasBlob.error) {
+        return hasBlob
+      } else if (!hasBlob.ok) {
+        return {
+          error: new AllocatedMemoryHadNotBeenWrittenTo(),
+        }
+      }
+
+      // TODO: we need to support multihash in claims, or specify hardcoded codec
+      const digest = new Digest(sha256.code, 32, blob.digest, blob.digest)
```
Gozala, line 35: This seems incorrect because because bytes and digest in the constructor aren't the same because `bytes` are supposed to be multihash prefixed byte array and and `digest` parameter supposed to be bytes without multihash prefix.

I think you really should just use `decode(blob.digest)` exported by the [digest module](https://github.com/multiformats/js-multiformats/blob/master/src/hashes/digest.ts#L21-L25) it will also automatically figure out hashing algorithm etc...

packages/upload-api/src/blob/add.js
```diff
@@ -0,0 +1,348 @@
...
+        return putRes
+      }
+
+      // Create accept task, get its receipt if available and execute if necessary and ready
+      const acceptRes = await accept({
+        context,
+        blob,
+        space,
+        putTask: putRes.ok.task,
+        putReceipt: putRes.ok.receipt,
```
Gozala, line 60: Nit: perhaps you should just pass `putRes.ok` ?

packages/upload-api/src/ucan/conclude.js
Gozala: > Ok, I think we are not aligned on what the schedule is. For me it is an async execution of the invocation out of this context. 

I mean the same thing, however I am confused because I would expect that `blobAccept` was already invoked before, so I guess we are rerunning it here ? Which is fine but perhaps adding a comment explaining that would help.

>  So schedule it to be run, can we move forward this way while we do not have the real scheduler?

Yeah, I did not meant to imply we need real scheduler was just confused by what was going on there. Assuming above guess is correct I understand how it works now, but again some code comments would really help to demistify

packages/upload-api/src/ucan/conclude.js
```diff
@@ -0,0 +1,151 @@
...
+              audience: id,
+              with: id.toDIDKey(),
+              nb: {
+                blob: cap.nb.body,
+                space: allocateCapability.nb.space,
+                _put: {
+                  'ucan/await': ['.out.ok', ranInvocation.link()],
+                },
+              },
+              expiration: Infinity,
```
Gozala, line 85: I suspect expiry is set to `Infinity` so that CID will come out the same or am I wrong to assume this ? If that is the case indeed can you please add a comment because it's easy to miss and someone may change this without realizing problem it could cause otherwise.

Gozala (approved):
Looks good, lets land and iterate in followups. Only thing that I would like to see address before landing is the way digest is created because currently it is incorrect. Feel free to merge once that is fixed without another review round.

Thanks

PR #1344, opened 2024-03-26
feat!: deprecate issuer

Gozala:
Per https://github.com/web3-storage/w3up/issues/1339#issuecomment-2015852195 this PR deprecates `issuer` field from table record.

### DIFF
--- a/packages/upload-api/src/types.ts
+++ b/packages/upload-api/src/types.ts
@@ -515,7 +515,12 @@ export interface StoreAddInput {
   link: UnknownLink
   size: number
   origin?: UnknownLink
-  issuer: DID
+
+  /**
+   * @deprecated - Issuer of the invocation is irrelevant as long as
+   * they have authorization to invoke on subject `space`.
+   */
+  issuer?: DID
   invocation: UCANLink
 }
 


packages/upload-api/src/types.ts
```diff
@@ -515,7 +515,12 @@ export interface StoreAddInput {
   link: UnknownLink
   size: number
   origin?: UnknownLink
-  issuer: DID
+
+  /**
+   * @deprecated - Issuer of the invocation is irrelevant as long as
+   * they have authorization to invoke on subject `space`.
+   */
+  issuer?: DID
```
alanshaw, line 523: Ideally we'd like tracability with a `cause` field but probably not so bad now we're moving to blob protocol.

alanshaw (approved):

packages/upload-api/src/types.ts
vasco-santos: that would be the invocation right?

alanshaw: Oh, yeah I'm getting mixed up with the upload table. Ignore me!

PR #1370, opened 2024-04-15
fix: export test handlers and storages

vasco-santos:
Need these exported in order to test stores for w3infra


vasco-santos:
@alanshaw merging this in as it is only changing `test` and need this to proceed :)

PR #1373, opened 2024-04-16
feat: Add blob index class Blobdex

gammazero:
Add a new class to represent a blob index. The blob index is a set of multihash, offset, and length values associated with the data stored in a blob.

The blob index in created when a client stores a blob of data. The multihash values in the blob index are indexed by IPNI if opting into IPNI indexing. When a blob index is retrieved with a claims bundle, the client can use the items in the blob index to request specific data items, identified by multihash, from a stored blob.


packages/blobdex/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * addStr adds a new item to this Blobdex with base-58 key.
+   *
+   * @param {string} b58Str
+   * @param {number} offset
+   * @param {number} length
+   */
+  addStr(b58Str, offset, length) {
```
alanshaw, line 39: When might you need this? This seems a little error prone and exposes the internals of the encoding - we're also not doing any validation so could cause errors on read. Personally I'd not to expose this publically.

packages/blobdex/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * add adds a new item to this Blobdex multihash digest bytes key.
+   *
+   * @param {import('multiformats').MultihashDigest} mh
+   * @param {number} offset
+   * @param {number} length
+   */
+  add(mh, offset, length) {
+    const b58str = base58btc.encode(mh.bytes)
+    this.addStr(b58str, offset, length)
```
alanshaw, line 29: Perhaps validate `offset` and `length` are integers and > 0? `Number.isInteger(offset) && offset > 0`

packages/blobdex/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * len numbers the count of items in this Blobdex.
+   *
+   * @returns {number}
+   */
+  len() {
+    return this.#items.size
+  }
```
alanshaw, line 18: FYI you could expose this as a property accessor:

```suggestion
  get len() {
    return this.#items.size
  }
```

Usage:

```js
const idx = new Blobdex()
console.log(idx.len)
```

Maybe consider using `size` since this is more like a set...

packages/blobdex/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+   */
+  addStr(b58Str, offset, length) {
+    this.#items.set(b58Str, [offset, length])
+  }
+
+  /**
+   * pos returns the offset and length of the block specified by multihash.
+   *
+   * @param {import('multiformats').MultihashDigest} digest
+   * @returns {[number, number]}
```
alanshaw, line 47:
```suggestion
   * @returns {[number, number]|undefined}
```

packages/blobdex/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+    return this.#items.get(b58Str)
+  }
+
+  /**
+   * encode returns a sorted map of items in this Blobdex.
+   *
+   * @returns {Map<string, [number, number]>}
+   */
+  encode() {
+    return new Map([...this.#items.entries()].sort())
```
alanshaw, line 70: You can't call `sort` on an array of tuples - well, you can but the result will probably not be what you expect. You need to provide the sorting function to sort by keys.

i.e. `[...this.#items.entries()]` will give you an array like:

```js
[ ['zQ0...', [1, 2]], ['zQ1...', [1, 2]] ]
```

To sort, you'd want to do something like:

```js
[...this.#items.entries()].sort((a, b) => {
  if (a[0] < b[0]) return -1
  if (a[0] > b[0]) return 1
})
```

I also think that you're going to want to pass this to an IPLD encoder, which doesn't support `Map` instances, you need an actual object. So you might want to do something more like this instead:

```js
Object.fromEntries([...this.#items.entries()].sort((a, b) => a[0] < b[0] ? -1 : 1))
```

packages/blobdex/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * Populate this Blobdex with the given items.
+   *
+   * @param {Map<string, [number, number]>} items
+   */
+  decode(items) {
+    this.#items = new Map([...items.entries()])
+  }
```
alanshaw, line 80: I would export static functions that do encoding/decoding and not include as instance methods. I think it's unexpected for decode to mutate an existing index, I'd have it return a new instance, and add each item (to ensure it's valid).

```js
/** @param {Record<string, [number, number]>} data */
export const decode = (data) => {
  const idx = new Blobdex()
  for (const [k, v] of Object.entries(data)) {
    idx.add(k, v)
  }
  return idx
}

/** @param {Blobdex} idx */
export const encode = (idx) => {
  // need to implement an `entries` method...
  return Object.fromEntries([...idx.entries()].sort((a, b) => a[0] < b[0] ? -1 : 1))
}
```

gammazero:
Replaced by PR https://github.com/w3s-project/w3up/pull/1434

PR #1374, opened 2024-04-16
fix: storage operation failed type name instead of store

vasco-santos:
See https://github.com/web3-storage/w3up/blob/main/packages/capabilities/src/types.ts#L530


vasco-santos:
merging given only name change to align with types

PR #1388, opened 2024-04-23
fix: migrate repo

joaosa:
- Fixed CI permissions
- Added a comment so we can trigger a release

### DIFF
--- a/packages/upload-api/src/customer/get.js
+++ b/packages/upload-api/src/customer/get.js
@@ -40,6 +40,9 @@ class UnknownProvider extends Provider.Failure {
     this.name = /** @type {const} */ ('UnknownProvider')
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):
I assume that given this is only NPM package publishing, we only need to test one of the packages flow (i.e. this will only trigger release to `upload-api`). Probably is good, but if you feel safer you can also do small change in each single package

joaosa:
> I assume that given this is only NPM package publishing, we only need to test one of the packages flow (i.e. this will only trigger release to `upload-api`). Probably is good, but if you feel safer you can also do small change in each single package

We should be fine. Most of the package CI jobs are quite similar and self-contained (besides interacting with npm). If one runs correctly others should too.

PR #1409, opened 2024-04-25
fix!: storefront content store rename and separation for test

vasco-santos:
With the aim to facilitate https://github.com/w3s-project/w3up/issues/1349 storefront `dataStore` is renamed to `contentStore` (note alignment of name with the `content` property in Filecoin Pipeline capabilities/message namings). My goal is to make `contentStore.stream()` perform HTTP Request to read thing from Roundabout (if `filecoin/offer` comes with `content` CARCID code, it will try to get a CAR, if `content` comes as RAW, it will try to get RAW).

The main point in this change is to not require `contentStore` to have a `put` method, like `dataStore` was requiring just for testing. It extracts that requirement only to run tests, so that we can implement a `contentStore` as something with just a `stream` function that behind the scenes can perform a HTTP Request to Roundabout instead of being an abstraction on top of a S3 bucket like today.

Note that this is not strictly needed, we could just use this as is and in `dataStore`  implementation and make `put` https://github.com/w3s-project/w3infra/blob/main/filecoin/store/data.js#L39 `throw new Error('not implemented')` and just use an extended `DataStore` class for w3infra testing. But I feel this way is nicer, and also like more new naming as it aligns with everything including content name defined in w3index https://github.com/w3s-project/specs/blob/main/w3-index.md#sharded-dag-index-example

BREAKING CHANGE: dataStore in storefront renamed to contentStore

### DIFF
--- a/packages/filecoin-api/src/storefront/api.ts
+++ b/packages/filecoin-api/src/storefront/api.ts
@@ -194,5 +194,5 @@ export interface ContentStore<RecKey, Rec> {
   /**
    * Gets a record from the store.
    */
-  stream: (key: RecKey) => Promise<Result<AsyncIterable<Rec>, StoreGetError>>
+  stream: (key: RecKey) => Promise<Result<ReadableStream<Rec>, StoreGetError>>
 }
--- a/packages/filecoin-api/src/storefront/api.ts
+++ b/packages/filecoin-api/src/storefront/api.ts
@@ -8,6 +8,7 @@ import type {
   DID,
   Proof,
   ConnectionView,
+  Result,
 } from '@ucanto/interface'
 import { PieceLink } from '@web3-storage/data-segment'
 import {
@@ -18,9 +19,9 @@ import {
 import {
   Store,
   UpdatableAndQueryableStore,
-  StreammableStore,
   Queue,
   ServiceConfig,
+  StoreGetError
 } from '../types.js'
 
 export type PieceStore = UpdatableAndQueryableStore<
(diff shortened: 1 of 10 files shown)


packages/filecoin-api/src/storefront/api.ts
```diff
@@ -189,3 +189,10 @@ export interface PieceOfferMessage {
 export interface DataAggregationProofNotFound extends Failure {
   name: 'DataAggregationProofNotFound'
 }
+
+export interface ContentStore<RecKey, Rec> {
+  /**
+   * Gets a record from the store.
+   */
+  stream: (key: RecKey) => Promise<Result<AsyncIterable<Rec>, StoreGetError>>
```
alanshaw, line 197: Not sure if useful, but `ReadableStream` _is_ async iterable and you get extra utility for controlling flow and pipelining. Also developer familiarity with https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream

```suggestion
  stream: (key: RecKey) => Promise<Result<ReadableStream<Rec>, StoreGetError>>
```

packages/filecoin-api/test/context/store-implementations.js
```diff
@@ -77,7 +77,7 @@ export const getStoreImplementations = (
         return Array.from(items).find((i) => i.ran.link().equals(record))
       },
     }),
-    dataStore: new StreammableStore({
+    contentStore: new StreammableStore({
```
alanshaw, line 80:
```suggestion
    contentStore: new StreamableStore({
```

```suggestion
    contentStore: new StreamingStore({
```

```suggestion
    contentStore: new ReadableStreamStore({
```

packages/filecoin-api/test/types.ts
```diff
@@ -53,3 +60,10 @@ export interface StorefrontTestEventsContext
     >
   }>
 }
+
+export interface TestContentStore<RecKey, Rec> extends StorefrontInterface.ContentStore<RecKey, Rec> {
+  /**
+   * Puts a record in the store.
+   */
+  put: (record: Rec) => Promise<Result<Unit, StorePutError>>
```
alanshaw, line 68: yes this is the correct way 😄

alanshaw (approved):
I'm not sure about double m in streamable. I made a few suggestions. Otherwise this LGTM.

PR #1410, opened 2024-04-25
feat(capabilities)!: add `index/add` capability

alanshaw:
Adds `index/add` capability.

refs https://github.com/w3s-project/specs/blob/main/w3-index.md

BREAKING CHANGE: `BlobMultihash` type in `@web3-storage/capabilities` renamed to `Multihash`.

### DIFF
--- a/packages/capabilities/src/types.ts
+++ b/packages/capabilities/src/types.ts
@@ -35,6 +35,7 @@ import * as StorefrontCaps from './filecoin/storefront.js'
 import * as AggregatorCaps from './filecoin/aggregator.js'
 import * as DealTrackerCaps from './filecoin/deal-tracker.js'
 import * as DealerCaps from './filecoin/dealer.js'
+import * as IndexCaps from './index/index.js'
 import * as AdminCaps from './admin.js'
 import * as UCANCaps from './ucan.js'
 import * as PlanCaps from './plan.js'
@@ -52,6 +53,8 @@ export interface UCANAwait<Selector extends string = string, Task = unknown> {
  * An IPLD Link that has the CAR codec code.
  */
 export type CARLink = Link<unknown, typeof CAR.codec.code>
+
+export type Multihash = Uint8Array
 
 export type AccountDID = DID<'mailto'>
 export type SpaceDID = DID<'key'>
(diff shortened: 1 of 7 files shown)


packages/capabilities/src/types.ts
```diff
@@ -458,9 +492,8 @@ export type ServiceBlob = InferInvokedCapability<typeof W3sBlobCaps.blob>
 export type BlobAllocate = InferInvokedCapability<typeof W3sBlobCaps.allocate>
 export type BlobAccept = InferInvokedCapability<typeof W3sBlobCaps.accept>
 
-export type BlobMultihash = Uint8Array
```
vasco-santos, line 461: This is a breaking change, I think this is imported somewhere else IIRC. Either should be kept, or need to be flagged

packages/capabilities/src/types.ts
```diff
@@ -449,6 +452,37 @@ export type UploadGetFailure = UploadNotFound | Ucanto.Failure
...
+  /** Multihash digest of the shard that could not be found. */
+  digest: Multihash
+}
+
+/** A slice referenced by the index was not found in the specified shard. */
+export interface SliceNotFound extends Failure {
+  name: 'SliceNotFound'
+  /** Multihash digest of the slice that could not be found. */
+  digest: Multihash
+}
```
vasco-santos, line 484: just a question, are we going to validate this?

packages/capabilities/src/types.ts
vasco-santos: perhaps is just in `upload-api`, but honestly I would keep this or signal a Breaking Change

packages/capabilities/test/capabilities/index.test.js
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    assert.deepEqual(result.ok.audience.did(), w3.did())
+    assert.equal(result.ok.capability.can, 'index/add')
+    assert.deepEqual(result.ok.capability.nb, {
+      index: await createCarCid('test'),
+    })
+  })
+
+  it('index/add sholud fail when escalating index constraint', async () => {
```
vasco-santos, line 124:
```suggestion
  it('index/add should fail when escalating index constraint', async () => {
```

vasco-santos (commented):
LGTM, provided minor feedback below. 

I think the `BlobMultihash` thing needs to be addressed

packages/capabilities/src/types.ts
alanshaw: We have talked previously about sampling a few in https://github.com/w3s-project/specs/pull/85

vasco-santos (approved):
👌🏼

gammazero (approved):
I cannot see any required changes, but I am not sure I understand the ramifications/reason for changing from `BlobMultihash` to `Multihash`

PR #1411, opened 2024-04-25
feat: add "plan/create-admin-session" capability

travis:
Our billing system has an external admin customer portal system and their API lets us generate sessions for it on demand - add a capability that will let our customers delegate this ability the email address that is responsible for paying the bills.

See https://github.com/w3s-project/console/pull/98 for an example of this in action.


packages/w3up-client/README.md
```diff
@@ -9,6 +9,7 @@
 
 ## About
 
+
```
vasco-santos, line 12: could you add to the client the new methods folks can use for plans?

packages/w3up-client/test/helpers/mocks.js
vasco-santos, line 1: do we need it?

vasco-santos (approved):
As I was out of the loop, am assuming we decided this capability even though it may not be supported with future billing providers.

LGTM!  could we please add some docs though?

PR #1415, opened 2024-04-26
feat!: restrict store API to CARs

Gozala:
Restricts store API to CARs per our decision.


vasco-santos (approved):
LGTM! should this be treated as braking change?

Gozala:
> LGTM! should this be treated as braking change?

That thought did cross my mind & I thought I did not wanted to make it breaking so clients would get updates automatically, but now I'm realizing that we don't need clients to get updated to prevent non CAR uploads so I think breaking change is a good idea as this can cause lot of TS complaints

PR #1421, opened 2024-04-29
feat!: add `index/add` handler

alanshaw:
Adds a Ucanto handler for `index/add` invocations.

* Fetches index archive from the network
* Ensures index and all DAG shards are stored in the agent's space
* Publishes to IPNI

refs https://github.com/w3s-project/w3up/issues/1401


packages/upload-api/src/index/add.js
```diff
@@ -0,0 +1,75 @@
...
+  )
+  if (!idxAllocRes.ok) return idxAllocRes
+
+  // fetch the index from the network
+  const idxBlobRes = await context.blobRetriever.stream(idxLink.multihash)
+  if (!idxBlobRes.ok) {
+    if (idxBlobRes.error.name === 'BlobNotFound') {
+      return error(
+        /** @type {API.IndexNotFound} */
+        ({ name: 'IndexNotFound', digest: idxLink.multihash.bytes })
```
gammazero, line 38: Shouldn't we keep the `BlobNotFound` error to differentiate between when an index does not exist in the set of blob indexes (IndexNotFound) from when the index exists but what it points to cannot be found (`BlobNotFound`)?

alanshaw: Yes, exactly, the index is a blob also, this is just differentiating between generic blob not being found and something that says the actual index was not found.

packages/upload-api/src/index/lib/sharded-dag-index.js
```diff
@@ -0,0 +1,109 @@
...
+    /** multihash bytes, offset, length. */
+    Schema.tuple([Schema.bytes(), Schema.number(), Schema.number()])
+  ),
+])
+
+/**
+ * @param {ReadableStream<Uint8Array>} archive
+ * @returns {Promise<API.Result<API.ShardedDAGIndex, API.DecodeFailure|API.UnknownFormat>>}
+ */
+export const extract = async (archive) => {
```
Gozala, line 30: Other than CAR decoding, lot of the stuff here is provided by the schema facility e.g. `ShardedDAGIndexSchema.match` does quite a bit of work see here for some details https://github.com/web3-storage/ucanto/blob/main/packages/core/src/message.js

P.S. `view` function from link expects result from [`CAR.decode`](https://github.com/web3-storage/ucanto/blob/main/packages/core/src/car.js#L101) with `blocks` renamed to `store` (I know this was silly).

Gozala (approved):
Provided some feedback inline, pretty happy with how it looks

packages/upload-api/src/index/add.js
```diff
@@ -0,0 +1,75 @@
...
+      )
+    }
+    return idxBlobRes
+  }
+
+  const idxRes = await ShardedDAGIndex.extract(idxBlobRes.ok)
+  if (!idxRes.ok) return idxAllocRes
+
+  // ensure indexed shards are allocated in the agent's space
```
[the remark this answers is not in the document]
alanshaw: I think we should, as we're thinking of eventually being able to remove from IPNI when an agent removes from their space. I think this is harder to do later if we don't enforce now. We can always relax restriction later if we want to.

I will parallelize though 👍

packages/upload-api/src/index/add.js
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * @param {{ allocationsStorage: import('../types.js').AllocationsStorage }} context
+ * @param {API.SpaceDID} space
+ * @param {import('multiformats').MultihashDigest} digest
+ * @param {'IndexNotFound'|'ShardNotFound'|'SliceNotFound'} errorName
+ * @returns {Promise<API.Result<API.Unit, API.IndexNotFound|API.ShardNotFound|API.SliceNotFound|API.Failure>>}
+ */
+const assertAllocated = async (context, space, digest, errorName) => {
+  const result = await context.allocationsStorage.exists(space, digest.bytes)
+  if (result.ok == null) return result
```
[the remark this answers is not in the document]
alanshaw: `exists(...)` returns `ok: boolean` so we can't just check for falsey as it could literally be `false`.

I'll swap to check for error which is what I'm looking for here.

```suggestion
  if (result.error) return result
```

packages/upload-api/src/index/lib/digest-map.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+  return str
+}
+
+/**
+ * @template {import('multiformats').MultihashDigest} Key
+ * @template Value
+ * @implements {Map<Key, Value>}
+ */
+export class DigestMap extends Map {
```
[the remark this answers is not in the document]
alanshaw: No I don't think so, good suggestion 👍

packages/upload-api/src/index/lib/sharded-dag-index.js
alanshaw: ACK, I have split into `view` and `extract`.

PR #1422, opened 2024-04-30
fix: filecoin test use blob

vasco-santos:
Filecoin api exported tests still using CARs instead of Blobs, which means https://github.com/w3s-project/w3infra/pull/360 need to have TestContentStore to use CARs to make these tests work https://github.com/w3s-project/w3infra/pull/360/files#diff-cd843762832d0767495ff56f3f8692a15262dce245db89690794a1316868e25aR99

But, we want to use Blobs already, so let's make the tests also want Blobs across Filecoin pipeline :)

note that in practise both Blobs and CARs work, but for testing we are only writing one of them and is a CAR today https://github.com/w3s-project/w3up/blob/main/packages/filecoin-api/test/events/storefront.js#L42


packages/filecoin-api/test/utils.js
```diff
@@ -21,57 +19,34 @@ export async function randomBytes(size) {
...
- * @param {number} size
- */
-export async function randomCARs(length, size) {
-  return (
-    await Promise.all(Array.from({ length }).map(() => randomCAR(size)))
-  ).map((car) => ({
-    link: car.cid,
-    size: car.size,
-  }))
+  const blob = new Blob([bytes])
```
alanshaw, line 27: Does it need to be a `Blob` though?

alanshaw (approved):

PR #1425, opened 2024-04-30
feat: add blob protocol to upload-client

joaosa:
Clsoes https://github.com/w3s-project/project-tracking/issues/47

- Replaces the `store/add` call with `blob/add` on the upload-client;
- Add `blob/remove` and `blob/list`;
- Ports store-related tests to create a blob protocol counterpart; 
- Fixes associated tests.


packages/upload-client/test/helpers/utils.js
```diff
@@ -1 +1,86 @@
...
 export const validateAuthorization = () => ({ ok: {} })
+
+export const setupBlobAddResponse = async function({ issuer, with: space, proofs, audience }, invocation, blob) {
+  const blobAllocateTask = await W3sBlob.allocate
+    .invoke({
+      issuer,
+      audience,
+      with: space.did(),
+      nb: {
+        blob,
```
Gozala, line 17: Something like this `invocation.capabilities[0].nb.blob`

Gozala:
@joaosa I suspect that w3up-client tests are failing because the upload-api in-memory service isn't wired with new capabilities, I'm not exactly sure why though

joaosa:
Something is eluding me as suddenly the CI decided to start firing prettier errors with no changes from my side 🤔

packages/w3up-client/test/capability/blob.test.js
```diff
@@ -0,0 +1,111 @@
...
+      // @ts-expect-error
+      provider: connection.id.did(),
+      account: alice.agent.did(),
+      consumer: space.did(),
+    })
+
+    const car = await randomCAR(128)
+    const multihash = await alice.capability.blob.add(car)
+
+    assert.deepEqual(await blobsStorage.has(car.cid.multihash.bytes), {
```
joaosa, line 35: this is failing

packages/upload-client/src/blob.js
```diff
@@ -0,0 +1,398 @@
...
+ * invocation(s). It is typically the user _agent_.
+ *
+ * The `with` is the resource the invocation applies to. It is typically the
+ * DID of a space.
+ *
+ * The `proofs` are a set of capability delegations that prove the issuer
+ * has the capability to perform the action.
+ *
+ * The issuer needs the `blob/add` delegated capability.
+ * @param {Blob|Uint8Array} data Blob file data.
```
alanshaw, line 164:
```suggestion
 * @param {Blob|Uint8Array} data Blob data.
```

packages/upload-client/src/blob.js
```diff
@@ -0,0 +1,398 @@
...
+    throw new Error(`failed ${BlobCapabilities.add.can} invocation`, {
+      cause: result.out.error,
+    })
+  }
+
+  return multihash
+}
+
+/**
+ * List Blob files stored by the issuer.
```
alanshaw, line 305:
```suggestion
 * List Blobs stored in the space.
```

packages/upload-client/src/blob.js
```diff
@@ -0,0 +1,398 @@
...
+    throw new Error(`failed ${BlobCapabilities.list.can} invocation`, {
+      cause: result.out.error,
+    })
+  }
+
+  return result.out.ok
+}
+
+/**
+ * Remove a stored Blob file by digest.
```
alanshaw, line 353:
```suggestion
 * Remove a stored Blob by it's digest.
```

packages/upload-client/src/blob.js
```diff
@@ -0,0 +1,398 @@
...
+ * The `issuer` is the signing authority that is issuing the UCAN
+ * invocation(s). It is typically the user _agent_.
+ *
+ * The `with` is the resource the invocation applies to. It is typically the
+ * DID of a space.
+ *
+ * The `proofs` are a set of capability delegations that prove the issuer
+ * has the capability to perform the action.
+ *
+ * The issuer needs the `store/remove` delegated capability.
```
alanshaw, line 367:
```suggestion
 * The issuer needs the `blob/remove` delegated capability.
```

packages/upload-client/src/types.ts
```diff
@@ -95,6 +125,18 @@ export interface ProgressStatus extends XHRProgressStatus {
...
+  ucan: {
+    conclude: ServiceMethod<
+      UCANConclude,
+      UCANConcludeSuccess,
+      UCANConcludeFailure
+    >
+  }
+  blob: {
+    add: ServiceMethod<BlobAdd, BlobAddSuccess, BlobAddFailure>
+    remove: ServiceMethod<BlobRemove, BlobRemoveSuccess, BlobRemoveFailure>
```
alanshaw, line 137: Argh, I completely forgot, we're also going to need `blob/get` for the console. Could you send a follow up PR to add? So sorry.

packages/upload-client/test/blob.test.js
```diff
@@ -0,0 +1,702 @@
...
+  setupBlobAdd4xxResponse,
+  setupBlobAdd5xxResponse,
+} from './helpers/utils.js'
+import { fetchWithUploadProgress } from '../src/fetch-with-upload-progress.js'
+
+describe('Blob.add', () => {
+  it('stores a DAG with the service', async () => {
+    const space = await Signer.generate()
+    const agent = await Signer.generate()
+    const car = await randomCAR(128)
```
alanshaw, line 25: I'd use just random bytes here - there's no need to associate this with CARs. Same for all of these tests.

alanshaw (changes requested):
Looks great, I've left some minor feedback but this should be good to merge after it's addressed IMO.

🚀

packages/upload-client/test/blob.test.js
joaosa: addressed

packages/w3up-client/test/capability/blob.test.js
```diff
@@ -0,0 +1,115 @@
+import { AgentData } from '@web3-storage/access/agent'
+import { randomCAR } from '../helpers/random.js'
+import { Client } from '../../src/client.js'
+import * as Test from '../test.js'
+
+export const BlobClient = Test.withContext({
+  'should store a CAR file': async (
```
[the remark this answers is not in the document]
joaosa: addressed

packages/upload-api/src/blob/add.js
```diff
@@ -161,6 +161,11 @@ async function allocate({ context, blob, space, cause }) {
 
   // 3. if not already allocated (or expired) execute `blob/allocate`
   if (!blobAllocateReceipt) {
+    // Create allocation task and save it
+    const saveTask = await context.tasksStorage.put(task)
```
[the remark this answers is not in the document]
joaosa: This is actually part of @Gozala's PR into my branch, but good to get more context 👍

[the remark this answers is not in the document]
Gozala: Yeah I have added this before I realized what was going on, I have branch that I'm actively trying to finish that should fix this and will remove these lines also.

packages/upload-client/src/blob.js
```diff
@@ -0,0 +1,398 @@
...
+  )
+  const ucanConclude = await httpPutConcludeInvocation.execute(conn)
+
+  if (!ucanConclude.out.ok) {
+    throw new Error(`failed ${BlobCapabilities.add.can} invocation`, {
+      cause: result.out.error,
+    })
+  }
+
+  return multihash
```
Gozala, line 301: I would expect blob/add not to return until task corresponding to `result.out.ok.site` is done, which I realize from the current state of the backend implementation have no good way of doing. Never the less I'll try to make a PR that will add missing pieces. I don't think we should block this in the meantime.

packages/upload-client/src/types.ts
joaosa: I don't really see this [capability in the spec](https://github.com/w3s-project/specs/blob/main/w3-blob.md). How should we go about this?

alanshaw: Ah yes it'll need a space PR also.

Essentially the same as `store/get` - it allows the caller to get the size of the blob (and potentially other stuff in the future). The allocations storage already has a method, you just need to create a handler, call it and return it: https://github.com/w3s-project/w3up/blob/9982d12b0a1f9a6da3f0d4264b9a35348e189dfb/packages/upload-api/src/types/blob.ts#L20-L23

PR #1426, opened 2024-05-01
chore: use nodejs 18 in `upload-api` tests

alanshaw:
We switched w3infra to sst v2 and with it upgraded to nodejs 18. We should test on same as infra.


alanshaw:
Merged with https://github.com/w3s-project/w3up/pull/1421

PR #1434, opened 2024-05-03
refactor: Move blob index logic from upload-api to blob-index lib

gammazero:
The blob index logic should be an independent module as it will be used in both client and service code.


packages/blob-index/src/sharded-dag-index.js
```diff
@@ -0,0 +1,251 @@
...
+  await archive.pipeThrough(reader).pipeTo(
+    new WritableStream({
+      write: (block) => {
+        blocks.set(block.cid.multihash, block.bytes)
+      },
+    })
+  )
+
+  const header = await reader.getHeader()
+  if (header.roots[0].code !== dagCBOR.code) {
```
alanshaw, line 50:
```suggestion
  if (header.roots[0]?.code !== dagCBOR.code) {
```

packages/blob-index/src/sharded-dag-index.js
```diff
@@ -0,0 +1,251 @@
...
+    return this.#reason ?? 'failed to decode'
+  }
+}
+
+/**
+ *
+ * @param {API.UnknownLink} content
+ * @returns {API.ShardedDAGIndex}
+ */
+export function makeShardedDAGINdex(content) {
```
alanshaw, line 124: More usually "create":

```suggestion
export function createShardedDAGIndex(content) {
```

packages/blob-index/src/sharded-dag-index.js
```diff
@@ -0,0 +1,251 @@
...
+      slices.set(mh, s[1])
+      //const [offset, length] = s[1]
+      //console.log('multihash:', mh, 'offset:', offset, 'lenght:', length)
+    }
+  }
+  return dagIndex
+}
+
+/**
+ * Create a sharded DAG index by indexing blocks in the the passed CAR shards.
```
alanshaw, line 219:
```suggestion
 * Create a sharded DAG index by indexing blocks in the passed CAR shards.
```

packages/blob-index/package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "@web3-storage/blob-index",
```
alanshaw, line 2: The blob index is a subset of the sharded DAG index should this be named something different?

Suggestion, call it `@web3-storage/index`? Then have exports for `@web3-storage/index/dag-sharded` and `@web3-storage/index/blob`?

alanshaw (changes requested):

packages/blob-index/package.json
gammazero: Well, it is for indexing blobs. I mean, a DAG is what is being stored, but the DAG 's chunks that are stored in one or more blobs is what is being indexed. Whereas if a DAG was stored in some other construct, say bit-bucket, then it would be a bit-bucket-index, even though a sharded-dag is still being indexed. That's what I was thinking.

If you see this same indexing scheme being used universally, whether the DAG is stored in a blob, bit-bucket, etc., then yes, let's call it an "index" as you suggest.

alanshaw: Ok makes sense lets go with it!

packages/blob-index/src/sharded-dag-index.js
```diff
@@ -0,0 +1,251 @@
...
+    const cid = Link.create(dagCBOR.code, digest)
+    return ok(CAR.encode({ roots: [{ cid, bytes }], blocks }))
+  }
+}
+
+/**
+ * @param {Uint8Array} carData
+ * @returns {Promise<ShardedDAGIndex>}
+ */
+export const fromArchive = async (carData) => {
```
alanshaw, line 167: This is what `extract` does above no?

gammazero: Extract expects to read from a `ReadableStream` whereas `fromArchive` reads from a `UInt8Array` in memory. Does this mean that `fromArchive` should just convert the `Uint8Array` into a `ReadableStream`?

Also, `extract` will not know how to do things like discarding the overall dag shard from the encoded shards, before building the ShardedDAGIndex.

alanshaw (approved):

PR #1438, opened 2024-05-08
fix: test against actual api

Gozala:
Removes mocks from tests everywhere in w3up client and tests against in-memory w3up service. This is aimed to help with #1425, which was suffering from issues caused by mocks


joaosa (approved):
LGTM 🙏

PR #1442, opened 2024-05-08
fix: blob add

Gozala:
Fixes various issues to get tests passing

### DIFF
--- a/packages/upload-api/src/blob/add.js
+++ b/packages/upload-api/src/blob/add.js
@@ -161,6 +161,11 @@ async function allocate({ context, blob, space, cause }) {
 
   // 3. if not already allocated (or expired) execute `blob/allocate`
   if (!blobAllocateReceipt) {
+    // Create allocation task and save it
+    const saveTask = await context.tasksStorage.put(task)
+    if (!saveTask.ok) {
+      return saveTask
+    }
     // Execute allocate invocation
     const allocateRes = await allocate.execute(context.getServiceConnection())
     if (allocateRes.out.error) {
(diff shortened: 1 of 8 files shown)


packages/upload-api/src/blob/add.js
```diff
@@ -161,6 +161,11 @@ async function allocate({ context, blob, space, cause }) {
 
   // 3. if not already allocated (or expired) execute `blob/allocate`
   if (!blobAllocateReceipt) {
+    // Create allocation task and save it
+    const saveTask = await context.tasksStorage.put(task)
+    if (!saveTask.ok) {
+      return saveTask
+    }
```
Gozala, line 168: `ucan/conclude` seems to assume that `blob/allocate` task is stored here, but nothing was storing it which caused errors when `http/put` receipts were submitted. I'm not sure what @vasco-santos had intended here, as some tests seemed to manually store these. But then some handlers seemed to save task from the handler.

I think it is reasonable to save task from the handler especially if we intend to look those up later.

packages/upload-client/src/blob.js
```diff
@@ -173,7 +173,7 @@ export async function add(
   const bytes =
     car instanceof Uint8Array ? car : new Uint8Array(await car.arrayBuffer())
   const link = await CAR.codec.link(bytes)
-  const digest = link.bytes
+  const digest = link.multihash.bytes
```
Gozala, line 176: @joaosa this was the culprit that caused test failures, link is a CID which is CID prefix + multihash which is why errors where saying digest length was greater than expected.

packages/upload-client/src/blob.js
```diff
@@ -279,30 +279,10 @@ export async function add(
...
-      },
-      headers: {
-        'ucan/await': ['.out.ok.address.headers', nextTasks.allocate.task.cid],
-      },
-    },
-    facts: nextTasks.put.task.facts,
-    expiration: Infinity,
-  })
 
-  const httpPutDelegation = await httpPut.delegate()
```
Gozala, line 302: I removed this because it effectively recreates same task that we already have reference to.

joaosa (approved):
LGTM

PR #1444, opened 2024-05-13
feat: integrate agent store for idempotence & invocation/receipt persistence 

Gozala:
This is a draft that I could not finished on Friday, but I thought I'd share in case @vasco-santos or @alanshaw are looking into related stuff on Monday.

PR introduces AgentStore that is used for writing ucanto AgentMessage's from which receipts and invocations get indexed. This store is then consulted by a service during invocations to provide idempotence or more simply to return receipt without executing a task if we already have one.

Main goal here is to take out custom logic out of w3infra and to remove need for manual receipt / invocation storing in the handlers.  

⚠️ It is worth calling out that this is significantly changes how things work. In the past sending same invocation used to re-run the handler, with this change that is no longer the case and you'll get a same receipt back (which is a neat way to lookup receipts actually). However that also means that bunch of our tests were affected. I don't expect it would affect real world cases however, because same invocation send second later will have unique CID (due to how we generate expiry) however it does affect tests because we don't wait a second between calls. That is why nonce's were added in a lot of places, but that is not something I expect users would have to do.


packages/upload-api/src/ucan/conclude.js
```diff
@@ -1,106 +1,33 @@
...
+    // way even if invocation occurs later we will have receipt already.
+    const message = await Message.build({ receipts: [receipt] })
+    const save = await context.agentStore.messages.write(message)
+    if (save.error) {
+      return save
     }
 
-    return {
-      ok: { time: Date.now() },
+    // If palling errors we propagate errors that is because referenced
```
vasco-santos, line 23: we should keep the comment that this is an HACK for now

packages/upload-api/src/ucan/conclude.js
```diff
@@ -1,106 +1,33 @@
...
-
-    const scheduleErrors = scheduleRes.filter((res) => res.error)
-    if (scheduleErrors.length && scheduleErrors[0].error) {
-      return {
-        error: scheduleErrors[0].error,
-      }
+    // First of all that we create a message and save it to agent store, that
+    // way even if invocation occurs later we will have receipt already.
+    const message = await Message.build({ receipts: [receipt] })
+    const save = await context.agentStore.messages.write(message)
```
vasco-santos, line 18: is the goal here to save the message and all the indexes to it like w3infra does today?

packages/upload-api/src/ucan/conclude.js
```diff
@@ -1,106 +1,33 @@
...
+    if (save.error) {
```
vasco-santos, line 19: also, I don't think we should save a message with a receipt for an unknown invocation (which is prevented in previous code)

packages/upload-api/src/blob/accept.js
```diff
@@ -60,3 +62,76 @@ export function blobAcceptProvider(context) {
...
+    //
+    // ℹ️ This works around the fact that we index receipts by invocation link
+    // as opposed to task link which would not care about the expiration.
+    expiration: result.ok.expiration,
+  })
+
+  // We do not care about the result we just want receipt to be issued and
+  // stored.
+  await blobAccept.execute(context.getServiceConnection())
+
```
vasco-santos, line 135: we do not want the task scheduler abstraction anymore?

packages/upload-api/src/blob/allocate.js
```diff
@@ -3,93 +3,120 @@ import * as W3sBlob from '@web3-storage/capabilities/web3.storage/blob'
...
+   * @param {API.Capability} source.capability
+   */
+  constructor({ capability: { with: subject, can } }) {
+    super()
+
+    this.capability = { with: subject, can }
+  }
+  describe() {
+    return `${this.capability.with} does not have a "${this.capability.can}" capability provider`
+  }
```
vasco-santos, line 121: This should actually also be in accept right? Can we move this to another file where it would feel more reasonable to be shared across several server providers?

packages/upload-api/src/types.ts
```diff
@@ -183,15 +186,10 @@ import { SubscriptionsStorage } from './types/subscriptions.js'
...
 import { UsageStorage } from './types/usage.js'
 export type { UsageStorage }
-import { ReceiptsStorage, TasksScheduler } from './types/service.js'
-export type { ReceiptsStorage, TasksScheduler }
-import {
-  AllocationsStorage,
-  BlobsStorage,
-  TasksStorage,
-  BlobAddInput,
-} from './types/blob.js'
```
vasco-santos, line 193: should we clean up the Tasks Storage from here?

packages/upload-api/test/access-client-agent.js
```diff
@@ -162,7 +162,10 @@ export const test = {
     const accountProofs = [delegationFromAccountToSession, attestation]
     assert.ok(accountProofs)
   },
-  'same agent, multiple accounts, provider/add': async (assert, context) => {
+  'skip same agent, multiple accounts, provider/add': async (
```
vasco-santos, line 165: should not be skipped

packages/upload-api/test/handlers/blob.js
```diff
@@ -153,7 +149,7 @@ export const test = {
       firstNext.allocate.task.link().equals(secondNext.allocate.task.link())
     )
   },
-  'blob/add schedules allocation and returns effects for allocate, accept and put together with their receipts (when stored)':
+  'only blob/add schedules allocation and returns effects for allocate, accept and put together with their receipts (when stored)':
```
vasco-santos, line 152:
```suggestion
  'blob/add schedules allocation and returns effects for allocate, accept and put together with their receipts (when stored)':
```
```suggestion
  'only blob/add schedules allocation and returns effects for allocate, accept and put together with their receipts (when stored)':
```

packages/upload-api/test/helpers/context.js
```diff
@@ -47,7 +47,7 @@ export const createContext = async (
     storefront: { filecoinSubmitQueue, pieceOfferQueue },
   } = getFilecoinQueueImplementations(queuedMessages)
   const {
-    storefront: { pieceStore, receiptStore, taskStore },
+    storefront: { pieceStore },
```
vasco-santos, line 50: should we not do in same PR `filecoin-api` and `upload-api` refactor for this? This would make us need to update both infras as well, and perhaps better one at a time?

vasco-santos: See https://github.com/w3s-project/w3up/blob/main/packages/filecoin-api/src/storefront/api.ts#L61

vasco-santos (changes requested):
I am happy that we abstract this logic out of the infra, which also enables us to take the assumptions from the infra. However, it would be great to understand more on the context of this change now, as it has significant changes on the infra behind and would love to only do that when we have a first thing working end to end. Is the intention to do this at same time?

Did a first pass here, and left some thoughts inline.

Also, be aware on the impact I mentioned inline to the filecoin-api as well.

packages/upload-api/src/blob/accept.js
```diff
@@ -60,3 +62,76 @@ export function blobAcceptProvider(context) {
...
+    ? { ok: receipt.ran }
+    : await context.agentStore.invocations.get(receipt.ran)
+
+  // If can not find an invocation for this receipt there is nothing to do here,
+  // if it was receipt for `http/put` we would have invocation record.
+  if (!ran.ok) {
+    return { ok: {} }
+  }
+
+  // Detect if this is a receipt is for an `http/put` invocation
```
alanshaw, line 86:
```suggestion
  // Detect if this receipt is for an `http/put` invocation
```

packages/upload-api/src/blob/accept.js
```diff
@@ -60,3 +62,76 @@ export function blobAcceptProvider(context) {
...
+  // If it's not an http/put invocation nothing to do here.
+  if (put == null) {
+    return { ok: {} }
+  }
+
+  // Otherwise we are going to lookup allocation corresponding to this http/put
+  // in order to issue blob/accept.
+  const [, allocation] = /** @type {API.UCANAwait} */ (put.nb.url)['ucan/await']
+  const result = await context.agentStore.invocations.get(allocation)
+  // If count not lookup blob/allocate invocation there is something wrong in
```
alanshaw, line 100:
```suggestion
  // If can not lookup blob/allocate invocation there is something wrong in
```

packages/upload-api/src/blob/accept.js
```diff
@@ -60,3 +62,76 @@ export function blobAcceptProvider(context) {
...
+  // the system and we return error so it could be propagated to the user. It is
+  // not a proper solution as user can not really do anything, but still seems
+  // better than silently ignoring, this way user has a chance to report a
+  // problem. Clint test could potentially also catch errors.
```
alanshaw, line 104:
```suggestion
  // problem. Client test could potentially also catch errors.
```

packages/upload-api/src/blob/accept.js
```diff
@@ -60,3 +62,76 @@ export function blobAcceptProvider(context) {
     },
   })
 }
+
+/**
+ * Polls `blob/accept` whenever we receive a receipt. It may error if passed
```
alanshaw, line 67: Do you mean "this is called whenever we receive a receipt"? It doesn't do the polling, it is the thing that is polled.

packages/upload-api/src/executor.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @template {Record<string, any>} T
+ * @param {API.AgentMessage} input
+ * @param {API.Server<T>} server
+ * @param {Context} context
+ * @returns {Promise<API.Result<API.AgentMessage<T>, Error>>}
+ */
+export const execute = async (input, server, context) => {
+  const saveInput = await context.messageStore.put(input)
```
alanshaw, line 24: What is the difference between this and `agentStore.messages`?

alanshaw (commented):
I like where this is going - thanks for dealing with it! I left a few comments but I realized it's still WIP so paused!

Gozala:
> I am happy that we abstract this logic out of the infra, which also enables us to take the assumptions from the infra. However, it would be great to understand more on the context of this change now, as it has significant changes on the infra behind and would love to only do that when we have a first thing working end to end. Is the intention to do this at same time?

Yeah I was going to create PR for infra after I have this is complete. Then land two in concert. I also hope that old infra would be able to continue to work with these changes without utilizing the new stuff and therefor could choose to opt-in when it so chooses.

packages/upload-api/src/blob/accept.js
Gozala: I do think we want a scheduler yes, but I don't think what we had could be called scheduler and at that point it just introduced extra indirection which is why I removed it. We can keep it if you think otherwise, but unless you do I suggest we introduce scheduler when it actually does scheduling

packages/upload-api/src/executor.js
Gozala: `agentStore.messages` is not part of the `AgentMessage` interface, it's merely an implementation detail of the in-memory `AgentStore` (which in fact returns `this`). Different implementation e.g. w3infra may not have `.messages` field.

Gozala: Oh wait never mind this file was not meant to be committed it was a sketch that got integrated into `./lib.js` instead.

packages/upload-api/src/ucan/conclude.js
Gozala: Conceptually I'm thinking of `AgentStore` as ucanto agent cache. Implementation of such cache can choose cache invalidation strategy best suited for the use case. E.g. w3infra could invalidate receipts for invocations it does not care about (has seen) or it could store them all in issuers space, if there is one or discard otherwise. It could also just store everything and worry about GC afterwards perhaps utilizing something like LRU evection strategy.

Bigger picture aside, yes I was intending to just store receipts in the receipt store we have setup. I realize it could end up storing things we don't care about, but that is also the case with delegations store and bunch of other things. I don't worry about it much because there is very little to be gained by sending us pointless receipts, but perhaps not storing receipts for invocations we have not seen is a better default, which I suspect is you're concern here.

Simply storing and not worrying we get following

1. Simpler code that is easier to maintain.
2. Support for out of order delivery (that is we may see receipt first and invocation after)

What do you think ?

packages/upload-api/src/ucan/conclude.js
Gozala, line 23:
```suggestion
    // If polling errors we propagate errors that is because referenced
```

packages/upload-api/test/handlers/blob.js
```diff
@@ -281,11 +275,20 @@ export const test = {
       }
 
       // Invoke `blob/add` for the third time (after invoking conclude)
-      const thirdBlobAdd = await invocation.execute(connection)
+      const thirdBlobAdd = await BlobCapabilities.add
+        .invoke({ ...task, nonce: 'third' })
```
Gozala, line 279: @vasco-santos I'm running into this issue that I don't know yeh how to address so writing here as it is related to something we talked today on the call.

- Every `blob/allocate` will have a unique task because we have [`nb.cause` field](https://github.com/w3s-project/specs/blob/main/w3-blob.md#allocate-blob). I can't recall the reason why it was introduced but I think you made a case for it.
- Consequently every `http/put` also gets unique task because it references the `blob/allocate` which is also unique.
- Each `blob/accept` is also unique and depends on unique `http/put` receipt.

All of this combined with blob/add handler works out such that we have each upload same blob even if we already have it stored, which is not what we want.

This test used to pass before changes in the PR because same exact invocation was passed three times, which after this PR would just hit the cache and return same exact receipt. With this PR I had to update to use nonce to have those invocations have different CIDs, which in turn caused http/put invocation to be different in each call and therefor we are failing to find a receipt in the last one which what test depends upon.

Gozala: I think there are two possible solutions here:

1. We should check if we have a blob already in the system and if so simply issue blob/accept instead of looking up an `http/put` receipt.
2. Do a proper receipt resolution that we talked about on the call that is invocation with some awaits in them should resolve to same receipt as invocation with awaits substituted with values.
    - 😓 Doing this is non trivial because we don't identify tasks by task CID but rather by an invocation CID which in retrospect was a mistake, but also I don't think there is a simple way to change right now.

packages/upload-api/test/helpers/blob.js
```diff
@@ -54,6 +60,18 @@ export function parseBlobAddReceiptNext(receipt) {
...
+  if (acceptReceipt?.out.ok?.site) {
+    const blocks = new Map(
+      [...receipt.iterateIPLDBlocks()].map((block) => [`${block.cid}`, block])
+    )
+
+    acceptSite.site = Delegation.view({
+      root: /** @type {API.UCANLink} */ (acceptReceipt.out.ok.site),
+      blocks,
+    })
+  }
```
Gozala, line 73: @joaosa this is exact code (which we discussed) that resolves site commitment from the receipt

packages/upload-api/test/handlers/blob.js
vasco-santos: I was making a case to put it in a receipt, but we found and alternative. The `cause` is already there since `store/*` and I think it became a pattern, as well as would allow us to have indexes over receipts as we previously talked. Also, from billing stand point, this is needed so that we know which space to bill for space. 

`blob/allocate` should only run a second time if presigned url expired though, otherwise we should be able to find a receipt for it. But maybe we should check the allocations table instead? That way we would have the CID for the `allocate` receipt?

packages/upload-api/src/blob/accept.js
vasco-santos: we could easily introduce a queue on the infra side to allow retries out of the box, so that was the main advantage I was seeing with current implementation. But no strong feelings

packages/upload-api/src/ucan/conclude.js
vasco-santos: What I was meaning was more, if this first iteration of the AgentStore would do exactly same as we do in w3infra and store things into `workflow`, `task` and `invocation` buckets.

I believe in the long run, we should do what we talked about and store this at the appropriate space, but now would likely be good to not think about a new way of storing and reading things, as a migration and so on would also need to be done

vasco-santos: in other words, if the store initial impl would so something like what happens in https://github.com/w3s-project/w3infra/blob/main/upload-api/functions/ucan-invocation-router.js#L318

packages/upload-api/test/handlers/blob.js
Gozala: > would allow us to have indexes over receipts as we previously talked.

My memory is failing me here, not sure what indexing this would relate to now.

> Also, from billing stand point, this is needed so that we know which space to bill for space.

I'm not sure that's accurate because we have `nb.space` which tells us which space to bill without having to lookup through `cause`

> `blob/allocate` should only run a second time if presigned url expired though, otherwise we should be able to find a receipt for it. 

Well that is kind of the problem, we won't find a receipt if user invoked `blob/add` a second time because invocation cid will be different and consequently every other invocation due to that `cause` field. If we remove the `cause` field than we may find receipt for the same allocation.


> But maybe we should check the allocations table instead? That way we would have the CID for the `allocate` receipt?

I have updated implementation and tests which I think behave as expected now, so probably best to discuss details inline there.

packages/upload-api/src/ucan/conclude.js
Gozala: > in other words, if the store initial impl would so something like what happens in https://github.com/w3s-project/w3infra/blob/main/upload-api/functions/ucan-invocation-router.js#L318

It is exactly what I defined the interface from with intention to do exactly that

packages/upload-api/test/handlers/blob.js
vasco-santos: > My memory is failing me here, not sure what indexing this would relate to now.

https://github.com/w3s-project/RFC/pull/23/files#diff-2969327d85eb81a450f0d878aa3d25d484fc3404e6497f98a1db3a5cdb2db5ddR5

This is what I mean! 

> I'm not sure that's accurate because we have nb.space which tells us which space to bill without having to lookup through cause

oh yeah, you are right!

packages/upload-api/src/ucan/conclude.js
Gozala: @vasco-santos I attempted to provide some rational behind the change in the comments above. Please let me if you're ok with storing receipts without having to find invocations for those or if you do still think they should be rejected.

Gozala: I removed this lines entirely since I have since realized that all of this was kind of a moot point. If we received message with `ucan/conclude` invocation we have already stored receipt contained by it either way. So we really don't need to create another message and save it again regardless whether we are aware of invocation or not.

What I end up doing is made `AgentStore` aware of `ucan/conclude` so it would index contained receipt.

In the end things end up looking even simpler because ucan conclude handler doesn't really need to do anything other than poll the accept handler if receipt is for the `http/put`.

packages/upload-api/src/blob/accept.js
```diff
@@ -60,3 +72,76 @@ export function blobAcceptProvider(context) {
...
```
Gozala, line 77:
```suggestion
 * Polls `blob/accept` task whenever we receive a receipt. It may error if passed
```

Does this make more sense ? For more context I have been modeling invocations after Rust's Async Tasks a.k.a Futures. Scheduler [`poll`](https://doc.rust-lang.org/std/future/trait.Future.html#the-poll-method)s pending tasks (which is modeled after [epoll](https://en.wikipedia.org/wiki/Epoll)) to attempt make progress.

What we do here is very similar, on a system event in form of receipt we poll tasks and progress ones that were awaiting on it.

packages/upload-api/src/blob/accept.js
Gozala:
```suggestion
 * Polls `blob/accept` task whenever we receive a receipt. It may error if passed
```

Does this make more sense ? For more context I have been modeling invocations after Rust's Async Tasks a.k.a Futures. Scheduler [`poll`](https://doc.rust-lang.org/std/future/trait.Future.html#the-poll-method)s pending tasks (which is modeled after [epoll](https://en.wikipedia.org/wiki/Epoll)) to attempt make progress.

What we do here is very similar, on a system event in form of receipt we poll tasks and progress ones that were awaiting on it.

packages/upload-api/src/ucan/conclude.js
vasco-santos: > I removed this lines entirely since I have since realized that all of this was kind of a moot point. If we received message with ucan/conclude invocation we have already stored receipt contained by it either way. So we really don't need to create another message and save it again regardless whether we are aware of invocation or not.

That seems reasonable if we make sure AgentStore implementations really make the indexing

packages/upload-api/src/blob/add.js
```diff
@@ -16,101 +33,54 @@ export function blobAddProvider(context) {
...
 
-      // Verify blob is within accept size
-      if (blob.size > context.maxUploadSize) {
-        return {
-          error: new BlobSizeOutsideOfSupportedRange(
-            blob.size,
-            context.maxUploadSize
-          ),
-        }
-      }
```
vasco-santos, line 31: Why is this being removed?

packages/upload-api/src/blob/allocate.js
```diff
@@ -1,95 +1,129 @@
...
-        blob.size,
-        expiresIn
-      )
-      if (createUploadUrl.error) {
-        return createUploadUrl
+  // Verify blob is within the max upload size.
+  if (capability.nb.blob.size > context.maxUploadSize) {
+    // While blob may exceed current maxUploadSize limit it could be that limit
+    // was higher in the past and user had this blob uploaded already in which
+    // case we should not error.
```
vasco-santos, line 53: I am not sure why this moved here instead of `blob/add`. Would you expect that write targets could have different max sizes? Otherwise, I don't understand the change as it results in way more processing for same end result

packages/upload-api/src/ucan/conclude.js
```diff
@@ -1,106 +1,34 @@
...
-            .delegate()
-
-          return tasksScheduler.schedule(blobAccept)
-        })
+    // 🚧 THIS IS A TEMPORARY HACK 🚧
+    // When we receive a receipt for the invocation we want to resume the tasks
+    // that were awaiting in the background. In the future task scheduler is
+    // expected to handle coordination of tasks and their dependencies. In the
+    // meantime we poll `blob/allocate` tasks that were awaiting for the
+    // `http/put` receipt.
```
vasco-santos, line 18: nit: if we need to add further triggers like this in the future, we would need to decouple this into where it is. There is why I would check on it, and I still think it is better to do

packages/upload-api/test/storage/agent-store-tests.js
```diff
@@ -0,0 +1,135 @@
...
+import { sha256 } from 'multiformats/hashes/sha2'
+import * as BlobCapabilities from '@web3-storage/capabilities/blob'
+import { Console } from '@web3-storage/capabilities'
+
+import { alice, registerSpace } from '../util.js'
+
+/**
+ * @type {API.Tests}
+ */
+export const test = {
```
vasco-santos, line 12: This MUST test the special case for `ucan/conclude` or implementations could not adhere with the expectations we have for the blob flow to work

vasco-santos (commented):
Overall looks good to me!

Not approved essentially because agent store tests should be expanded as reviewed inline

Also worth mentioning, there is a big refactor here on blob/add that would be much easier to be a separate PR instead of intertwined with a lot of other changes. The code seems way better documented for folks without big prior knowledge, so I am happy with these changes still.  But made reviewing much more complicated and time consuming

packages/upload-api/src/blob/add.js
Gozala: It moved to an allocation handler

packages/upload-api/src/blob/accept.js
```diff
@@ -59,3 +73,76 @@ export function blobAcceptProvider(context) {
...
+  // If it's not an http/put invocation nothing to do here.
+  if (put == null) {
+    return { ok: {} }
+  }
+
+  // Otherwise we are going to lookup allocation corresponding to this http/put
+  // in order to issue blob/accept.
+  const [, allocation] = /** @type {API.UCANAwait} */ (put.nb.url)['ucan/await']
+  const result = await context.agentStore.invocations.get(allocation)
+  // If could not find blob/allocate invocation there is something wrong in
```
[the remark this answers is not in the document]
Gozala: We could probably set this up as capability handler and dispatch to it, that way we would be able to throw here which will end up triggering `catch` handler of the server and get logged instead.

Does that sound like a better option ?

packages/upload-api/src/blob/allocate.js
Gozala: I think you are making some good points on this. That said, I also think that I also think there should be no `blob/add` handler it should be a static workflow like one illustrated here https://github.com/w3s-project/RFC/pull/25#issuecomment-2070641427

Perhaps what we miss is a prior step that checks that blob size is below the max limit (and that step can even run locally on the invoking agent (like http/put) as opposed to service.

Overall sounds like spec needs to address this by calling out the verification step that needs to take place. In the meantime I suggest we keep this here and we can address this as part of router design which will actually deal with multiple allocation sites. Does this sound reasonable ?

If you feel very strongly about it I can move this back into the blob/add handler, however please note that allocation site would need to have this kind of check anyway as they may have different constraints to work with.

packages/upload-api/src/ucan/conclude.js
Gozala: I'm not sure I follow what you are saying here. Are you saying what if we need to resume some other blocked task ? If so I think answer there would to add something like `Scheduler.poll(....)` which would go and poll each task we need to poll that way logic around individual tasks lives where those task are defined as opposed to been entangled with conclude handler with a branched logic

packages/upload-api/test/handlers/blob.js
Gozala: How does removing `cause` here sounds (in a separate PR & spec) we can introduce it back when we really need it.

packages/upload-api/test/storage/agent-store-tests.js
Gozala: Not sure I understand can you please help me better understand what special case are you talking about ? Also could we add this as a followup PR so I don't have to keep rebasing it all the time. I don't think I have introduced any new edge cases that before, but maybe I'm missing something

vasco-santos (approved):
LGTM after mentioned test in

packages/upload-api/test/storage/agent-store-tests.js
Gozala: Added tests to ensure that receipts inside `ucan/conclude` get indexed.

PR #1445, opened 2024-05-13
fix: add format specifier to blob location claim URL

alanshaw:
Specifier ensures only the single raw block is served by the gateway. It is not a DAG, it's a single block of content.


vasco-santos (commented):
this looks good to me, but tests need to be updated I believe

vasco-santos (approved):
this looks good,

PR #1450, opened 2024-05-14
fix: cid for filecoin offer must be raw

vasco-santos:
As we previously decided, `filecoin/offer` should receive raw CIDs https://github.com/w3s-project/w3up/issues/1408#issuecomment-2082091981 so that everything works out of the box and Blobs can be read from roundabout for validation + from SPs


alanshaw (approved):

packages/upload-client/src/index.js
```diff
@@ -134,12 +134,15 @@ async function uploadBlockStream(
           const bytes = new Uint8Array(await car.arrayBuffer())
           // Invoke blob/add and write bytes to write target
           const multihash = await Blob.add(conf, bytes, options)
+          // Should this be raw instead?
```
vasco-santos, line 137:
```suggestion
```

PR #1456, opened 2024-05-15
fix: revert blob add by default

vasco-santos:
As we did a release while not ready

### DIFF
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -156,7 +156,7 @@ export class Client extends Base {
   /**
    * Get a receipt for an executed task by its CID.
    *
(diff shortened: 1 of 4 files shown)


packages/w3up-client/src/client.js
```diff
@@ -156,7 +156,7 @@ export class Client extends Base {
   /**
    * Get a receipt for an executed task by its CID.
    *
-   * @param {import('multiformats').UnknownLink} taskCid
+   * @param {import('multiformats').UnknownLink} taskCid - Task CID for receipt.
```
vasco-santos, line 159: to then trigger release please for w3up-client

alanshaw (approved):

PR #1459, opened 2024-05-15
fix: check for blob/accept receipts before blob/add is concluded

joaosa:
- Verifies that `blob/accept` has succeeded before concluding `blob/add` as per https://github.com/w3s-project/w3up/pull/1425#discussion_r1597298359. A location commitment is returned instead of the blob multihash.


packages/upload-client/test/helpers/utils.js
```diff
@@ -1,12 +1,67 @@
...
 import * as HTTP from '@web3-storage/capabilities/http'
 import * as W3sBlobCapabilities from '@web3-storage/capabilities/web3.storage/blob'
 import { W3sBlob } from '@web3-storage/capabilities'
 import { createConcludeInvocation } from '../../../upload-client/src/blob.js'
+import { Assert } from '@web3-storage/content-claims/capability'
 
 export const validateAuthorization = () => ({ ok: {} })
 
+// @ts-ignore Parameter
+export const setupGetReceipt = (contentGenFn) => {
```
vasco-santos, line 15: I would recommend to simplify this to either have a separate process mocking the getReceipt endpoint, or create a http server within this process instead of mocking the actual fetch implementation.

vasco-santos: in w3up-client we are using a mock server as a different process.
run process https://github.com/w3s-project/w3up/blob/604d300dde17521eea7333b949bb3f6796bd80a2/packages/w3up-client/package.json#L115
see that mock runs on npm run test https://github.com/w3s-project/w3up/blob/604d300dde17521eea7333b949bb3f6796bd80a2/packages/w3up-client/package.json#L109
https://github.com/w3s-project/w3up/blob/604d300dde17521eea7333b949bb3f6796bd80a2/packages/w3up-client/test/helpers/receipts-server.js

This is one way to go there, and if you see upload-client already uses this strategy for bucket-server https://github.com/w3s-project/w3up/blob/604d300dde17521eea7333b949bb3f6796bd80a2/packages/upload-client/package.json#L25 to write to a presigned URL

vasco-santos: there is an alternative that maybe is more on the direction you are taking that you could also consider.  you can rely on node http Server directly, freeway uses this kind of approach here https://github.com/web3-storage/freeway/blob/main/test/helpers/bucket.js. So, instead of a different process to run a mock server, the server runs in same process as the tests, and its URL and port are used by fetch  to perfrom requests against.

joaosa: thank you! I don't think the first approach would work as I need to load dynamic content into the receipt server. the 2nd approach should work and it does make sense, but I'm wondering if it's worth given the client refactor 🤔

packages/w3up-client/test/capability/blob.test.js
```diff
@@ -32,17 +35,30 @@ export const BlobClient = Test.withContext({
...
+        yield link
+      }),
+    })
 
     // TODO we should check blobsStorage as well
     assert.deepEqual(
-      await allocationsStorage.exists(space.did(), multihash.bytes),
+      await allocationsStorage.exists(
+        space.did(),
+        // @ts-ignore Element
```
vasco-santos, line 49: same as above, would be good to know more about this ignore, and use the expect-error to help out when it is not relevant anymore

vasco-santos (changes requested):

packages/upload-client/package.json
```diff
@@ -94,6 +94,7 @@
     "@types/varint": "^6.0.1",
     "@ucanto/principal": "^9.0.1",
     "@ucanto/server": "^10.0.0",
+    "@web3-storage/content-claims": "^4.0.4",
```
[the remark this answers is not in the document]
joaosa: addressed

packages/upload-client/src/blob.js
```diff
@@ -26,6 +27,41 @@ function createUploadProgressHandler(url, handler) {
   return onUploadProgress
 }
 
+// FIXME this code was copied over from w3up-client and modified to parameterise receiptsEndpoint
+export const receiptsEndpoint = 'https://up.web3.storage/receipt/'
+/**
```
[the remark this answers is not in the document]
joaosa: happy to solve that 🙏

packages/upload-client/src/blob.js
```diff
@@ -26,6 +27,41 @@ function createUploadProgressHandler(url, handler) {
   return onUploadProgress
 }
 
+// FIXME this code was copied over from w3up-client and modified to parameterise receiptsEndpoint
+export const receiptsEndpoint = 'https://up.web3.storage/receipt/'
```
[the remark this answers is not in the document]
joaosa: thanks that looks much better than before

packages/upload-client/src/blob.js
```diff
@@ -303,7 +339,55 @@ export async function add(
...
+      } catch (err) {
+        throw new Error(`failed ${BlobCapabilities.add.can} invocation`, {
+          cause: err,
+        })
+      }
+    },
+    {
+      onFailedAttempt: console.warn,
+      retries: options.retries ?? REQUEST_RETRIES,
+    }
```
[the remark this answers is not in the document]
joaosa: I have not tried this but that is a good point

packages/upload-client/test/index.test.js
```diff
@@ -258,6 +267,14 @@ describe('uploadFile', () => {
         // so we actually end up with a shard for each block - 5 CARs!
         shardSize: 1024 * 1024 * 2 + 1,
         onShardStored: (meta) => carCIDs.push(meta.cid),
+        fetch: setupGetReceipt(function* () {
+          yield link
+          yield link
+          yield link
+          yield link
+          yield link
+          yield link
```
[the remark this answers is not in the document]
joaosa: I addressed this particular kind of case and simplified the receipt generator on the upload-client side

packages/upload-client/test/index.test.js
```diff
@@ -453,6 +475,9 @@ describe('uploadDirectory', () => {
         onShardStored: (meta) => {
           carCID = meta.cid
         },
+        fetch: setupGetReceipt(() => {
+          return links[0]
```
vasco-santos, line 479: could we by the path from the url used in fetch figure out which one is the CID for this link instead?

packages/upload-client/test/helpers/utils.js
```diff
@@ -1,12 +1,65 @@
...
+      !url.pathname ||
+      (url.pathname.contains && !url.pathname.contains('/receipt/'))
+    ) {
+      return await fetch(url, options)
+    }
+
+    const taskID = url.pathname.replace('/receipt/', '')
+    const issuer = await Signer.generate()
+
+    const content = contentGen()
```
vasco-santos, line 30: I think this content gen should pass the CID of the receipt being requested do that caller does what it needs

packages/upload-client/src/blob.js
```diff
@@ -26,6 +27,40 @@ function createUploadProgressHandler(url, handler) {
   return onUploadProgress
 }
 
+// FIXME this code was copied over from w3up-client and modified to parameterise receiptsEndpoint
```
vasco-santos, line 30: I though I wrote this earlier, but can't find it now. Can we make w3up to use this function from `upload-client` instead of having same code in both places?

vasco-santos (changes requested):

packages/upload-client/src/blob.js
```diff
@@ -26,6 +27,41 @@ function createUploadProgressHandler(url, handler) {
...
+ * Get a receipt for an executed task by its CID.
+ *
+ * @param {import('multiformats').UnknownLink} taskCid
+ * @param {import('./types.js').RequestOptions} [options]
+ */
+async function getReceipt(taskCid, options = {}) {
+  // Fetch receipt from endpoint
+  const url = new URL(
+    taskCid.toString(),
+    options.receiptsEndpoint ?? receiptsEndpoint
```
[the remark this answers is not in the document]
vasco-santos: yes, I think this is the pattern here of defaulting to prod. So would say is good

packages/upload-client/src/blob.js
```diff
@@ -303,7 +338,57 @@ export async function add(
...
   }
 
-  return multihash
+  // Ensure the blob has been accepted
+  const acceptReceipt = await retry(
+    async () => {
+      try {
+        return await getReceipt(nextTasks.accept.task.link(), options)
+      } catch (err) {
+        throw new Error(`failed ${BlobCapabilities.add.can} invocation`, {
```
alanshaw, line 347: Is this the correct message here?

alanshaw: It would be nice to distinguish between not there yet vs an actual error, and quit early from retry if the latter. Typically new code returns `Result` types, which allow us to properly type possible error states. So we might do something like:

```js
const acceptReceipt = await retry(async () => {
  const res = await getReceipt(nextTasks.accept.task.link(), options)
  if (res.error) {
    if (res.error.name === 'ReceiptNotFound') {
      // throw an error that will cause `p-retry` to retry with
      throw new Error('receipt not yet available')
    } else {
      // bail the retry using `import { AbortError } from 'p-retry'`
      throw new AbortError(new Error('failed to fetch blob/accept receipt', { cause: res.error }))
    }
  }
  return res.ok
})
```

Not 100% necessary but I would then consider extracting this as a utility like:

```js
const acceptReceipt = await Receipt.poll(nextTasks.accept.task.link(), options)
```

packages/upload-client/test/blob.test.js
```diff
@@ -177,6 +187,138 @@ describe('Blob.add', () => {
...
+    await assert.rejects(
+      Blob.add(
+        { issuer: agent, with: space.did(), proofs, audience: serviceSigner },
+        bytes,
+        {
+          connection,
+          retries: 0,
+          fetch: async (url) => {
+            // @ts-ignore Parameter
+            if (!url.pathname) {
```
alanshaw, line 307: There is always a `pathname` in URL no? It's `/` if no path...

alanshaw: I'm guessing this works because you're only using this value for `fetch` in the receipts call.

If yes, the `fetch` option should not be documented in the types since it's not used for all requests made by `Blob.add()`.

This is convenient for testing but I think spinning up an actual server to do this would be better.

packages/upload-client/src/blob.js
joaosa: addressed. I think I somehow mixed it with doing the same for the receipts endpoint

packages/upload-client/test/index.test.js
alanshaw: Oh yes that would be nicer, and allow it to respond appropriately when not found.

packages/w3up-client/test/helpers/utils.js
```diff
@@ -1,14 +1,69 @@
...
 import * as Types from '../../src/types.js'
+import * as Signer from '@ucanto/principal/ed25519'
+import { Assert } from '@web3-storage/content-claims/capability'
 
 export const validateAuthorization = () => ({ ok: {} })
 
+// @ts-ignore Parameter
+export const setupGetReceipt = (contentGenFn) => {
+  // @type {Generator<import('multiformats').Link,void,unknown>} contentGenFn
+  const contentGen = contentGenFn()
```
alanshaw, line 19: I'm not sure I understand why this is an async generator...why not just pass an array of links?

alanshaw (changes requested):

packages/upload-client/test/index.test.js
joaosa: sadly, I don't think we can as we would need to compute the task ID from the initial links. I think it's best I remove this altogether

packages/upload-client/src/blob.js
joaosa: working on this 🙏

packages/upload-client/test/blob.test.js
joaosa: this is gone now

packages/upload-client/src/blob.js
joaosa: done

packages/upload-client/src/blob.js
```diff
@@ -303,7 +304,46 @@ export async function add(
...
+    nextTasks.accept.task.link()
+  )
+
+  // @ts-ignore Property
+  /* c8 ignore next 5 */
+  if (!acceptReceipt?.out.ok?.site) {
+    throw new Error(`failed ${BlobCapabilities.add.can} invocation`, {
+      cause: 'failed to get blob/accept receipt',
+    })
+  }
```
vasco-santos, line 318: nit: would be good to distinct this error, from Not Found, or Other errors

packages/upload-client/src/index.js
```diff
@@ -144,9 +145,9 @@ async function uploadBlockStream(
           async transform(car, controller) {
             const bytes = new Uint8Array(await car.arrayBuffer())
             // Invoke blob/add and write bytes to write target
-            const multihash = await Blob.add(conf, bytes, options)
+            const { multihash } = await Blob.add(conf, bytes, options)
```
vasco-santos, line 148: let's make sure to create an issue where we can follow up to have user make the site delegation public by default by invoking the Claim delegation

packages/upload-client/src/receipts.js
```diff
@@ -0,0 +1,110 @@
...
+   */
+  async poll(taskCid) {
+    return await retry(
+      async () => {
+        const res = await this.get(taskCid)
+        if (res.error) {
+          // @ts-ignore
+          if (res.error.name === 'ReceiptNotFound') {
+            // throw an error that will cause `p-retry` to retry with
+            throw new Error('blob/accept receipt not yet available')
```
vasco-santos, line 52: can we propagate this error so as receipt not found so that consumer can know the real error and give different info to user? Like a not found would probably mean more of a problem, while other error may mean some downtime on the service

packages/upload-client/src/receipts.js
```diff
@@ -0,0 +1,110 @@
...
+   * @returns {Promise<import('@ucanto/client').Result<import('@ucanto/interface').Receipt, Error>>}
+   */
+  async get(taskCid) {
+    // Fetch receipt from endpoint
+    const url = new URL(taskCid.toString(), this.receiptsEndpoint)
+    const workflowResponse = await this.fetch(url)
+    /* c8 ignore start */
+    if (!workflowResponse.ok) {
+      return {
+        error: new ReceiptNotFound(taskCid),
```
vasco-santos, line 84: shouldn't this only be on not response ok + Status 404?

packages/upload-client/src/receipts.js
```diff
@@ -0,0 +1,110 @@
...
+    // Decode message
+    const agentMessage = await CAR.request.decode({
+      body: agentMessageBytes,
+      headers: {},
+    })
+    // Get receipt from the potential multiple receipts in the message
+    const receipt = agentMessage.receipts.get(taskCid.toString())
+    if (!receipt) {
+      return {
+        error: new Error(
```
vasco-santos, line 101: would be great to have a specific error like the ReceiptNotFound for this one

packages/upload-client/src/blob.js
```diff
@@ -303,7 +304,46 @@ export async function add(
...
+  /* c8 ignore next 5 */
+  if (!site) {
+    throw new Error(`failed ${BlobCapabilities.add.can} invocation`, {
+      cause: 'failed to get blob/accept receipt',
+    })
+  }
+
+  return {
+    multihash,
+    site: site.link(),
```
vasco-santos, line 345: we should return the delegation, so that the client can invoke it if intended

packages/upload-client/test/blob.test.js
```diff
@@ -74,7 +75,7 @@ describe('Blob.add', () => {
 
     /** @type {import('../src/types.js').ProgressStatus[]} */
     const progress = []
-    const multihash = await Blob.add(
+    const { multihash } = await Blob.add(
```
vasco-santos, line 78: we should assert the site exists, and maybe it has capability with this multihash

packages/upload-client/test/helpers/receipts-server.js
```diff
@@ -0,0 +1,68 @@
...
+    },
+  })
+
+  const message = await Message.build({
+    receipts: [receipt],
+  })
+  return CAR.request.encode(message).body
+}
+
+const server = createServer(async (req, res) => {
```
vasco-santos, line 47: thanks for putting this together, I think it makes everything much simpler ✨

packages/upload-client/test/index.test.js
```diff
@@ -674,27 +680,39 @@ describe('uploadDirectory', () => {
...
+    ]
     const unsortedFiles = [
-      new File([await randomBytes(32)], '/b.txt'),
-      new File([await randomBytes(32)], '/b.txt'),
-      new File([await randomBytes(32)], 'c.txt'),
-      new File([await randomBytes(32)], 'a.txt'),
+      new File([bytesList[0]], '/b.txt'),
+      new File([bytesList[1]], '/b.txt'),
+      new File([bytesList[2]], 'c.txt'),
+      new File([bytesList[3]], 'a.txt'),
```
vasco-santos, line 693: probably we can revert this back?

vasco-santos (changes requested):
Thanks for this new iteration! We are quite quite close 🎉  I will test this against staging as is, most changes in review request should be quite simple now and not affect so much the end flow

packages/upload-api/src/blob/add.js
```diff
@@ -5,7 +5,6 @@ import * as Blob from '@web3-storage/capabilities/blob'
 import * as W3sBlob from '@web3-storage/capabilities/web3.storage/blob'
 import * as HTTP from '@web3-storage/capabilities/http'
 import * as API from '../types.js'
-
```
vasco-santos, line 8: could we take this out to avoid a `upload-api` release?

packages/upload-client/src/blob.js
joaosa: addressed

packages/upload-client/src/receipts.js
joaosa: addressed

packages/upload-client/src/receipts.js
joaosa: true! addressed

packages/upload-client/src/receipts.js
joaosa: addressed

packages/upload-client/src/blob.js
joaosa: good point! addressed

packages/upload-client/test/index.test.js
joaosa: addressed

packages/upload-api/src/blob/add.js
joaosa: addressed

packages/upload-client/src/index.js
joaosa: created https://github.com/w3s-project/w3up/issues/1486 👍

packages/upload-client/src/receipts.js
```diff
@@ -0,0 +1,128 @@
...
+   */
+  async poll(taskCid) {
+    return await retry(
+      async () => {
+        const res = await this.get(taskCid)
+        if (res.error) {
+          // @ts-ignore
+          if (res.error.name === 'ReceiptNotFound') {
+            // throw an error that will cause `p-retry` to retry with
+            throw new ReceiptNotFound(taskCid)
```
vasco-santos, line 72:
```suggestion
            throw res.error
```

this should be possible right? And we do not add additional entries to the stack trace

packages/upload-client/src/receipts.js
```diff
@@ -0,0 +1,128 @@
...
+   * @param {import('multiformats').UnknownLink} taskCid
+   */
+  constructor(taskCid) {
+    super()
+    this.taskCid = taskCid
+  }
+
+  /* c8 ignore start */
+  get reason() {
+    return `receipt not found for task ${this.taskCid}`
```
vasco-santos, line 17:
```suggestion
    return `receipt not found for task ${this.taskCid} in the indexed workflow`
```

vasco-santos (approved):
LGTM ❤️

packages/upload-client/src/blob.js
```diff
@@ -303,7 +304,38 @@ export async function add(
...
+    [...acceptReceipt.iterateIPLDBlocks()].map((block) => [
+      `${block.cid}`,
+      block,
+    ])
+  )
+  const site = Delegation.view(
+    {
+      root: /** @type {import('@ucanto/interface').UCANLink} */ (
+        // @ts-ignore Property
+        acceptReceipt.out.ok.site
```
alanshaw, line 322: You should check for `error` and then you don't need to `ts-ignore` here.

packages/upload-client/src/blob.js
```diff
@@ -303,7 +304,38 @@ export async function add(
...
+      ),
+      blocks,
+    },
+    null
+  )
+  /* c8 ignore next 5 */
+  if (!site) {
```
alanshaw, line 329: Why would this be falsey?

alanshaw: Oh you need to not pass `null` above (pass `undefined` or simply omit) for `view(...)` to throw in the case where the root does not exist in the passed blocks.

packages/upload-client/src/index.js
```diff
@@ -221,3 +217,19 @@ async function uploadBlockStream(
...
 }
+
+/**
+ * Indexes a sharded DAG
+ *
+ * @param {import('multiformats').Link} root
+ * @param {import('./types.js').CARLink[]} shards
+ * @param {Array<Map<import('./types.js').SliceDigest, import('./types.js').Position>>} shardIndexes
+ */
+export async function indexShardedDAG(root, shards, shardIndexes) {
```
alanshaw, line 228: This should be an export from the `blob-index` utils.

packages/upload-client/src/receipts.js
```diff
@@ -0,0 +1,128 @@
...
+   *
+   * @param {import('multiformats').UnknownLink} taskCid
+   * @returns {Promise<import('@ucanto/client').Result<import('@ucanto/interface').Receipt, Error>>}
+   */
+  async get(taskCid) {
+    // Fetch receipt from endpoint
+    const url = new URL(taskCid.toString(), this.receiptsEndpoint)
+    const workflowResponse = await this.fetch(url)
+    /* c8 ignore start */
+    if (!workflowResponse.ok && workflowResponse.status === 404) {
```
alanshaw, line 102:
```suggestion
    if (workflowResponse.status === 404) {
```

packages/upload-client/src/receipts.js
```diff
@@ -0,0 +1,128 @@
...
+    const agentMessageBytes = new Uint8Array(
+      await workflowResponse.arrayBuffer()
+    )
+    // Decode message
+    const agentMessage = await CAR.request.decode({
+      body: agentMessageBytes,
+      headers: {},
+    })
+    // Get receipt from the potential multiple receipts in the message
+    const receipt = agentMessage.receipts.get(taskCid.toString())
```
alanshaw, line 118: nit: It's kinda odd that this endpoint returns an `AgentMessage` but is not a ucanto endpoint...

I wonder if this should just always return an agent message (but with no receipts in it if not found) and then you wouldn't have to check for HTTP 404 status as well.

alanshaw (changes requested):

packages/upload-client/src/blob.js
joaosa: I don't think I need at all anymore, as previously the receipt acquisition method could return null. Now it's either a receipt or it throws.

packages/upload-client/src/receipts.js
vasco-santos: This is because we in the past decided to redirect to the workflow... We will need to UCANify the gets for workflows/receipts and fix this

packages/upload-client/src/receipts.js
```diff
@@ -0,0 +1,128 @@
...
+export class Receipt {
+  /**
+   * @param {import('./types.js').RequestOptions} [options]
+   */
+  constructor(options = {}) {
+    /* c8 ignore start */
+    this.receiptsEndpoint = options.receiptsEndpoint ?? receiptsEndpoint
+    this.retries = options.retries ?? REQUEST_RETRIES
+    this.fetch = options.fetch ?? globalThis.fetch.bind(globalThis)
+    /* c8 ignore stop */
```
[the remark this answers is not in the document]
joaosa: alright sounds good. addressed as well

packages/upload-client/src/blob.js
joaosa: addressed

packages/upload-client/src/index.js
joaosa: addressed

alanshaw (approved):
LGTM

packages/upload-client/src/receipts.js
alanshaw: Ah ok gotcha!

PR #1469, opened 2024-05-16
feat!: add download URL method to `BlobsStorage`

alanshaw:
Ensures we don't hardcode the downlaod URL in the implementation.


vasco-santos (approved):
LGTM 

could you add a test for the blob-storage to get the URL? https://github.com/w3s-project/w3up/blob/main/packages/upload-api/test/storage/blobs-storage-tests.js

alanshaw:
Oh, yeah sorry I completely forgot!

PR #1479, opened 2024-05-28
feat!: updates agent-store api to unblock integration with w3infra

Gozala:
While working on https://github.com/w3s-project/w3infra/pull/380 I have realized that agent-store API was poorly designed because it imposed:

1. Store to traverse agent message in order to index each invocations and receipts within it and code that dealt with traversal was trapped in util module here that was not even exported.
2. Store has to encode message into bytes in order to persist it, which is redundant given that we received message in encoded form.

This PR fixes above limitations by switching store interface from receiving `AgentMessage` to `ParsedAgentMessage` which wraps `AgentMessage` along with message bytes and index freeing store from doing any kind of traversal or encoding.

I also removed legacy code that was left behind by previous PR.


vasco-santos (approved):
LGTM

PR #1480, opened 2024-05-29
feat: use digest in `blob/accept` location commitment

alanshaw:
This upgrades the content-claims dependency, allowing us to use multihash digest in location claim. It brings implementation in line with the spec.


vasco-santos (approved):
LGTM

PR #1484, opened 2024-05-30
feat: add blob/get

joaosa:
Adds support for `blob/get` as defined in https://github.com/w3s-project/specs/pull/126

### DIFF
--- a/packages/capabilities/src/blob.js
+++ b/packages/capabilities/src/blob.js
@@ -137,6 +137,41 @@ export const list = capability({
   },
 })
 
+/**
+ * Capability can be used to get the stored Blob from the (memory)
+ * space identified by `with` field.
+ */
+export const get = capability({
+  can: 'space/blob/get',
+  /**
+   * DID of the (memory) space where Blob is stored.
+   */
+  with: SpaceDID,
+  nb: Schema.struct({
+    /**
+     * A multihash digest of the blob payload bytes, uniquely identifying blob.
+     */
+    digest: Schema.bytes(),
+  }),
+  derives: (claimed, delegated) => {
+    if (claimed.with !== delegated.with) {
+      return fail(
+        `Expected 'with: "${delegated.with}"' instead got '${claimed.with}'`
+      )
+    } else if (
+      delegated.nb.digest &&
+      !equals(delegated.nb.digest, claimed.nb.digest)
+    ) {
+      return fail(
+        `Link ${
+          claimed.nb.digest ? `${claimed.nb.digest}` : ''
+        } violates imposed ${delegated.nb.digest} constraint.`
+      )
+    }
+    return ok({})
+  },
+})
+
 // ⚠️ We export imports here so they are not omitted in generated typedefs
 // @see https://github.com/microsoft/TypeScript/issues/51548
 export { Schema }
--- a/packages/capabilities/src/blob.js
+++ b/packages/capabilities/src/blob.js
@@ -142,7 +142,7 @@ export const list = capability({
  * space identified by `with` field.
  */
 export const get = capability({
-  can: 'space/blob/get',
+  can: 'space/blob/get/0/1',
   /**
    * DID of the (memory) space where Blob is stored.
    */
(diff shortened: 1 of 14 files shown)


packages/upload-api/src/blob/get.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * @param {API.BlobServiceContext} context
+ * @returns {API.ServiceMethod<API.BlobGet, API.BlobGetSuccess, API.BlobGetFailure>}
+ */
+export function blobGetProvider(context) {
+  return Server.provide(Blob.get, async ({ capability }) => {
+    const { digest } = capability.nb
+    if (!digest) {
+      return Server.fail('nb.digest must be set')
+    }
```
vasco-santos, line 16: the validator from the capability should already guarantee this validation. Did you have any typescript error? This should really not be needed

packages/upload-client/test/blob.test.js
```diff
@@ -733,3 +733,113 @@ describe('Blob.remove', () => {
     )
   })
 })
+
+describe('Blob.get', () => {
+  it('get a stored CAR file', async () => {
```
vasco-santos, line 738:
```suggestion
  it('get a stored Blob', async () => {
```

packages/upload-client/test/blob.test.js
```diff
@@ -733,3 +733,113 @@ describe('Blob.remove', () => {
...
+            assert.equal(invocation.issuer.did(), agent.did())
+            assert.equal(invocation.capabilities.length, 1)
+            const invCap = invocation.capabilities[0]
+            assert.equal(invCap.can, BlobCapabilities.get.can)
+            assert.equal(invCap.with, space.did())
+            assert.equal(String(invCap.nb?.digest), bytesHash.bytes)
+            return {
+              ok: {
+                blob: { digest: bytesHash.bytes, size: bytes.length },
+              },
```
vasco-santos, line 767: shouldn't `cause` be needed?

vasco-santos (changes requested):

packages/upload-api/src/blob/get.js
joaosa: I did not. I ended up adjusting the `store/add` code and that's why we got this check here. addressed!

packages/capabilities/src/blob.js
```diff
@@ -137,6 +137,41 @@ export const list = capability({
   },
 })
 
+/**
+ * Capability can be used to get the stored Blob from the (memory)
+ * space identified by `with` field.
+ */
+export const get = capability({
+  can: 'space/blob/get',
```
Gozala, line 145: After talking with @joaosa and @alanshaw I feel more confident that it would be best to take approach outlined here https://github.com/w3s-project/specs/pull/126#issuecomment-2140178593

In order to avoid blocking on the ☝️ and ship things sooner I suggest use of namespace that we can easily discard in the future e.g. 

```suggestion
  can: 'space/blob/get@0.1',
```

That way in the future we could add `space/blob/get` that would simply get return receipt pointer, without breaking clients that we have shipped.

Gozala (approved):
Approving preconditioned on capability rename (details inline).

packages/capabilities/src/blob.js
Gozala: Note that funky `/0/1` implies that one could delegate major, minor and patch versions, which would not be possible in more traditional encoding like `space/blob/get@0.1`

packages/upload-api/src/blob/accept.js
```diff
@@ -50,7 +52,7 @@ export function blobAcceptProvider(context) {
         audience: DID.parse(space),
         with: context.id.toDIDKey(),
         nb: {
-          content: { digest: digest.bytes },
+          content: Link.create(rawCode, digest),
```
vasco-santos, line 55: Alan just did this change to use `digest`: https://github.com/w3s-project/w3up/pull/1480

This is supposed to use digest now, did you find any issue? Maybe you need to update dependencies

vasco-santos (changes requested):

packages/upload-api/src/blob/accept.js
joaosa: one of the tests was failing 🤔 . I'll try to fix it

alanshaw (approved):

packages/upload-api/src/blob/accept.js
joaosa: alright this should be addressed

vasco-santos (approved):
LGTM

vasco-santos (approved):

PR #1487, opened 2024-06-04
feat: publish index claim

alanshaw:
This PR updates the `index/add` handler to publish an [index claim](https://github.com/w3s-project/content-claims?tab=readme-ov-file#index-claim) for the added index.

Reading an index from the index claim allows freeway and hoverboard to serve blocks from all of the shards without having to request a location claim for each block.

refs https://github.com/w3s-project/blob-fetcher/pull/10


packages/upload-api/test/handlers/index.js
```diff
@@ -171,4 +171,78 @@ export const test = {
     assert.ok(receipt.out.error)
     assert.equal(receipt.out.error?.name, 'ShardNotFound')
   },
+  'index/add should publish index claim': async (assert, context) => {
+    const { proof, spaceDid } = await registerSpace(alice, context)
+    const contentCAR = await randomCAR(32)
```
vasco-santos, line 176: nit: should we randomBlob instead since we upload blob later and uniform more towards new world?

packages/upload-api/src/index/add.js
```diff
@@ -87,3 +96,17 @@ const assertAllocated = async (context, space, digest, errorName) => {
...
+/**
+ * @param {API.ClaimsClientContext} ctx
+ * @param {{ content: API.UnknownLink, index: API.CARLink }} params
+ */
+const publishIndexClaim = async (ctx, { content, index }) => {
+  const { invocationConfig, connection } = ctx.claimsService
+  const { issuer, audience, with: resource, proofs } = invocationConfig
+  const expiration = Infinity
+  const nb = { content, index }
+  const conf = { issuer, audience, with: resource, nb, expiration, proofs }
```
vasco-santos, line 109: nit: I would add this in the `invoke` with new lines per property to be more visible

vasco-santos (approved):
LGTM

packages/upload-api/test/handlers/index.js
alanshaw: Hmm, `index/add` only really makes sense if you're storing a CAR though...I can't create an index for a `randomBlob`.

PR #1495, opened 2024-06-04
fix: reuse http/put and blob/accept receipts when available

joaosa:
From @vasco-santos:
1. if we already have a http/put receipt, we should not send one
2. if we already receive in next a blob accept receipt, we should not await on the receipt generation

### DIFF
--- a/packages/upload-client/src/blob.js
+++ b/packages/upload-client/src/blob.js
@@ -280,28 +280,29 @@ export async function add(
   }
 
   // Invoke `conclude` with `http/put` receipt
-  const derivedSigner = ed25519.from(
-    /** @type {import('@ucanto/interface').SignerArchive<import('@ucanto/interface').DID, typeof ed25519.signatureCode>} */
-    (nextTasks.put.task.facts[0]['keys'])
-  )
-
-  const httpPutReceipt = await Receipt.issue({
-    issuer: derivedSigner,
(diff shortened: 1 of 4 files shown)


packages/upload-client/src/blob.js
```diff
@@ -280,32 +280,36 @@ export async function add(
...
+      result: { ok: {} },
+    })
+  }
   const httpPutConcludeInvocation = createConcludeInvocation(
     issuer,
     // @ts-expect-error object of type unknown
     audience,
     httpPutReceipt
   )
   const ucanConclude = await httpPutConcludeInvocation.execute(conn)
```
joaosa, line 301: I am thinking this call should only be executed if we're missing a http/put receipt. however, I'm leaving it here as a "validation" the receipt we got makes sense.

alanshaw: hmm, I think we don't need to conclude if we already have the receipt for a concluded task.

@vasco-santos is that right?

alanshaw (changes requested):

packages/upload-client/src/blob.js
joaosa: addressed

alanshaw (approved):

vasco-santos (approved):

PR #1499, opened 2024-06-06
chore: disable docs task on release

alanshaw:
This hasn't been working for a long time now and I don't think  we even use the docs it generates.


.github/workflows/release.yml
```diff
@@ -51,20 +51,20 @@ jobs:
...
-      contains(fromJson(needs.release.outputs.paths_released), 'packages/blob-index') ||
-      contains(fromJson(needs.release.outputs.paths_released), 'packages/capabilities') ||
-      contains(fromJson(needs.release.outputs.paths_released), 'packages/did-mailto') ||
-      contains(fromJson(needs.release.outputs.paths_released), 'packages/upload-client') ||
-      contains(fromJson(needs.release.outputs.paths_released), 'packages/upload-api') ||
-      contains(fromJson(needs.release.outputs.paths_released), 'packages/filecoin-client') ||
-      contains(fromJson(needs.release.outputs.paths_released), 'packages/filecoin-api') ||
-      contains(fromJson(needs.release.outputs.paths_released), 'packages/w3up-client')
-    uses: './.github/workflows/reusable-deploy-docs.yml'
+  # docs:
```
joaosa, line 54: Wouldn't it be cleaner to delete it? One can always restore from history if needed

PR #1506, opened 2024-06-13
feat: add use cases package

travis:
Add an example of how to delegate all capabilities on a space to a `did:mailto` account. This has come up a few times recently and it's one potential solution to the "I want to change my email address" problem.

I'd like to start collecting use cases for w3up client here. It might be a good idea to run some or all of them as integration tests.

### DIFF
--- a/packages/use-cases/README.md
+++ b/packages/use-cases/README.md
@@ -1,0 +1,3 @@
+# Use Cases
+
+This package is a "cookbook" of different uses for `@web3-storage/w3up-client`.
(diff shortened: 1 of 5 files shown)


alanshaw:
I made this https://github.com/storacha-network/w3up-examples (very similar timing lol).

I don't really care much for whether they are part of this repo or separate, but they should all be together.

Note: I'd probably lean more towards separate repo in order to not have loads of frontend dependencies or other dependencies used only in examples when `npm install`ing.

PR #1507, opened 2024-06-17
feat!: allow invocation configuration to be generated on demand

alanshaw:
Since the blob protocol landed, it's become harder to obtain a delegation to upload a specific CAR.

It was tricky before because of sharding, but assuming you created a CAR below the shard threshold, its easy enough to obtain a `store/add` delegation tied to the CAR CID.

With the blob protocol, we now generate an index, and `blob/add` that as well. It means we can't just create a CAR and delegate `can: blob/add, nb: { link, size }` for the CAR, because we also need to delegate `blob/add` for the index, but we don't know the index CID or it's size up front...

It's ok if the delegation is not restricted to the specific CAR, but for certain use cases it's necessary to delegate just what is needed.

This PR alters the client to allow invocation configuration to be generated on demand, when the proofs are required. This is a backwards compatible change.

So, for example instead of calling: `uploadFile({ issuer, with, proofs }, file)` you now call:

```js
const configure = caps => {
  const proofs = []
  for (const { can, nb } of caps) {
    // decide if want to delegate and create delegation
    proofs.push(...)
  }
  if (!proofs.length) throw new Error('no proofs available')
  return { issuer, with, proofs }
}

uploadFile(configure, file)
```

...and `configure` will be called for each proof that is needed (`blob/add`, `index/add` and `upload/add`).

It means also that we no longer have to recommend folks to create a CAR, get the CID and then upload it, they can just use `uploadFile` and pass in the invocation config function to obtain the correct delegations when required.


packages/upload-client/src/blob/add.js
```diff
@@ -165,19 +161,20 @@ export function createConcludeInvocation(id, serviceDid, receipt) {
...
+ * @param {import('multiformats').MultihashDigest} digest
  * @param {Blob|Uint8Array} data Blob data.
- * @param {import('./types.js').RequestOptions} [options]
- * @returns {Promise<import('./types.js').BlobAddOk>}
+ * @param {import('../types.js').RequestOptions} [options]
+ * @returns {Promise<import('../types.js').BlobAddOk>}
  */
 export async function add(
   { issuer, with: resource, proofs, audience },
+  digest,
```
vasco-santos, line 171: I think this needs to be updated: https://github.com/storacha-network/w3up/tree/main/packages/w3up-client#capabilityblobadd

vasco-santos (approved):
LGTM, I am curious if there is other option explored with the client refactor we can consider.

Worth adding docs for this to the readme, maybe with an example

packages/upload-client/src/blob/add.js
alanshaw: I did not change this... w3up client takes hash of blob.

PR #1521, opened 2024-07-22
feat(filecoin-api): paginated queries

alanshaw:
This PR updates filecoin-api to add pagination to the query interfaces. Currently only the first page of results is returned.

This means in the fileocin pipeline cron jobs we never process more than 1 page of items, even though there are sometimes many more items to process and ample execution time remaining in the lambda.

If only the first page of items is processed it gives a false sense of everything operating ok, when actually things may be backing up. It essentially smothers the issue. If the lambda times out because there are too many items to process then this is a good indication that it needs some attention.

Essentially, the interface changes from:

```ts
interface Queryable<Q, T> {
  query (q: Q): Promise<Result<T[]>>
}
```

To this:

```ts
interface Pageable {
  cursor?: string
  size?: number
}

interface ListSuccess<T> {
  results: T[]
  cursor?: string
}

interface Queryable<Q, T> {
  query (q: Q, options?: Pageable): Promise<Result<ListSuccess<T>>>
}
```

Context: we currently have a _lot_ of `filecoin/accept` receipts we need to issue that are being done in a cron job.

### DIFF
--- a/packages/filecoin-api/src/aggregator/api.ts
+++ b/packages/filecoin-api/src/aggregator/api.ts
@@ -13,17 +13,15 @@ import {
   ServiceConfig,
 } from '../types.js'
 
-export type PieceStore = UpdatableStore<PieceRecordKey, PieceRecord>
+export type PieceStore = Store<PieceRecordKey, PieceRecord> &
+  UpdatableStore<PieceRecordKey, PieceRecord>
 export type PieceQueue = Queue<PieceMessage>
 export type BufferQueue = Queue<BufferMessage>
 export type BufferStore = Store<Link, BufferRecord>
 export type AggregateStore = Store<AggregateRecordKey, AggregateRecord>
 export type PieceAcceptQueue = Queue<PieceAcceptMessage>
-export type InclusionStore = QueryableStore<
-  InclusionRecordKey,
-  InclusionRecord,
-  InclusionRecordQueryByGroup
->
+export type InclusionStore = Store<InclusionRecordKey, InclusionRecord> &
+  QueryableStore<InclusionRecordQueryByGroup, InclusionRecord>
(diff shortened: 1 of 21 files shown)


packages/filecoin-api/src/aggregator/api.ts
```diff
@@ -13,17 +13,15 @@ import {
...
 export type BufferStore = Store<Link, BufferRecord>
 export type AggregateStore = Store<AggregateRecordKey, AggregateRecord>
 export type PieceAcceptQueue = Queue<PieceAcceptMessage>
-export type InclusionStore = QueryableStore<
-  InclusionRecordKey,
-  InclusionRecord,
-  InclusionRecordQueryByGroup
->
+export type InclusionStore = Store<InclusionRecordKey, InclusionRecord> &
+  QueryableStore<InclusionRecordQueryByGroup, InclusionRecord>
```
alanshaw, line 24: I flattened the store types to give us more flexibility and simplicity. They can be composed as shown and we now have the ability to express in types a read only store for querying data (for example).

hannahhoward: +1 here for composable types :)

hannahhoward (approved):
LGTM

Overall this looks really good. Appreciate the type changes and the paging interface looks really good. Do we have adequate integration testing to insure a safe deploy?

alanshaw:
> Do we have adequate integration testing to insure a safe deploy?

Yeah, I'm confident this is safe to deploy. We already have tests that cover the code that uses `.query(...)`. I added coverage reporting to double check:

Screenshot 2024-07-23 at 10 30 52
Screenshot 2024-07-23 at 10 29 54
Screenshot 2024-07-23 at 10 28 35

PR #1534, opened 2024-08-30
chore: Warn if mocks aren't running

Peeja:
It's not obvious that the mocks need to be running for tests to run. Now the output will warn you and tell you how to run them.


hannahhoward (approved):
Also LGTM

PR #1535, opened 2024-08-30
refactor: `BlobLike`'s `stream()` matches `Blob`'s

Peeja:
Should return a `ReadableStream<Uint8Array>`, not a `ReadableStream<any>`. Get it straight from `Blob` to be sure it matches.

This bit me while trying to work with the API.


hannahhoward (approved):
LGTM

PR #1545, opened 2024-09-06
refactor: Avoid conflation of Space and Account

Peeja:
* `space.signer.withDID(account)` gives us a Signer with the Space's key, but which reports the Account's DID. It looks like this might have been needed at some point when this thing was actually used to sign something. But as of now, we don't need a *Signer*, just a DID, and a `did:mailto:` is fine. So, simplify this.

* This was being passed as the `agent`, but it's not an Agent. `createAuthorization()` doesn't need an Agent, it needs an *audience*. In some cases that's an Agent, in some cases it's an Account. This changes the name to clarify.

### DIFF
--- a/.github/workflows/access-client.yml
+++ b/.github/workflows/access-client.yml
@@ -28,8 +28,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: pnpm/action-setup@v4
-        with:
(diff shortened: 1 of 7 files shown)


alanshaw (approved):
LGTM, can we have a separate PR for the actions changes next time please?

PR #1546, opened 2024-09-06
feat(client): optional account recovery

fforbeck:
Issue https://github.com/storacha-network/project-tracking/issues/124

# Current Workflow (CLI)

### `w3cli` Project
The [`space.create(name, options)`](https://github.com/storacha-network/w3cli/blob/3f59da7b096a63f14def9946682160fefacd0702/space.js#L24) function in `space.js`  allows the user to pass the following options:
- **recovery**: `false` (default)
- **caution**: `false` (default)
- **customer**: `Email | false`
- **account**: `string | false`

If the `account` attribute is provided, the following steps are triggered:
1. `space.createdRecovery` is called.
2. Then, `client.capabilities.access.delegate` is executed.

This process occurs during **space creation**, not during provisioning.  
Provisioning only happens when the `customer` attribute is provided in the `options` argument.

---

# Issue (JS Client)

### `w3up` Project
The [`client.createSpace(name)`](https://github.com/storacha-network/w3up/blob/fb8b8677c4c633cdf8c259db55357a1794eed3ab/packages/w3up-client/src/client.js#L239) function in the `w3up` project currently does **not** accept any options, which means the recovery account must be created manually **after** space creation and provisioning. This introduces a risk of forgetting the manual step and potentially losing access to the space.

---

# Expected Outcome

We propose a small modification to the `client.createSpace(name)` function to support optional parameters, allowing clients to pass an `account` directly, automating the recovery setup.

### Solution:
- Clients can now call `client.createSpace(name, { account })` to include the `account` in the creation process.
- This ensures that the recovery account is created immediately after the space is created, provisioned, and saved, eliminating the need for a manual step.

### Benefits:
- **Backward Compatibility**: The change does not break existing client implementations since the `account` parameter is optional, keeping calls like `client.createSpace(name)` intact.
- **Simplified Workflow**: By simply passing the `account` when calling `createSpace`, the recovery account setup is handled automatically during space creation.
- **Flexibility**: If needed in the future, the `account` attribute can be made required to enforce stricter recovery account management.

### DIFF
--- a/.github/workflows/w3up-client.yml
+++ b/.github/workflows/w3up-client.yml
@@ -31,8 +31,6 @@ jobs:
         uses: actions/checkout@v3
       - name: Install
         uses: pnpm/action-setup@v4
-        with:
-          version: 9.x.x
       - name: Setup
         uses: actions/setup-node@v3
         with:
--- a/.github/workflows/w3up-client.yml
+++ b/.github/workflows/w3up-client.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Install
         uses: pnpm/action-setup@v4
         with:
-          version: 9
+          version: 9.x.x
       - name: Setup
         uses: actions/setup-node@v3
         with:
--- a/packages/w3up-client/README.md
+++ b/packages/w3up-client/README.md
@@ -9,15 +9,16 @@
 
 ## About
 
+
 `@web3-storage/w3up-client` is a JavaScript library that provides a convenient interface to the w3up platform, a simple "on-ramp" to the content-addressed decentralized IPFS network.
 
 This library is the user-facing "porcelain" client for interacting with w3up services from JavaScript. It wraps the lower-level [`@web3-storage/access`][access-client-github] and [`@web3-storage/upload-client`][upload-client-github] client packages, which target individual w3up services. We recommend using `w3up-client` instead of using those "plumbing" packages directly, but you may find them useful if you need more context on w3up's architecture and internals.
 
 **`w3up-client` requires Node 18 or higher**.
 
-> ⚠️❗ **Public Data** 🌎: All data uploaded to w3up is available to anyone who requests it using the correct CID. Do not store any private or sensitive information in an unencrypted form using w3up.
+> ⚠️❗ __Public Data__ 🌎: All data uploaded to w3up is available to anyone who requests it using the correct CID. Do not store any private or sensitive information in an unencrypted form using w3up.
 
-> ⚠️❗ **Permanent Data** ♾️: Removing files from w3up will remove them from the file listing for your account, but that doesn’t prevent nodes on the decentralized storage network from retaining copies of the data indefinitely. Do not use w3up for data that may need to be permanently deleted in the future.
+> ⚠️❗ __Permanent Data__ ♾️: Removing files from w3up will remove them from the file listing for your account, but that doesn’t prevent nodes on the decentralized storage network from retaining copies of the data indefinitely. Do not use w3up for data that may need to be permanently deleted in the future.
 
 - [Install](#install)
 - [Usage](#usage)
@@ -91,11 +92,11 @@ flowchart TD
 
 All uses of `w3up-client` to upload with web3.storage follow the flow above. This section shows the most basic way to use the client to start storing data. For more complex integration options, check out the [integration options][https://github.com/web3-storage/w3up/blob/main/packages/w3up-client/README.md#integration-options] docs. For reference, check out the [API reference docs][docs] or the source code of the [`w3cli` package][w3cli-github], which uses `w3up-client` throughout.
 
-> By you or your users registering a w3up Space via email confirmation with [web3.storage](http://web3.storage), you agree to the [Terms of Service](https://web3.storage/docs/terms/).
+> By you or your users registering a w3up Space via email confirmation with [web3.storage](http://web3.storage), you agree to the [Terms of Service](https://web3.storage/docs/terms/). 
 
 #### Creating a client object
 
-The package provides a [static `create` function][docs-create] that returns a [`Client` object][docs-Client].
+The package provides a [static `create` function][docs-create] that returns a [`Client` object][docs-Client]. 
 
 ```js
 import { create } from '@web3-storage/w3up-client'
@@ -185,7 +186,7 @@ import { CarReader } from '@ipld/car'
 import * as Client from '@web3-storage/w3up-client'
 import { StoreMemory } from '@web3-storage/w3up-client/stores/memory'
 
-async function main() {
+async function main () {
   // from "bring your own Agent" example in `Creating a client object" section`
   // used command line to generate KEY and PROOF (stored in env variables)
   // KEY: `npx ucan-key ed --json` in command line, which returns private key and DID for Agent (the private key is stored in KEY)
@@ -193,17 +194,17 @@ async function main() {
   const principal = Signer.parse(process.env.KEY)
   const store = new StoreMemory()
   const client = await Client.create({ principal, store })
-
+  
   // now give Agent the delegation from the Space
   const proof = await parseProof(process.env.PROOF)
   const space = await client.addSpace(proof)
   await client.setCurrentSpace(space.did())
-
+  
   // READY to go!
 }
 
 /** @param {string} data Base64 encoded CAR file */
-async function parseProof(data) {
+async function parseProof (data) {
   const blocks = []
   const reader = await CarReader.fromBytes(Buffer.from(data, 'base64'))
   for await (const block of reader.blocks()) {
@@ -221,7 +222,7 @@ Call [`uploadFile`][docs-Client#uploadFile] to upload a single file, or [`upload
 
 `uploadFile` expects a "Blob like" input, which can be a [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob) or [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) when running in a browser. On node.js, see the [`filesFromPath` library](https://github.com/web3-storage/files-from-path), which can load compatible objects from the local filesystem.
 
-`uploadDirectory` requires `File`-like objects instead of `Blob`s, as the file's `name` property is used to build the directory hierarchy.
+`uploadDirectory` requires `File`-like objects instead of `Blob`s, as the file's `name` property is used to build the directory hierarchy. 
 
 You can control the directory layout and create nested directory structures by using `/` delimited paths in your filenames:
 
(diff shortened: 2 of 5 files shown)


packages/w3up-client/src/client.js
```diff
@@ -233,12 +235,39 @@ export class Client extends Base {
...
+   * 
+   * @typedef {object} CreateOptions
+   * @property {string|false} [account]
+   * 
    * @param {string} name
+   * @param {CreateOptions} options
    */
-  async createSpace(name) {
-    return await this._agent.createSpace(name)
+  async createSpace(name, options = {}) {
```
hannahhoward, line 246: at this API level, I think this should take an actual account object, rather than an email address. 

the code to obtain the account from the email can be left as an exercise above this code level.

also, the default flow for setting up the client often contains and instruction to login:

```
const myAccount = await client.login('zaphod@beeblebrox.galaxy')
```

So often the JS user already has the account.

hannahhoward (changes requested):
This approach is fine, except I have one nit around using an account object rather than an email.

Also we should submit a PR to github.com/storacha-network/docs to update the docs for the JS client.

packages/w3up-client/test/client.test.js
```diff
@@ -259,7 +259,33 @@ export const testClient = {
...
+      assert.ok(space)
+
+      const proof = alice.agent.proofs()
+      .find(p => p.issuer.did() === alice.did())
+      if (!proof) {
+        throw new Error('Recovery Proof not found')
+      }
+      
+      assert.ok(proof)
+      assert.equal(proof.audience.did(), alice.did())
```
fforbeck, line 282: Thanks for the feedback @hannahhoward. I've applied the suggested changes. What is the best way to check if that account is a recovery account?

hannahhoward (approved):
Yea, this is obviously not ideal to have all this logic in creating a space, on the otherhand, with the chain of dependencies there's no way to seperate it. And it will make for a much shorter setup tutorial, which will be quite nice. We should definitely update docs for this.

hannahhoward (approved):
LGTM

alanshaw:
@fforbeck how does this affect the space creation steps in the [CLI](https://github.com/storacha/w3cli)?

packages/w3up-client/src/client.js
```diff
@@ -233,12 +235,50 @@ export class Client extends Base {
...
+      // Delegate space access to the recovery
+      const result = await this.capability.access.delegate({
+        space: space.did(),
+        delegations: [recovery],
+      })
+
+      if (result.error) {
+        throw new Error(
+          `⚠️ Failed to authorize recovery account: ${result.error.name}:${result.error.message}`
+        )
```
alanshaw, line 276: I would just use the message in the error message, and add `result.error` as the cause for context. The `name` is largely for use by code and I don't think it adds anything to this message.

```suggestion
        throw new Error(`failed to authorize recovery account: ${result.error.message}`, { cause: result.error })
```

packages/w3up-client/src/client.js
```diff
@@ -233,12 +235,50 @@ export class Client extends Base {
...
+    const space = await this._agent.createSpace(name)
+
+    const account = options.account
+    if (account) {
+      // Provision the account with the space
+      const provisionResult = await account.provision(space.did())
+      if (provisionResult.error) {
+        throw new Error(
+          `⚠️ Failed to provision account: ${provisionResult.error.name}:${provisionResult.error.message}`
+        )
```
alanshaw, line 258: I recommend against using emojis in error messages - let a UI add that.

```suggestion
        throw new Error(`failed to provision account: ${provisionResult.error.message}`, { cause: provisionResult.error })
```

packages/w3up-client/src/client.js
```diff
@@ -233,12 +235,50 @@ export class Client extends Base {
...
+      }
+
+      // Save the space to authorize the client to use the space
+      await space.save()
```
alanshaw, line 262: Not sure about this - if I don't pass an `account` the space is not saved to my store...should it always be saved?

I think it's ok to only save if an account is passed but it needs to be mentioned in the jsdoc.

If you switch to always saving the space perhaps add a `noSave` option to retain the ability for ephemeral spaces to be created?

alanshaw (changes requested):
This is a great but please could you try to keep non essential changes to a minimum? There's a lot of code formatting change in here that makes it harder for me to review. Just submit a separate PR in the future for code formatting changes 🙏.

packages/w3up-client/src/client.js
fforbeck: I believe it makes sense to save only if you provide the `account`, mainly because if you provide the account we need to get the authorizations done via `.save` call to be able to delegate to the recovery. Then if you want to create an ephemeral space you don't need the account because there is no reason to create a recovery. I will update the `jsdoc` to take that into account 😃

fforbeck:
> This is a great but please could you try to keep non essential changes to a minimum? There's a lot of code formatting change in here that makes it harder for me to review. Just submit a separate PR in the future for code formatting changes 🙏.

Yeah, my bad. I will revert these formatting changes to simplify this PR.

fforbeck:
> @fforbeck how does this affect the space creation steps in the [CLI](https://github.com/storacha/w3cli)?

The CLI should remain unaffected by this change because the `account` parameter is optional. The existing logic in the CLI already supports space creation without requiring an `account`, and it includes the necessary recovery logic (e.g., [this line in the code](https://github.com/storacha/w3cli/blob/main/space.js#L74)).

In the future, we could create a separate PR to simplify the CLI code by leveraging the new version of the `createSpace` function. However, given the current implementation, there should be no impact on the CLI’s functionality.

.github/workflows/w3up-client.yml
```diff
@@ -31,8 +31,6 @@ jobs:
         uses: actions/checkout@v3
       - name: Install
         uses: pnpm/action-setup@v4
-        with:
-          version: 9
```
fforbeck, line 35: I've removed the version because it was causing conflicts with the version defined in the root package.json. It is optional when you have it in package.json: https://github.com/pnpm/action-setup?tab=readme-ov-file#version

packages/w3up-client/README.md
```diff
@@ -126,55 +126,37 @@ A [`Space`][docs-Space] acts as a namespace for your uploads, and what your Agen
...
 
 ```js
 // wait for payment plan to be selected
 while (true) {
   const res = await account.plan.get()
   if (res.ok) break
   console.log('Waiting for payment plan to be selected...')
-  await new Promise(resolve => setTimeout(resolve, 1000))
+  await new Promise((resolve) => setTimeout(resolve, 1000))
 }
```
alanshaw, line 138: It would be rad if this was a method of the client? Would you be up for adding something like that quickly?

alanshaw (approved):
LGTM

packages/w3up-client/README.md
fforbeck: Good idea. I will add that.

packages/w3up-client/src/client.js
```diff
@@ -98,13 +99,45 @@ export class Client extends Base {
...
+      }
+
+      if (Date.now() - startTime > timeout) {
+        throw new Error('Timeout: Payment plan selection took too long.')
+      }
+
+      console.log('Waiting for payment plan to be selected...')
+      await new Promise((resolve) => setTimeout(resolve, interval))
+    }
+  }
```
fforbeck, line 139: what about this @alanshaw ?

packages/w3up-client/README.md
```diff
@@ -129,13 +129,8 @@ const account = await client.login('zaphod@beeblebrox.galaxy')
...
 ```js
-// wait for payment plan to be selected
-while (true) {
-  const res = await account.plan.get()
-  if (res.ok) break
-  console.log('Waiting for payment plan to be selected...')
-  await new Promise((resolve) => setTimeout(resolve, 1000))
-}
+// Wait for a payment plan with a 1-second polling interval and 60-second timeout
+await client.waitForPaymentPlan(account, 1000, 60000)
```
alanshaw, line 133: 🙏 can we land as a separate PR? Apologies - I should have asked...

I think it might be better for this to be an instance method of `AccountPlan` rather than the top level? WDYT?

```js
await account.plan.wait()
```

In the docs I think it's simplier to just use the defaults (i.e. don't specify the interval/timeout).

packages/w3up-client/src/client.js
```diff
@@ -107,6 +107,37 @@ export class Client extends Base {
...
 
+  /**
+   * Waits for a payment plan to be selected for the given account.
+   * This method continuously checks the account's payment plan status
+   * at a specified interval until a valid plan is selected or the timeout is reached.
+   *
+   * @param {Account.Account} account - The account object to check the payment plan for.
+   * @param {number} [interval=1000] - The polling interval in milliseconds (default is 1000ms).
+   * @param {number} [timeout=180000] - The maximum time to wait in milliseconds before throwing a timeout error (default is 3 minutes).
+   * @returns {Promise<void>} - Resolves once a payment plan is selected within the timeout.
```
alanshaw, line 118: I would return the plan information when it is available.

packages/w3up-client/src/client.js
```diff
@@ -107,6 +107,37 @@ export class Client extends Base {
     return Account.list(this)
   }
 
+  /**
+   * Waits for a payment plan to be selected for the given account.
+   * This method continuously checks the account's payment plan status
+   * at a specified interval until a valid plan is selected or the timeout is reached.
+   *
+   * @param {Account.Account} account - The account object to check the payment plan for.
+   * @param {number} [interval=1000] - The polling interval in milliseconds (default is 1000ms).
```
alanshaw, line 116: I'd pass an options object with these two properties.

packages/w3up-client/src/client.js
```diff
@@ -107,6 +107,37 @@ export class Client extends Base {
...
+   * @param {number} [timeout=180000] - The maximum time to wait in milliseconds before throwing a timeout error (default is 3 minutes).
```
alanshaw, line 117: Might be better to pass an `AbortSignal` instead so that if folks are using this in a UI they can have a cancel button. Another reason is that it could take longer than 3 minutes, considering you have to log into your email, click the link, add credit card details etc.

packages/w3up-client/test/client.test.js
```diff
@@ -698,6 +698,48 @@ export const testClient = {
...
+          plan: {
+            get: async () => ({ ok: false }),
+          },
+        }
+        // @ts-expect-error
+        await assert.rejects(alice.waitForPaymentPlan(account, 100, 500), {
+          message: 'Timeout: Payment plan selection took too long.',
+        })
+      },
+  },
```
alanshaw, line 742: ❤️

alanshaw (changes requested):

alanshaw (approved):

PR #1547, opened 2024-09-16
feat: wait for plan selection

fforbeck:
# Add `wait()` Method to `AccountPlan` Class

## Summary

This PR introduces a new `wait()` method to the `AccountPlan` class, which simplifies the process of waiting for a payment plan to be selected. By integrating configurable polling intervals, timeouts, and abort signals, the method abstracts the common logic required in this scenario, making it easier for developers to implement and customize their workflows.

## Changes

- **New `wait()` Method**:
  - The `wait()` method repeatedly checks the account's payment plan status at a given interval until:
    - A valid plan is selected.
    - The timeout is reached.
    - The operation is aborted using an `AbortSignal`.
  - Available options:
    - `interval` (default: 1000ms): Sets the polling interval.
    - `timeout` (default: 15 minutes): Defines the maximum wait time.
    - `signal`: An optional `AbortSignal` to allow for cancellation.
  - This method improves the developer experience by abstracting the repetitive logic required to wait for a payment plan and offering flexibility through configurable options.

## How to Test

1. Run the test suite with `npm run build && npm run test`.
2. Ensure all existing tests pass.
3. Verify that the new test cases for the `wait()` method in `account.test.js` are passing, including tests for:
   - Successful payment plan selection.
   - Timeout errors.
   - Aborted operations.

### DIFF
--- a/packages/w3up-client/src/account.js
+++ b/packages/w3up-client/src/account.js
@@ -236,6 +236,46 @@ export class AccountPlan {
   }
 
   /**
+   * Waits for a payment plan to be selected.
+   * This method continuously checks the account's payment plan status
+   * at a specified interval until a valid plan is selected, or when the timeout is reached,
+   * or when the abort signal is aborted.
+   *
+   * @param {object} [options]
+   * @param {number} [options.interval=1000] - The polling interval in milliseconds (default is 1000ms).
+   * @param {number} [options.timeout=300000] - The maximum time to wait in milliseconds before throwing a timeout error (default is 5 minutes).
+   * @param {AbortSignal} [options.signal] - An optional AbortSignal to cancel the waiting process.
+   * @returns {Promise<import('@web3-storage/access').PlanGetSuccess>} - Resolves once a payment plan is selected within the timeout.
+   * @throws {Error} - Throws an error if there is an issue retrieving the payment plan or if the timeout is exceeded.
+   */
+  async wait(options) {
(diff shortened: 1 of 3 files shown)


packages/w3up-client/src/account.js
```diff
@@ -235,6 +235,46 @@ export class AccountPlan {
...
 
+  /**
+   * Waits for a payment plan to be selected.
+   * This method continuously checks the account's payment plan status
+   * at a specified interval until a valid plan is selected, or when the timeout is reached,
+   * or when the abort signal is aborted.
+   *
+   * @param {object} [options]
+   * @param {number} [options.interval=1000] - The polling interval in milliseconds (default is 1000ms).
+   * @param {number} [options.timeout=300000] - The maximum time to wait in milliseconds before throwing a timeout error (default is 5 minutes).
```
alanshaw, line 246: I think this is a bit short. Login will timeout after 15 minutes (i.e. it's not possible to login and select a plan after 15 minutes) so it might be an idea to align with that.

https://github.com/storacha/w3up/blob/ea02adb77cfe207a5823e24d886d28b6650ecc03/packages/upload-api/src/access/authorize.js#L41-L42

alanshaw (approved):
🚀 This is awesome.

I think this could be merged as is but personally I think we should increase the timeout to 15 minutes as mentioned in comment.

PR #1548, opened 2024-09-16
fix: remove redundant package manager version

alanshaw:
Apparently this is not needed when `packageManager` is set in root `package.json`.


fforbeck (approved):
LGTM!

PR #1557, opened 2024-10-04
refactor: De-deprecate `AgentData`'s `addSpace()`

Peeja:
This was deprecated in https://github.com/storacha/w3up/pull/433, but it's still what we use, and no clear migration path was given.

@alanshaw Do you remember what the plan was? Should we do this and reverse it, or is there a better way we should write down?

### DIFF
--- a/packages/access-client/src/agent-data.js
+++ b/packages/access-client/src/agent-data.js
@@ -112,7 +112,6 @@ export class AgentData {
   }
 
   /**
-   * @deprecated
    * @param {import('@ucanto/interface').DID} did
    * @param {import('./types.js').SpaceMeta} meta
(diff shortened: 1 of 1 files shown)


alanshaw (approved):
Yeah, not fully remembered (sorry) but I _think_ the idea was to signal early that we wanted to move away from the client storing state like the list of spaces, or the "current" space, beyond the delegations it has received/created. i.e. the list of spaces should be derivable from the delegations in the store.

My apologies - I _should_ have documented and probably done in a separate PR...but hey, younger me was a maverick.

I'm happy to merge this but I still desire the goal above and so `addSpace` should not really exist, and `addDelegation` should be the only method here for adding spaces.

PR #1560, opened 2024-10-08
fix(filecoin-api): parallel put to piece accept queue

alanshaw:
Simply uses `p-map` to put items to the queue in parallel and speed up execution time of this lambda (which is currently timing out).


travis (commented):
lgtm

PR #1562, opened 2024-10-11
feat: usage/record capability definition

fforbeck:
### Add `usage/record` Capability definition for Egress Traffic Tracking

**Summary:**
- Introduced a new capability `usage/record` definition to track egress traffic metrics.
- The `w3upInfra` upload-api will be updated to implement this new capability by leveraging Stripe's Usage Records API for accurate tracking and billing.

RFC: https://github.com/storacha/RFC/pull/37


packages/capabilities/src/types.ts
```diff
@@ -161,6 +165,17 @@ export interface UsageData {
   }>
 }
 
+export interface EgressData {
+  /** Id of the customer that is being billed. */
+  customer: AccountDID
+  /** CID of the resource that was served. */
+  resourceCID: string
```
alanshaw, line 172:
```suggestion
  resource: Link
```

packages/capabilities/src/usage.js
```diff
@@ -40,3 +40,22 @@ export const report = capability({
...
+ * Capability can be invoked by an agent to record usage data for a given resource.
+ */
+export const record = capability({
+  can: 'usage/record',
+  with: SpaceDID,
+  nb: Schema.struct({
+    /** MailTo DID of the customer that is being billed. */
+    customer: AccountDID,
+    /** CID of the resource that was served. */
+    resourceCID: Schema.string(),
```
alanshaw, line 54:
```suggestion
    resource: Schema.link(),
```

packages/upload-api/test/storage/usage-storage.js
```diff
@@ -64,4 +68,29 @@ export class UsageStorage {
...
+
+  /**
+   * Simulate a record of egress data for a customer.
+   *
+   * @param {import('../types.js').AccountDID} customer
+   * @param {string} resourceCID
+   * @param {number} bytes
+   * @param {Date} servedAt
+   */
+  async record(customer, resourceCID, bytes, servedAt) {
```
alanshaw, line 80: Please lets not specify types in variable names.

packages/w3up-client/src/capability/usage.js
```diff
@@ -31,6 +31,32 @@ export class UsageClient extends Base {
...
+   * @param {API.EgressData} egressData
+   * @param {object} [options]
+   * @param {string} [options.nonce]
+   */
+  async record(space, egressData, options) {
+    const out = await record(
+      { agent: this.agent },
+      { ...options, space, egressData }
+    )
+    /* c8 ignore next 7 */
```
alanshaw, line 51: 7 includes the return value

alanshaw (changes requested):

packages/upload-api/src/usage/record.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * @param {API.Input<Usage.record>} input
+ * @param {API.UsageServiceContext} context
+ * @returns {Promise<API.Result<API.EgressRecordSuccess, API.EgressRecordFailure>>}
+ */
+const record = async ({ capability }, context) => {
+  const res = await context.usageStorage.record(
+    capability.nb.customer,
+    capability.nb.resourceCID,
```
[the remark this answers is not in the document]
alanshaw: Use `UnkonwnLink` for links you don't know the CID version of.

packages/capabilities/src/usage.js
```diff
@@ -40,3 +40,22 @@ export const report = capability({
...
+
+/**
+ * Capability can be invoked by an agent to record usage data for a given resource.
+ */
+export const record = capability({
+  can: 'usage/record',
+  with: SpaceDID,
+  nb: Schema.struct({
+    /** MailTo DID of the customer that is being billed. */
+    customer: AccountDID,
```
travis, line 52: question: will this be available in Freeway where this invocation will be created? If I remember correctly I think we may only have information about the space that was authorized, and the handler for this will need to determine who to bill based on that

packages/upload-api/src/usage/record.js
```diff
@@ -0,0 +1,24 @@
...
+export const provide = (context) =>
+  Provider.provide(Usage.record, (input) => record(input, context))
+
+/**
+ * @param {API.Input<Usage.record>} input
+ * @param {API.UsageServiceContext} context
+ * @returns {Promise<API.Result<API.EgressRecordSuccess, API.EgressRecordFailure>>}
+ */
+const record = async ({ capability }, context) => {
+  const res = await context.usageStorage.record(
```
travis, line 15: thought: this may be where we take the space DID and look up the customer currently registered as the "provider" - I'm not sure if we have APIs to do this yet so that may need to be added to another interface? 

It's also possible we want to do this resolution inside the `UsageStorage` implementation, in which case we should just pass the space DID through - I'm not sure which makes the most sense but happy to kick it around in higher bandwidth if that's helpful!

packages/w3up-client/src/capability/usage.js
```diff
@@ -31,6 +31,32 @@ export class UsageClient extends Base {
...
+   * @param {API.EgressData} egressData
+   * @param {object} [options]
+   * @param {string} [options.nonce]
+   */
+  async record(space, egressData, options) {
+    const out = await record(
+      { agent: this.agent },
+      { ...options, space, egressData }
+    )
+    /* c8 ignore next 5 */
```
travis, line 51: question: is it possible to add a test for the error case so we can avoid the c8 ignore here?

travis (dismissed):
One question about the customer account DID that needs to be resolved - I think this PR probably needs to do the space DID -> account DID resolution OR needs to assume the underlying `UsageStore` implementation does that

packages/capabilities/src/usage.js
fforbeck: Yeah, I've discussed this with @Peeja and it will be available in Freeway at the time the invocation is triggered, here is the convo: https://github.com/storacha/RFC/pull/36#discussion_r1790818236

packages/upload-api/src/usage/record.js
fforbeck: We will have the customer ID provided during the invocation in Freeway. We could do it here in another moment, but I think that for the first iteration, it should be easy enough to get that in Freeway considering the changes that @Peeja will be working on.

packages/w3up-client/src/capability/usage.js
fforbeck: Sounds good. I will add that.

packages/capabilities/src/usage.js
travis: I think that conversation is missing the step that I'm talking about - @Peeja is right that we'll get the space DID from the content commitment, but there is an additional database query needed to go from space DID to account DID that cannot happen in Freeway (since it doesn't have access to Dynamo, and would take too long anyway)

packages/upload-api/src/usage/record.js
travis: I think there's a misunderstanding here - the content commitment will have the space DID, but as far as I know it will not have the account DID - maybe worth a quick checkin to confirm in standup today before digging into this too much deeper, as I could be wrong, but I don't think we include accountDIDs in content commitments (particularly since they contain email addresses and are therefore very sensitive from a GDPR perspective)

packages/capabilities/src/usage.js
travis: yep! and the space DID should already be available in the `with` so you don't need to add that. thanks!

packages/w3up-client/src/capability/usage.js
fforbeck: @travis I tried to mock the `agent.invokeAndExecute` to return an error using Sinon, but for some reason, it doesn't work. I guess that's why the other capability also had the `c8 ignore` for this part.

travis: ok - I'm ok skipping this in the interest of shipping if you think it's going to be a rabbithole!

packages/capabilities/src/usage.js
fforbeck: @travis The Provider DID returned by that function is `did:web:string`, but the customer DID is type `AccountID` = `did:mailto:...`. Any suggestions on how to convert it to a mailto DID?

travis: ack, sorry, I think I mislead you - apologies! I think the function I linked must be returning the list of "providers" (ie, services) a space is registered to, not the list of "customers" of those providers, which is what you want - let me know if you're unable to find that function, but from Discord I think you're already on it

packages/upload-api/src/usage/record.js
```diff
@@ -0,0 +1,40 @@
...
+        message: `Failed to get consumer`,
+      },
+    }
+  }
+  const consumer = consumerResponse.ok
+  const res = await context.usageStorage.record(
+    consumer.customer,
+    capability.nb.resource,
+    capability.nb.bytes,
+    new Date(capability.nb.servedAt * 1000)
```
fforbeck, line 35: Updated the function to find the `customer` using the provider `invocation.audience` and consumer (`capability.with`), the `customer` represent the account we want to charge for the egress traffic.

packages/capabilities/src/types.ts
```diff
@@ -193,6 +208,7 @@ export interface ConsumerGetSuccess {
   allocated: number
   limit: number
   subscription: string
+  customer: AccountDID
```
fforbeck, line 211: `Customer` is now part of the `ConsumerGetSuccess` response.

packages/upload-api/src/usage/record.js
alanshaw: Correct, no account in location commitment, just space.

packages/upload-api/src/usage/record.js
```diff
@@ -0,0 +1,40 @@
...
+  )
+  const consumerResponse = await context.provisionsStorage.getConsumer(
+    provider,
+    capability.with
+  )
+  if (consumerResponse.error) {
+    return {
+      error: {
+        name: 'EgressRecordFailure',
+        message: `Failed to get consumer`,
```
alanshaw, line 26:
```suggestion
        message: 'Failed to get consumer',
```

packages/upload-api/src/usage/record.js
```diff
@@ -0,0 +1,40 @@
...
+const record = async ({ capability, invocation }, context) => {
+  const provider = /** @type {`did:web:${string}`} */ (
+    invocation.audience.did()
+  )
+  const consumerResponse = await context.provisionsStorage.getConsumer(
+    provider,
+    capability.with
+  )
+  if (consumerResponse.error) {
+    return {
```
alanshaw, line 23: This actually returns `ConsumerNotFound` error when there's no subscription setup. It's arguably more useful than "Failed to get consumer".

```suggestion
    return consumerResponse
```

Really we need to be better at defining error types, as the failure type for `getConsumer` is currently the catchall `Failure`.

If you're up for it, it would be good to add a type for `ConsumerNotFound` and include it as a possible value for `EgressRecordFailure`.

i.e.
```ts
type EgressRecordFailure = ConsumerNotFound | Ucanto.Failure
```

packages/upload-api/src/types/usage.ts
```diff
@@ -13,4 +15,10 @@ export interface UsageStorage {
     space: SpaceDID,
     period: { from: Date; to: Date }
   ) => Promise<Result<UsageData, Failure>>
+  record: (
+    customer: AccountDID,
+    resource: UnknownLink,
+    bytes: number,
+    servedAt: Date
```
alanshaw, line 22: We should add a `cause: Link` field to identify the invocation.

packages/upload-api/src/types/usage.ts
```diff
@@ -13,4 +15,10 @@ export interface UsageStorage {
     space: SpaceDID,
     period: { from: Date; to: Date }
   ) => Promise<Result<UsageData, Failure>>
+  record: (
+    customer: AccountDID,
+    resource: UnknownLink,
```
alanshaw, line 20: Would be good to get a comment here - what is the resource? It's the root CID of a DAG (not the CID of a CAR) or I imagine it could also be literally a CID of any sub-DAG within a DAG.

alanshaw, line 20: Do we have enough information to identify the blob(s)/shard(s) the resource was served from? Feels like a useful thing to track here, especially if the CID being used is not a root CID, but a sub-DAG. I'm thinking about users querying why a certain CID has used so much egress but it is not one of the root CIDs from `upload/add` - how do we prove it belongs to them?

packages/upload-api/src/usage/record.js
```diff
@@ -0,0 +1,40 @@
...
+  const consumer = consumerResponse.ok
+  const res = await context.usageStorage.record(
+    consumer.customer,
+    capability.nb.resource,
+    capability.nb.bytes,
+    new Date(capability.nb.servedAt * 1000)
+  )
+  if (res.error) return res
+
+  return res
```
alanshaw, line 39:
```suggestion
  return res
```

packages/w3up-client/src/capability/usage.js
```diff
@@ -61,3 +87,32 @@ export const report = async (
...
+export const record = async (
+  { agent },
+  { space, egressData, nonce, proofs = [] }
+) => {
+  const receipt = await agent.invokeAndExecute(UsageCapabilities.record, {
+    with: space,
+    proofs,
+    nonce,
+    nb: {
+      customer: egressData.customer,
```
alanshaw, line 111:
```suggestion
```

packages/capabilities/src/types.ts
```diff
@@ -161,6 +165,17 @@ export interface UsageData {
   }>
 }
 
+export interface EgressData {
+  /** Id of the customer that is being billed. */
+  customer: AccountDID
```
alanshaw, line 170: The service needs to obtain this.

```suggestion
```

packages/capabilities/src/usage.js
alanshaw, line 52:
```suggestion
```

packages/upload-api/src/types/usage.ts
```diff
@@ -13,4 +15,10 @@ export interface UsageStorage {
     space: SpaceDID,
     period: { from: Date; to: Date }
   ) => Promise<Result<UsageData, Failure>>
+  record: (
+    customer: AccountDID,
```
alanshaw, line 19: Record space DID also? Would be good to be able to enumerate usage by space.

packages/w3up-client/test/capability/usage.test.js
```diff
@@ -68,6 +69,47 @@ export const UsageClient = Test.withContext({
...
+
+      const car = await randomCAR(128)
+      const resource = car.cid
+      await alice.capability.upload.add(car.roots[0], [resource])
+
+      const result = await alice.capability.upload.get(car.roots[0])
+      assert.ok(result)
+
+      const record = await alice.capability.usage.record(space.did(), {
+        customer: 'did:mailto:alice@web.mail',
```
alanshaw, line 105:
```suggestion
```

alanshaw (changes requested):

packages/upload-api/src/usage/record.js
fforbeck: That makes sense to me. I will update the types. Thanks

packages/upload-api/src/types/usage.ts
fforbeck: It is the same data CID we use in the gateway to fetch the content. It is pulled from the URL via `withParsedIpfsUrl` handler from the `gateway-lib`: https://github.com/storacha/gateway-lib/blob/main/src/middleware.js#L118
I think this is the Root CID of a DAG, but I am not 100% sure. Do you know?

packages/upload-api/src/types/usage.ts
fforbeck: @alanshaw Where does it come from?

packages/upload-api/src/types/usage.ts
alanshaw: https://w3s.link/ipfs/bafybeibrqc2se2p3k4kfdwg7deigdggamlumemkiggrnqw3edrjosqhvnm/yesthisisdog.jpg

In the above URL, the root CID is `bafybeibrqc2se2p3k4kfdwg7deigdggamlumemkiggrnqw3edrjosqhvnm`.

If you look at the HTTP headers, the etag will tell you the CID of the resource at the end of the path i.e. the CID of `yesthisisdog.jpg`, which is `bafkreiem4twkqzsq2aj4shbycd4yvoj2cx72vezicletlhi7dijjciqpui`.

You can use this as a root CID too!
https://w3s.link/ipfs/bafkreiem4twkqzsq2aj4shbycd4yvoj2cx72vezicletlhi7dijjciqpui

If you look at the HTTP headers, the etag is `bafkreiem4twkqzsq2aj4shbycd4yvoj2cx72vezicletlhi7dijjciqpui`, since there is no path.

So when I say "root" CID I mean the CID that might be considered the root CID of a given upload. i.e. a CID that would be set as `root` in an `upload/add` invocation.

All I think is needed is a clarification that the `resource` as you have it there is _not_ the CID of a blob/shard, it's the CID of some gateway accessible content.

packages/upload-api/src/types/usage.ts
alanshaw: In a ucanto handler it is `invocation.cid`

fforbeck:
@alanshaw I've implemented the suggested changes and added an additional note to the `resource` field explaining that it is not a CID of a blob/shard. It is ready for another review when you get a chance.

fforbeck:
@travis I've implemented the suggested changes. It is ready for another review.

packages/w3up-client/src/capability/usage.js
```diff
@@ -61,3 +92,35 @@ export const report = async (
...
+
+/**
+ * Record egress data for a resource from a given space.
+ *
+ * @param {{agent: API.Agent}} client
+ * @param {object} egressData
+ * @param {API.SpaceDID} egressData.space
+ * @param {API.UnknownLink} egressData.resource
+ * @param {number} egressData.bytes
+ * @param {string} egressData.servedAt
```
alanshaw, line 104: You could make this optional and default to now.

alanshaw (approved):

PR #1571, opened 2024-10-31
fix: publish location claim to content claims service

alanshaw:
Big oversight here, although I think the original intention was to have the client publish these so perhaps not.

Anyways, since no location claims are being published the gateway cannot fetch an index and serve bytes quickly. It has to get a location claim for every single block using the old dynamodb. Being able to use the index will result in a speed increase for serving content via the gateway.

Depends on https://github.com/storacha/content-claims/pull/74 for passing tests.


Peeja (approved):
You're not making that claim?

Good catch! ❤️ LGTM!

fforbeck (approved):

PR #1572, opened 2024-10-31
fix(egress/record): rename capability

fforbeck:
### Overview

This PR restructures the `usage/record` capability, moving it under the `Space` namespace instead of `Usage`. As part of this change, the `usage/record` definition has been renamed to `space/content/serve/egress/record`, and a new top-level capability, `space/content/serve/*`, has been introduced.

### Key Changes

- **Namespace Update**: The `usage/record` capability now resides under the `Space` namespace.
- **New Naming Convention**:
  - `space/content/serve/egress/record`: This capability records egress for all served data.
  - `space/content/serve/*`: New top-level capability, representing general serve actions within the `Space.contentServe` namespace.


packages/capabilities/test/capabilities/usage.test.js
```diff
@@ -200,4 +203,47 @@ describe('usage capabilities', function () {
...
+      proofs: [usageRecordDelegationProof],
+    })
+
+    // W3 validates the delegation from Gateway to itself
+    const result = await access(await recordInvocation.delegate(), {
+      capability: Usage.record,
+      principal: Verifier,
+      authority: w3,
+      validateAuthorization,
+      resolveDIDKey: () => Schema.ok(gateway.toDIDKey()),
```
travis, line 238: question: is this going to resolve all `did:web`s to the gateway key? I'm concerned this might hide a test failure by thinking that the `w3` DID uses the same private key and is therefore the same entity as the `gateway` principal - maybe misunderstanding but curious what you think... it does look like `resolveDIDKey` is passed a `DID` - maybe it can return different DIDKeys for different DIDs with a lookup table here?

travis, line 238: resolveDIDKey: (k) => {
        const didKey = [w3, gateway].find(signer => signer.did() === k)?.toDIDKey()
        if (didKey) {
          return Schema.ok(didKey)
        } else {
          return { error: new DIDResolutionError(k) }
        }
      },

travis: ok so this is what I'm imagining:

```suggestion
      resolveDIDKey: (k) => {
        const didKey = [w3, gateway].find(signer => signer.did() === k)?.toDIDKey()
        if (didKey) {
          return Schema.ok(didKey)
        } else {
          return { error: new DIDResolutionError(k) }
        }
      },
```

packages/upload-api/test/helpers/utils.js
```diff
@@ -37,11 +37,15 @@ export const mallory = ed25519.parse(
...
-  .withDID('did:web:test.web3.storage')
+export const w3Signer = ed25519.parse(
+  'MgCYKXoHVy7Vk4/QjcEGi+MCqjntUiasxXJ8uJKY0qh11e+0Bs8WsdqGK7xothgrDzzWD0ME7ynPjz2okXDh8537lId8='
+)
+export const w3 = w3Signer.withDID('did:web:test.web3.storage')
+
+export const freewaySigner = ed25519.parse(
+  'MgCaNpGXCEX0+BxxE4SjSStrxU9Ru/Im+HGNQ/JJx3lDoI+0B3NWjWW3G8OzjbazZjanjM3kgfcZbvpyxv20jHtmcTtg='
+)
+export const freeway = freewaySigner.withDID('did:web:freeway.web3.storage')
```
alanshaw, line 48: Personally I'd use "gateway" here and "did:web:w3s.link".

alanshaw (approved):
As I mentioned before, I don't think this is the way we should do this. There's no delegatation chain back to the user. All we're doing here is using UCAN to verify we can call microservices within our own stack, for which we could simply use a plain HTTP API and a bearer token.

From discord: It's important that the gateway is authorized to record egress. In a future where there are multiple decentralized gateways, there should be no other gateway able to record egress usage for a given space unless authorized to do so. There should be a delegation chain from the user to the gateway that authorizes it to track egress usage so that the user can be billed accordingly.

I'm loathed to approve this - we're going to have to undo it in the future.

packages/upload-api/test/helpers/utils.js
fforbeck: Fixed

packages/capabilities/src/space.js
```diff
@@ -63,3 +63,32 @@ export const allocate = capability({
...
+  can: 'space/content/serve/*',
+  with: SpaceDID,
+  derives: equalWith,
+})
+
+/**
+ * Capability can be invoked by an agent to record egress data for a given resource.
+ * It can be derived from any of the `space/content/serve/*` capability that has matching `with`.
+ */
+export const recordEgress = capability({
```
alanshaw, line 82: ...to echo the ability name?

```suggestion
export const egressRecord = capability({
```

packages/upload-api/test/helpers/utils.js
alanshaw, line 48:
```suggestion
export const gatewaySigner = ed25519.parse(
  'MgCaNpGXCEX0+BxxE4SjSStrxU9Ru/Im+HGNQ/JJx3lDoI+0B3NWjWW3G8OzjbazZjanjM3kgfcZbvpyxv20jHtmcTtg='
)
export const gateway = gatewaySigner.withDID('did:web:w3s.link')
```

alanshaw (approved):
LGTM

PR #1586, opened 2024-11-16
feat: add pre-built browser bundle

alanshaw:
This PR adds a pre-built browser bundle.

Importing the bundle will add a `w3up` variable to the global context which contains all the exports from the client library.


fforbeck (commented):
LGTM!

fforbeck (approved):

PR #1590, opened 2024-11-26
feat!: content serve authorization

fforbeck:
### Context
To enable a gateway to serve content from a specific space, we must ensure that the space owner delegates the `space/content/serve/*` capability to the Gateway. This delegation allows the Gateway to serve content and log egress events appropriately.

I created a new function `authorizeContentServe` for this implementation and included it in the `createSpace` flow. This is a breaking change because now the user is forced to provide the DIDs of the Content Serve services, and the connection, or skip the authorization flow.

Additionally, with the `authorizeContentServe` function, we can implement a feature in the Console App that enables users to explicitly authorize the Freeway Gateway to serve content from existing/legacy spaces.

### Main Changes
- **New Functionality:** Added a new function, `authorizeContentServe`, in the `w3up-client` module to facilitate the delegation process. Integrated it with the `createdSpace` flow.
- **Testing:** Introduced test cases to verify the authorization of specified gateways.
- **Fixes:** Resolved issues with previously broken test cases (Egress Record).

### Related Issues
- https://github.com/storacha/project-tracking/issues/158
- https://github.com/storacha/project-tracking/issues/160

### DIFF
--- a/packages/upload-api/src/access/claim.js
+++ b/packages/upload-api/src/access/claim.js
@@ -13,6 +13,7 @@ export const provide = (ctx) =>
 
 /**
  * Checks if the given Principal is an Account.
+ *
  * @param {API.Principal} principal
  * @returns {principal is API.Principal<API.DID<'mailto'>>}
  */
@@ -20,6 +21,7 @@ const isAccount = (principal) => principal.did().startsWith('did:mailto:')
 
 /**
  * Returns true when the delegation has a `ucan:*` capability.
+ *
  * @param {API.Delegation} delegation
  * @returns boolean
  */
--- a/packages/upload-api/src/lib.js
+++ b/packages/upload-api/src/lib.js
@@ -147,8 +147,8 @@ export const execute = async (agent, input) => {
  * a receipt it will return receipt without running invocation.
  *
  * @template {Record<string, any>} S
- * @param {Types.Invocation} invocation
  * @param {Agent<S>} agent
+ * @param {Types.Invocation} invocation
  */
 export const run = async (agent, invocation) => {
   const cached = await agent.context.agentStore.receipts.get(invocation.link())
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -30,6 +30,7 @@ import { FilecoinClient } from './capability/filecoin.js'
 import { CouponAPI } from './coupon.js'
 export * as Access from './capability/access.js'
 import * as Result from './result.js'
+import { remove } from '@web3-storage/capabilities/store'
 
 export {
   AccessClient,
@@ -252,15 +253,22 @@ export class Client extends Base {
    * If an account is not provided, the space is created without any delegation and is not saved, hence it is a temporary space.
    * When an account is provided in the options argument, then it creates a delegated recovery account
    * by provisioning the space, saving it and then delegating access to the recovery account.
+   * In addition, it authorizes the listed Content Serve Services to serve content from the created space.
+   * It is done by delegating the `space/content/serve/*` capability to the Content Serve Service.
+   * User can skip the Content Serve authorization by setting the `skipContentServeAuthorization` option to `true`.
    *
-   * @typedef {object} CreateOptions
-   * @property {Account.Account} [account]
+   * @typedef {object} SpaceCreateOptions
+   * @property {boolean} [skipContentServeAuthorization] - Whether to skip the Content Serve authorization.
+   * @property {`did:${string}:${string}`[]} [authorizeContentServeServices] - The DID Key or DID Web of the Content Serve Service to authorize to serve content from the created space.
+   * @property {import('./types.js').ConnectionView<import('./types.js').ContentServeService>} [authorizeContentServeServices.connection] - The connection to the Content Serve Service that will handle, validate, and store the access/delegate UCAN invocation.
+   * @property {Account.Account} [account] - The account configured as the recovery account for the space.
+   * @property {string} [name] - The name of the space to create.
    *
-   * @param {string} name
-   * @param {CreateOptions} options
+   * @param {string} name - The name of the space to create.
+   * @param {SpaceCreateOptions} options - Options for the space creation.
    * @returns {Promise<import("./space.js").OwnedSpace>} The created space owned by the agent.
    */
-  async createSpace(name, options = {}) {
+  async createSpace(name, options) {
     const space = await this._agent.createSpace(name)
 
     const account = options.account
@@ -291,6 +299,31 @@ export class Client extends Base {
           `failed to authorize recovery account: ${result.error.message}`,
           { cause: result.error }
         )
+      }
+    }
+
+    // Authorize the listed Content Serve Services to serve content from the created space
+    if (options.skipContentServeAuthorization !== true) {
+      if (
+        !options.authorizeContentServeServices ||
+        options.authorizeContentServeServices.length === 0
+      ) {
+        throw new Error(
+          'failed to authorize Content Serve Services: missing <authorizeContentServeServices> option'
+        )
+      }
+
+      if (!options.connection) {
+        throw new Error(
+          'failed to authorize Content Serve Services: missing <connection> option'
+        )
+      }
+
+      for (const service of options.authorizeContentServeServices) {
+        await this.authorizeContentServe(space, {
+          audience: service,
+          connection: options.connection,
+        })
       }
     }
 
@@ -353,7 +386,7 @@ export class Client extends Base {
         )
       }
 
-      return delegation
+      return { ok: { ...verificationResult.out.ok, delegation } }
     } finally {
       if (currentSpace) {
         await this.setCurrentSpace(currentSpace.did())
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -345,7 +345,7 @@ export class Client extends Base {
   async authorizeContentServe(space, options) {
     const currentSpace = this.currentSpace()
     try {
-      // Set the current space to the space we are authorizing the gateway for
+      // Set the current space to the space we are authorizing the gateway for, otherwise the delegation will fail
       await this.setCurrentSpace(space.did())
 
       /** @type {import('@ucanto/client').Principal<`did:${string}:${string}`>} */
@@ -353,7 +353,7 @@ export class Client extends Base {
         did: () => options.audience,
       }
 
-      // Create the delegation
+      // Grant the audience the ability to serve content from the space, it includes existing proofs automatically
       const delegation = await this.createDelegation(
         audience,
         [SpaceCapabilities.contentServe.can],
@@ -363,12 +363,15 @@ export class Client extends Base {
       )
 
       // Publish the delegation to the content serve service
+      const accessProofs = this.proofs([
+        { can: AccessCapabilities.access.can, with: space.did() },
+      ])
       const verificationResult = await AccessCapabilities.delegate
         .invoke({
           issuer: this._agent.issuer,
           audience,
           with: space.did(),
-          proofs: [delegation],
+          proofs: [...accessProofs, delegation],
           nb: {
             delegations: {
               [delegation.cid.toString()]: delegation.cid,
--- a/packages/w3up-client/src/client.js
+++ b/packages/w3up-client/src/client.js
@@ -9,6 +9,7 @@ import {
   Index as IndexCapabilities,
   Upload as UploadCapabilities,
   Filecoin as FilecoinCapabilities,
+  Space as SpaceCapabilities,
 } from '@web3-storage/capabilities'
 import * as DIDMailto from '@web3-storage/did-mailto'
 import { Base } from './base.js'
@@ -253,6 +254,9 @@ export class Client extends Base {
    *
    * @typedef {object} CreateOptions
    * @property {Account.Account} [account]
+   * @property {boolean} [unauthorizeGateway] - If true, the gateway will not be authorized to serve content from the space.
+   * @property {`did:web:${string}`} [gateway] - The gateway to be authorized to serve content from the space. If not provided, the default did:web:w3s.link gateway will be authorized.
+   * @property {number} [gatewayExpiration] - The time in seconds to expire the gateway authorization.
    *
    * @param {string} name
    * @param {CreateOptions} options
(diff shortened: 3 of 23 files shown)


packages/w3up-client/src/client.js
```diff
@@ -291,9 +292,66 @@ export class Client extends Base {
...
 
+  /**
+   * Authorizes a gateway to serve content from the provided space and record egress events.
+   * Delegates the following capabilities to the gateway:
+   * - `space/content/serve/*`
+   *
+   * @param {import('./types.js').OwnedSpace} space - The space to authorize the gateway for.
+   * @param {object} [options] - Options for the authorization.
+   * @param {`did:web:${string}`} [options.gateway] - The Web DID of the gateway to authorize. If not provided, the default `did:web:w3s.link` gateway will be authorized.
+   * @param {number} [options.expiration] - The time in seconds to expire the authorization.
```
alanshaw, line 307: Not quite, this is the time at which the delegation expires in seconds from unix epoch.

packages/w3up-client/src/client.js
```diff
@@ -291,9 +292,66 @@ export class Client extends Base {
...
+   * @returns {Promise<import('./delegation.js').AgentDelegation<any>>} Resolves with the AgentDelegation instance once the gateway is successfully authorized.
+   */
+  async authorizeGateway(space, options = {}) {
+    const currentSpace = this.currentSpace()
+    try {
+      // Set the current space to the space we are authorizing the gateway for
+      await this.setCurrentSpace(space.did())
+
+      const authProof = await space.createAuthorization(this.agent)
```
alanshaw, line 316: You shouldn't need to do this, you have this proof in your store already. This should only be required when you first create a spece.

packages/w3up-client/src/client.js
```diff
@@ -291,9 +292,66 @@ export class Client extends Base {
...
+      const gateway = {
+        did: () => options.gateway ?? `did:web:w3s.link`,
+      }
+
+      const delegation = await this.createDelegation(
+        gateway,
+        [SpaceCapabilities.contentServe.can],
+        {
+          expiration: options.expiration ?? Infinity,
+          proofs: [authProof],
```
alanshaw, line 328: `createDelegation` _should_ include all the proofs you need.

packages/w3up-client/src/client.js
```diff
@@ -291,9 +292,66 @@ export class Client extends Base {
...
+      if (result.error) {
+        throw new Error(
+          `failed to authorize gateway: ${result.error.message}`,
+          {
+            cause: result.error,
+          }
+        )
+      }
+
+      // TODO: save the delegation into the DelegationsStore
```
alanshaw, line 346: I'm not certain, but I think `createDelegation` does this already...

packages/w3up-client/src/client.js
```diff
@@ -291,9 +292,66 @@ export class Client extends Base {
...
+   * Delegates the following capabilities to the gateway:
+   * - `space/content/serve/*`
+   *
+   * @param {import('./types.js').OwnedSpace} space - The space to authorize the gateway for.
+   * @param {object} [options] - Options for the authorization.
+   * @param {`did:web:${string}`} [options.gateway] - The Web DID of the gateway to authorize. If not provided, the default `did:web:w3s.link` gateway will be authorized.
+   * @param {number} [options.expiration] - The time in seconds to expire the authorization.
+   * @returns {Promise<import('./delegation.js').AgentDelegation<any>>} Resolves with the AgentDelegation instance once the gateway is successfully authorized.
+   */
+  async authorizeGateway(space, options = {}) {
```
alanshaw, line 310: Would also be nice to not tie this to gateways specifically, since in the future we'll want to authorize bitswap nodes.

I'd suggest something like `authorizeContentServe` and instead of `gateway` as the option name perhaps `audience`.

When you create a space I'd have a more specific option like `{ authorizeGateways: ['did:web:w3s.link'], authorizePeers: ['did:web:elastic.dag.house'] }`

packages/w3up-client/src/client.js
alanshaw: You probably need some connection config here right?

packages/w3up-client/src/client.js
Peeja: Hmm. This is a bit tricky because we need to both authorize the server _and_ send that server the delegation. We can know how to speak to our own gateway, but speaking to arbitrary services gets trickier.

That said, perhaps this is a great use case for a [DID service endpoint](https://www.w3.org/TR/did-core/#dfn-serviceendpoint)! If each DID is responsible for specifying its own UCAN endpoint, we can simply talk to that. That also means everywhere in our system we specify a DID and a URL next to each other, we can stop doing that.

That's not a small thing to do—it'll take a fair bit of specification—but it might be a good ideal to point towards. In the meantime, I'd recommend making explicit, separate options for Freeway and Hoverboard in the client, which can become aliases for something more general later on.

packages/w3up-client/test/capability/space.test.js
```diff
@@ -142,7 +142,7 @@ export const SpaceClient = Test.withContext({
         assert.equal(egressRecord.bytes, car.size, 'bytes should be the same')
         assert.equal(
           new Date(egressRecord.servedAt).getTime(),
-          Math.floor(new Date(egressData.servedAt).getTime() / 1000) * 1000,
+          Math.floor(new Date(egressData.servedAt).getTime() / 1000),
```
Peeja, line 145: **question:** What happened with these? Was it just incorrect before? I'm not seeing the corresponding implementation change.

packages/w3up-client/test/client.test.js
```diff
@@ -528,6 +529,45 @@ export const testClient = {
...
+        gateway: 'did:web:staging.w3s.link',
+        expiration: Infinity,
+      })
+      assert.ok(delegation)
+
+      // Step 4: Find the delegation for the default gateway
+      assert.equal(delegation.audience.did(), 'did:web:staging.w3s.link')
+      assert.ok(
+        delegation.capabilities.some(
+          // @ts-expect-error
```
Peeja, line 563: **suggestion:** Let's not use these, they keep biting us later. What's the error here? It looks like a `Capability => boolean` to me—isn't that what it expects?

Peeja (changes requested):

packages/w3up-client/test/capability/space.test.js
fforbeck: This is from another PR that has already been merged. I thought Github action would catch the failing test, but I don't think it is working. I just created a ticket to tackle that later (https://github.com/storacha/w3up/issues/1592), but for now, I executed the tests manually, found the broken test, and fixed.

packages/w3up-client/src/client.js
fforbeck: @alanshaw @Peeja For now, I think we can pass the `connection` option in the function arguments and assume that all the specified DIDs will be authorized, with their delegations stored in the provided service (the one associated with the connection). In the future, we can explore the idea of leveraging the DID service endpoint for this. WDYT?

Peeja: Love it! So the CLI will know about Freeway and Hoverboard (at least in the first iteration), but it stays abstract in the client library. I think that's a great way to go.

packages/w3up-client/test/mocks/service.js
```diff
@@ -0,0 +1,43 @@
...
+          }
+        }
+      ),
+    },
+  }
+}
+
+/**
+ * @param {any} service
+ * @param {any} id
```
Peeja, line 27: **suggestion:** Let's give these real types.

packages/w3up-client/test/capability/space.test.js
Peeja: Weird! Well, good catch, then. 😄

Peeja (dismissed):
Looking good! Just the TypeScript / JSDoc thing.

Not My Type Mexican

packages/w3up-client/test/mocks/service.js
fforbeck: I am not sure if we need it in this case. This is like a generic function to instantiate connections for any type of service with any identifier.

packages/w3up-client/src/client.js
```diff
@@ -250,15 +252,23 @@ export class Client extends Base {
...
+   * In addition, it authorizes the listed Content Serve Services to serve content from the created space.
+   * It is done by delegating the `space/content/serve/*` capability to the Content Serve Service.
+   * User can skip the Content Serve authorization by setting the `skipContentServeAuthorization` option to `true`.
    *
-   * @typedef {object} CreateOptions
-   * @property {Account.Account} [account]
+   * @typedef {object} SpaceCreateOptions
+   * @property {boolean} [skipContentServeAuthorization] - Whether to skip the Content Serve authorization. It means that the content of the space will not be served by any Content Serve Service.
+   * @property {`did:${string}:${string}`[]} [authorizeContentServeServices] - The DID Key or DID Web of the Content Serve Service to authorize to serve content from the created space.
+   * @property {import('./types.js').ConnectionView<import('./types.js').ContentServeService>} [connection] - The connection to the Content Serve Service that will handle, validate, and store the access/delegate UCAN invocation.
```
alanshaw, line 262: Each content serve service will have a different URL. For the `createSpace` function I think we should just require a DID and a URL rather than a connection i.e. something like `authorizeContentServeServices: Array<{ serviceDID: DID, serviceURL: URL }>` option. It's just easier to use. We can then create the connection object internally for `authorizeContentServe`.

I wonder if it is clearer to have `skipGatewayAuthorization` and `authorizeGatewayServices` and we add bitswap as seperate options (`skipPeerAuthroization`, `authorizePeerServices`) later? WDYT?

packages/w3up-client/src/client.js
```diff
@@ -291,9 +301,102 @@ export class Client extends Base {
...
+   * Delegates the following capabilities to the audience:
+   * - `space/content/serve/*`
+   *
+   * @param {import('./types.js').OwnedSpace} space - The space to authorize the audience for.
+   * @param {object} options - Options for the authorization.
+   * @param {`did:${string}:${string}`} options.audience - The Web DID of the audience (gateway or peer) to authorize.
+   * @param {import('./types.js').ConnectionView<import('./types.js').ContentServeService>} options.connection - The connection to the Content Serve Service that will handle, validate, and store the access/delegate UCAN invocation.
+   * @param {number} [options.expiration] - The time at which the delegation expires in seconds from unix epoch.
+   */
+  async authorizeContentServe(space, options) {
```
alanshaw, line 345: This is more an internal function but it would be useful to export for power users. Suggestion: move to exported function `export const authorizeContentServe = (agent, connection, space, options) => ...` and then call from `createSpace`.

packages/w3up-client/src/client.js
```diff
@@ -291,9 +301,102 @@ export class Client extends Base {
...
+  /**
+   * Authorizes an audience to serve content from the provided space and record egress events.
+   * It also publishes the delegation to the content serve service.
+   * Delegates the following capabilities to the audience:
+   * - `space/content/serve/*`
+   *
+   * @param {import('./types.js').OwnedSpace} space - The space to authorize the audience for.
+   * @param {object} options - Options for the authorization.
+   * @param {`did:${string}:${string}`} options.audience - The Web DID of the audience (gateway or peer) to authorize.
+   * @param {import('./types.js').ConnectionView<import('./types.js').ContentServeService>} options.connection - The connection to the Content Serve Service that will handle, validate, and store the access/delegate UCAN invocation.
```
alanshaw, line 342: Not optional - we need this to execute the invocation.

packages/w3up-client/src/client.js
```diff
@@ -291,9 +301,102 @@ export class Client extends Base {
...
+   */
+  async authorizeContentServe(space, options) {
+    const currentSpace = this.currentSpace()
+    try {
+      // Set the current space to the space we are authorizing the gateway for, otherwise the delegation will fail
+      await this.setCurrentSpace(space.did())
+
+      /** @type {import('@ucanto/client').Principal<`did:${string}:${string}`>} */
+      const audience = {
+        did: () => options.audience,
```
alanshaw, line 353: If not set then use `connection.id`

packages/w3up-client/src/client.js
```diff
@@ -291,9 +301,102 @@ export class Client extends Base {
...
   }
 
+  /**
+   * Authorizes an audience to serve content from the provided space and record egress events.
+   * It also publishes the delegation to the content serve service.
+   * Delegates the following capabilities to the audience:
+   * - `space/content/serve/*`
+   *
+   * @param {import('./types.js').OwnedSpace} space - The space to authorize the audience for.
+   * @param {object} options - Options for the authorization.
```
alanshaw, line 340: Options are optional. Otherwise param or config.

```suggestion
   * @param {object} [options] - Options for the authorization.
```

alanshaw (changes requested):

packages/w3up-client/src/client.js
fforbeck: @alanshaw what about `authorizeContentServeServices: Array<{ serviceDID: DID, connection: ConnectionView }>`, the reason for that is: it is better for testing because then we can easily mock the connection. WDYT?

> I wonder if it is clearer to have skipGatewayAuthorization and authorizeGatewayServices and we add bitswap as seperate options (skipPeerAuthroization, authorizePeerServices) later? WDYT?

I think it will be better. Easier to understand.

packages/w3up-client/src/client.js
fforbeck: Yep, makes sense. I was thinking about implementing a feature in the Console App where users can explicitly authorize our Gateway to serve content from existing spaces, then we can use this function.

fforbeck: @alanshaw, I noticed this is part of the Client class already, so a user will have access to it via `client.authorizeContentServe`. Are you thinking about moving it to an exported function in another place?

packages/w3up-client/src/client.js
alanshaw: Ah yes makes sense, what about `Array<ConnectionView | { serviceDID: DID, serviceURL: URL }>`?

packages/w3up-client/src/client.js
alanshaw: i.e. `import { authorizeContentServe } from '@web3-storage/w3up-client'`

The client instance would use the function internally, but power users can still import and use it.

packages/w3up-client/src/client.js
fforbeck: It could be optional to skip the authorization flow. it is useful for testing, and you don't have to set empty arrays + boolean flag to skip it.

fforbeck: Sounds good.

packages/w3up-client/src/client.js
fforbeck: Instead of using the `skipGatewayAuthorization` I think it is simple if we just check the array.

packages/w3up-client/src/client.js
fforbeck: I won't add this for now @alanshaw . We can always go back and export that if needed.

packages/w3up-client/src/client.js
fforbeck: Ok. I've moved forward with the `Array<ConnectionView>` option.

packages/upload-api/src/access/claim.js
```diff
@@ -13,13 +13,15 @@ export const provide = (ctx) =>
 
 /**
  * Checks if the given Principal is an Account.
+ *
```
fforbeck, line 16: fix lint issue

packages/upload-api/src/access/claim.js
```diff
@@ -13,13 +13,15 @@ export const provide = (ctx) =>
...
  * @param {API.Principal} principal
  * @returns {principal is API.Principal<API.DID<'mailto'>>}
  */
 const isAccount = (principal) => principal.did().startsWith('did:mailto:')
 
 /**
  * Returns true when the delegation has a `ucan:*` capability.
+ *
```
fforbeck, line 24: fix lint issue

packages/upload-api/src/lib.js
```diff
@@ -147,8 +147,8 @@ export const execute = async (agent, input) => {
  * a receipt it will return receipt without running invocation.
  *
  * @template {Record<string, any>} S
- * @param {Types.Invocation} invocation
  * @param {Agent<S>} agent
+ * @param {Types.Invocation} invocation
```
fforbeck, line 151: fixed lint issue

alanshaw (approved):

juliangruber:
> I created a new function authorizeContentServe for this implementation and included it in the createSpace flow. This is a breaking change because now the user is forced to provide the DIDs of the Content Serve services, and the connection, or skip the authorization flow.

Do you have any documentation / getting started example for this new behavior?

fforbeck:
> Do you have any documentation / getting started example for this new behavior?

Not yet @juliangruber . This functionality will align with other ongoing changes, like [this issue](https://github.com/storacha/project-tracking/issues/160). Once those are finalized, the documentation will be updated with all the necessary details to explain the new behavior.

juliangruber:
Got it, thanks. For now we shouldn't upgrade to that version, correct?

fforbeck:
> Got it, thanks. For now we shouldn't upgrade to that version, correct?

That's correct. If you keep your `w3up-client` on version [16.5.2](https://www.npmjs.com/package/@web3-storage/w3up-client/v/16.5.2), you won't be required to set the authorization details.

fforbeck:
@juliangruber the latest version of the `w3up-client` (17.1.2) will now authorize the Storacha Gateway to serve the content of the new space if you don't provide anything in the Create Space call options. You can find more info on that here: https://docs.storacha.network/w3up-client (Item # 5).

juliangruber:
@fforbeck thank you! Is there anything I need to do if I want to use the latest client with an existing space, created long before?

fforbeck:
> @fforbeck thank you! Is there anything I need to do if I want to use the latest client with an existing space, created long before?

You're welcome, @juliangruber! No, you don't need to do anything right now. Existing spaces created long ago are considered "legacy" spaces and currently don't require content authorization. However, in the future, we plan to enable some rate limiting for serving content from these legacy spaces. When that happens, we'll provide an option to migrate or authorize your space in advance to ensure a smooth transition.

PR #1595, opened 2024-11-29
fix: overflow slices

alanshaw:
In the unusual case where a sharding stream has to create a new shard because adding the root CID to the CAR header pushes it over the shard limit the slice indexing info for the new shard was set for the blocks in the previous shard, not the blocks in the new shard 🤦, this means that one or more blocks in DAG were not indexed at all.


Peeja (approved):
Shard

PR #1601, opened 2024-12-05
fix: retry when PUT request fails

alanshaw:
FML, we have a useless retry function that does not retry the PUT request when fetch response is non-200.

Not any more.


fforbeck (approved):
Great catch!!

PR #1603, opened 2024-12-09
feat(content-serve): delegate with token caveat

fforbeck:
### Context
We should enable setting a token when delegating the `space/content/serve/*` capability. This would allow us to validate the `token` included in gateway requests and determine whether to serve the content based on the validation result.

### Related Issues
- https://github.com/storacha/project-tracking/issues/213

### DIFF
--- a/packages/capabilities/src/space.js
+++ b/packages/capabilities/src/space.js
@@ -72,6 +72,9 @@ export const allocate = capability({
 export const contentServe = capability({
   can: 'space/content/serve/*',
   with: SpaceDID,
+  nb: Schema.struct({
+    authToken: Schema.string().optional(),
+  }),
   derives: equalWith,
 })
 
--- a/packages/capabilities/src/space.js
+++ b/packages/capabilities/src/space.js
@@ -65,15 +65,53 @@ export const allocate = capability({
 })
 
 /**
- * The capability grants permission for all content serve operations that fall under the "space/content/serve" namespace.
- * It can be derived from any of the `space/*` capability that has matching `with`.
+ * "Manage the serving of content owned by the subject Space."
+ *
+ * A Principal who may `space/content/serve/*` is permitted to perform all
+ * operations related to serving content owned by the Space, including actually
+ * serving it and recording egress charges.
  */
-
 export const contentServe = capability({
   can: 'space/content/serve/*',
+  /**
+   * The Space which contains the content. This Space will be charged egress
+   * fees if content is actually retrieved by way of this invocation.
+   */
   with: SpaceDID,
   nb: Schema.struct({
-    authToken: Schema.string().optional(),
+    /** The authorization token, if any, used for this request. */
+    token: Schema.string().nullable(),
+  }),
+  derives: equalWith,
+})
+
+/**
+ * "Serve content owned by the subject Space over HTTP."
+ *
+ * A Principal who may `space/content/serve/transport/http` is permitted to
+ * serve any content owned by the Space, in the manner of an [IPFS Gateway]. The
+ * content may be a Blob stored by a Storage Node, or indexed content stored
+ * within such Blobs (ie, Shards).
+ *
+ * Note that the args do not currently specify *what* content should be served.
+ * Invoking this command does not currently *serve* the content in any way, but
+ * merely validates the authority to do so. Currently, the entirety of a Space
+ * must use the same authorization, thus the content does not need to be
+ * identified. In the future, this command may refer directly to a piece of
+ * content by CID.
+ *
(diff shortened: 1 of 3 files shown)


packages/capabilities/src/space.js
```diff
@@ -72,6 +72,9 @@ export const allocate = capability({
 export const contentServe = capability({
   can: 'space/content/serve/*',
   with: SpaceDID,
+  nb: Schema.struct({
+    authToken: Schema.string().optional(),
```
Peeja, line 76: **issue:** This should be `token`, not `authToken`. It also isn't optional: it has to be `string | null`. We want people to explicitly opt in to public access.

**thought:** Which is easy enough to fix, but also has me concerned that it's _way_ too easy for wildcard capability definitions to be out of sync with their children. Not sure there's much we can do about that at the moment, though, at a fundamental level.

**suggestion:** Could we at least abstract the `token` part of the schema to something shared between the two? That means moving the transport one over here, which makes sense; they should be in the same place, wherever that is.

Peeja (changes requested):
BTW, I think the issues coming from CI are because `options` itself is optional and doesn't have a default; you may need to give it a default of `{}` for the case where it's not provided at all.

packages/capabilities/src/space.js
```diff
@@ -65,13 +65,54 @@ export const allocate = capability({
...
+ * Invoking this command does not currently *serve* the content in any way, but
+ * merely validates the authority to do so. Currently, the entirety of a Space
+ * must use the same authorization, thus the content does not need to be
+ * identified. In the future, this command may refer directly to a piece of
+ * content by CID.
+ *
+ * [IPFS Gateway]: https://specs.ipfs.tech/http-gateways/path-gateway/
+ */
+export const transportHttp = capability({
+  can: 'space/content/serve/transport/http',
```
alanshaw, line 106: IDK if "transport" is useful, makes it very verbose.

Suggestion: `space/content/serve/http`

alanshaw (changes requested):
I'm not sure it's a good idea to set the `token` on the wildcard capability.

What if a `space/content/serve/*` delegation is required for the gateway to serve data and record egress _at all_. That means that it is up to the service to decide the terms for rate limiting, which may be extremely restrictive eventually.

If you (the user) wants to lift those restrictions or manage your own restrictons you'd also delegate a more specific non-wildcard capability (`space/content/serve/http`) with a `token` (and eventually other optional restrictions, like CID, or IP address or referer for example).

I think this is preferable, because the user will want to create and revoke multiple access tokens, without removing the ability for the gateway to serve data / record egress in general.

You'd then perhaps manage tokens in the client like:

```ts
interface Client {
  // ... existing methods

  // Calls `access/delegate` with `space/content/serve/http`, restricting use
  // with a user defined token.
  // Options allows the service to authorize to be customized, defaulting to w3s.link.
  authorizeContentServeHTTP(restrictions: { token: string }, options: Options): Promise<Result<Delegation>>

  // Calls `ucan/revoke` for the given delegation.
  // Options allows the service to authorize to be customized, defaulting to w3s.link.
  revokeContentServe(delegation: Link, options: Options): Promise<Result<Unit, DelegationNotFound>>
}
```

So, the `createSpace` function would do what it does currently - delegate `space/content/serve/*` to enable general access and egress accounting, but for managing tokens you'd use the new methods.

On the server you can then have a KV mapping of `spaceDID+token: delegation` which you can query to determine if token access is allowed for a space.

packages/capabilities/src/space.js
```diff
@@ -65,13 +65,54 @@ export const allocate = capability({
...
 export const contentServe = capability({
   can: 'space/content/serve/*',
+  /**
+   * The Space which contains the content. This Space will be charged egress
+   * fees if content is actually retrieved by way of this invocation.
+   */
   with: SpaceDID,
+  nb: Schema.struct({
+    /** The authorization token, if any, used for this request. */
+    token: Schema.string().nullable(),
```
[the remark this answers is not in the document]
fforbeck: I agree that defining it at the top level could be overly restrictive. Allowing its use within the `transport/http` capability seems like a good approach, as it enables revocation without affecting the ability to serve content or record egress. @Peeja, what are your thoughts on this?

packages/capabilities/src/space.js
Peeja: I agree. I'm not sure now who originally proposed `transport` and why, but I don't _think_ there was strong reasoning that went with it as much as putting something down.

fforbeck:
Closing this PR for now. Any proposals for handling tokens will be tracked in [issue #213](https://github.com/storacha/project-tracking/issues/213).

PR #1610, opened 2024-12-20
fix(w3up-client): authorize agent to use space

fforbeck:
We need to save the space before setting it as the current space, otherwise the authorization flow will fail because the agent won't be allowed to use that space.

image


fforbeck:
The same change was approved in the `upload-service` repo: https://github.com/storacha/upload-service/pull/115
I will move forward and merge this.

PR #1627, opened 2025-02-19
feat(upload-client): Remove ipfs-utils

v1rtl:
This PR removes redundant ipfs-utils for the upload-client which only uses it in one place. I rewrote the code from the original ipfs-utils to use native fetch and be platform-agnostic.

`pnpm test` passed, while preserving 100% code coverage.

Tested on:
- [x] Node
- [x] Chromium
- [x] Firefox
- [x] Bun
- [x] Deno 
- [ ] Electron
- [ ] React Native


talentlessguy:
There a few more issues. Passing readable stream or setting duplex is not supported on Firefox.

https://bugzilla.mozilla.org/show_bug.cgi?id=1387483
https://bugzilla.mozilla.org/show_bug.cgi?id=1792434

this only happens with FF. I will fallback to XHR only for FF to ensure fetch progress is working correctly.

also, the reason why Node tests passed and Chromium didn't is most likely due to undici fetch using H1.1 by default

packages/upload-client/src/fetch-with-upload-progress.js
```diff
@@ -1,8 +1,160 @@
...
-  return ipfsUtilsFetch.fetch(url, init)
+export const fetchWithUploadProgress = (url, init = {}) => {
+  // https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/nextHopProtocol
+  /**
+   * @type {string | undefined}
+   */
+  const protocol =
+    // @ts-expect-error nextHopProtocol is missing from types but is widely available
+    performance.getEntriesByType('resource')[0]?.nextHopProtocol
+  const preH2 = protocol !== 'h2' && protocol !== 'h2c' && protocol !== 'h3'
```
alanshaw, line 153: Should this be the other way around - specify the protocols that do not support streaming?

packages/upload-client/src/fetch-with-upload-progress.js
```diff
@@ -1,8 +1,160 @@
...
+
+  /* c8 ignore next 3 */
+  if ((isBrowser || preH2) && !isNode) {
```
alanshaw, line 156: The test for `isBrowser` is an existence check for XHR, which could be `false`, but presumably `preH2` could be `true` and we could also not be in Nodejs...

We should not drop in here if XHR is not available.

packages/upload-client/src/types.ts
```diff
@@ -87,10 +83,13 @@ import {
...
 
 type FetchOptions = Override<
-  IpfsUtilsFetchOptions,
+  RequestInit,
   {
-    // `fetch` is a browser API and browsers don't have `Readable`
-    body: Exclude<IpfsUtilsFetchOptions['body'], import('node:stream').Readable>
+    /**
+     * Can be passed to track upload progress.
+     * Note that if this option in passed underlying request will be performed using `XMLHttpRequest` and response will not be streamed.
```
alanshaw, line 90: Note: Response not streaming is not an issue - we don't download anything big in the client.

alanshaw (changes requested):

packages/upload-client/src/fetch-with-upload-progress.js
talentlessguy: I could make it the other way around. the others are http/0.9, http/1.0 and http/1.1 I think

nextHopProtocol has a fixed set of values

packages/upload-client/src/fetch-with-upload-progress.js
talentlessguy: This might get confusing so I'll list the conditions/gotchas here:

1. Node.js fetch aka undici by default doesn't directly support HTTP/2, but supports body streaming either way.

2. Browser fetch supports streaming only in HTTP/2 (AFAIK)

3. Deno fetch is the same

4. Bun doesn't support HTTP/2 yet

5. All browsers implement XHR, so it's a safe assumption

perhaps a better set of conditions could be handled, not sure which one to handle all these cases (given that for some platforms streaming is available in both. for some in none and for others only in one)

also, in order to test this more thoroughly, a test case with HTTP/2 server perhaps should be added, but I haven't written one yet (should be trivial enough)

alanshaw:
This LGTM barring the CI failure for upload client...

packages/upload-client/src/fetch-with-upload-progress.js
```diff
@@ -1,8 +1,170 @@
...
+        const { done, value } = await reader.read()
+        if (done) break
+
+        loaded += value.byteLength
+        yield value // Yield the chunk
+        onUploadProgress({ total, loaded, lengthComputable })
+      }
+      // eslint-disable-next-line no-useless-catch
+    } catch (e) {
+      throw e
```
alanshaw, line 53:
```suggestion
```

packages/upload-client/src/fetch-with-upload-progress.js
```diff
@@ -1,8 +1,170 @@
...
+const withUploadProgress = (options) => {
+  const { onUploadProgress, body } = options
+
+  if (!onUploadProgress) {
+    return options
+  }
+
+  const rsp = new Response(body)
+  // @ts-expect-error web streams from node and web have different types
+  const source = iterateBodyWithProgress(rsp.body, onUploadProgress)
```
alanshaw, line 78: Why not pass `body` directly to `iterateBodyWithProgress`? Can you add a comment if this is necessary or change to pass directly? 🙏

packages/w3up-client/package.json
```diff
@@ -143,7 +143,7 @@
     "mock:gateway-server": "PORT=5001 node test/helpers/gateway-server.js",
     "coverage": "c8 report -r html && open coverage/index.html",
     "rc": "npm version prerelease --preid rc",
-    "docs": "npm run build && typedoc --out docs-generated"
+    "docs": "pnpm build && typedoc --out docs-generated"
```
alanshaw, line 146: Can we please not change these? If someone clones the repo and does not use pnpm we still want things to work.

alanshaw (changes requested):

PR #1629, opened 2025-02-24
feat: external login

alanshaw:
refs https://github.com/storacha/RFC/pull/43

This PR adds a new method `externalLogin()` that allow login to the system via an external method (e.g. OAuth). It essentially bypasses email confirmation, assuming this is done out of band. We re-join the regular auth flow at the point when we claim the delegation (and attestation) from an email to the agent.

### DIFF
--- a/packages/capabilities/src/access.js
+++ b/packages/capabilities/src/access.js
@@ -41,7 +41,7 @@ export const AuthorizationRequest = Schema.struct({
   /**
    * DID of the Account authorization is requested from.
    */
-  iss: Account,
+  iss: Account.optional(),
   /**
    * Capabilities agent wishes to be granted.
    */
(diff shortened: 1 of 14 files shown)


packages/capabilities/src/access.js
```diff
@@ -41,7 +41,7 @@ export const AuthorizationRequest = Schema.struct({
   /**
    * DID of the Account authorization is requested from.
    */
-  iss: Account,
+  iss: Account.optional(),
```
alanshaw, line 44: The is now optional, as per the RFC.

In the case of OAuth login, we don't know the email address of the user until after they have authorized and we have issued the delegation and attestation on their behalf.

fforbeck (approved):
Looks good to me!

PR #1640, opened 2025-04-30
fix: large directory upload (100K files)

fforbeck:
### Context
This PR addresses a series of stack overflow issues encountered when processing large datasets (100K+ files). Due to the difficulty in simulating the exact conditions that caused the "Maximum call stack size exceeded" error, I took a trial-and-error approach, working closely with the user to validate each fix. I focused on improving the code paths that had the highest probability of causing a stack overflow, based on the error patterns and user feedback.

The fixes were implemented and tested incrementally:
1. First, addressing the UnixFS directory processing
2. Then the ShardedDAGIndex archive process
3. Finally, the Index Add process

#### Problem
- The `UnixFSDirectoryBuilder` in `unixfs.js` was using recursive directory building
- With 100K files, this created a huge structure in memory
- The stack overflow occurred during the `finalize` operation
- This prevented the successful processing of large directory structures

#### Solution
- Replaced recursive directory building with an iterative approach
- Implemented explicit stack/queue for directory traversal
- Process directories and files in a breadth-first manner
- Maintain proper ordering while avoiding deep recursion
- Added memory limits and checks during processing

#### Tests
- ✅ The user was able to test the upload, and the process failed in a different stage, after the encoding.

#### Problem
- The `archive` function in `sharded-dag-index.js` was processing all shards and slices at once
- With 100K files, this created large arrays in memory during sorting operations
- The stack overflow occurred when trying to sort these large arrays
- This prevented successful uploads of datasets larger than ~50K files

#### Solution
- Implemented batch processing with configurable thresholds:
  ```javascript
  const LARGE_DATASET_ARCHIVE_THRESHOLD = 50_000
  const ARCHIVE_BATCH_SIZE = 10_000
  ```
- Process shards in batches of 10,000
- Process slices within each shard in batches
- Sort operations are performed on smaller chunks of data
- Maintains data integrity while preventing memory issues
- Preserves the sequential process for fewer than 50K shards

#### Tests
- [ ] Waiting for user feedback

#### Problem
- The `add` function in `upload-api/src/index/add.js` was processing all shard allocations concurrently
- With 100K files, this created 100K concurrent promises
- The JavaScript stack was overwhelmed by the number of concurrent operations
- This caused a stack overflow during the index registration phase

#### Solution
- Implemented batch processing for shard allocations:
  ```javascript
  const ALLOCATION_BATCH_SIZE = 10_000
  ```
- Process shard allocations in batches of 10,000
- Use `Promise.all` within each batch for concurrent processing
- Maintain proper error handling and propagation
- Prevents stack overflow while maintaining good performance

#### Tests
- [ ] Waiting for user feedback

### DIFF
--- a/packages/blob-index/src/sharded-dag-index.js
+++ b/packages/blob-index/src/sharded-dag-index.js
@@ -9,6 +9,17 @@ import { DigestMap } from './digest-map.js'
 
 export const version = 'index/sharded/dag@0.1'
 
+/**
+ * The threshold for the number of shards in a dataset that triggers the large dataset archive path.
+ * This is a heuristic to avoid memory issues when archiving large datasets.
+ */
+const LARGE_DATASET_ARCHIVE_THRESHOLD = 50_000
+/**
+ * The size of the batch to process when archiving large datasets.
+ * This is a heuristic to avoid memory issues when archiving large datasets.
+ */
+const ARCHIVE_BATCH_SIZE = 10_000
+
 export const ShardedDAGIndexSchema = Schema.variant({
   [version]: Schema.struct({
     /** DAG root. */
@@ -165,6 +176,14 @@ export const create = (content) => new ShardedDAGIndex(content)
  * @returns {Promise<API.Result<Uint8Array>>}
  */
 export const archive = async (model) => {
+  // Check if we're dealing with a large dataset
+  const totalEntries = model.shards.size
+
+  if (totalEntries > LARGE_DATASET_ARCHIVE_THRESHOLD) {
+    return await archiveLargeDataset(model)
+  }
+
+  // Original fast path for normal cases
   const blocks = new Map()
   const shards = [...model.shards.entries()].sort((a, b) =>
     compare(a[0].digest, b[0].digest)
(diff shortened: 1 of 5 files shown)


packages/blob-index/src/sharded-dag-index.js
```diff
@@ -188,3 +207,63 @@ export const archive = async (model) => {
...
+      const bytes = dagCBOR.encode([s[0].bytes, mappedSlices])
+      const digest = await sha256.digest(bytes)
+      const cid = Link.create(dagCBOR.code, digest)
+      blocks.set(cid.toString(), { cid, bytes })
+      index.shards.push(cid)
+    }
+  }
+
+  // Verify we processed all shards
+  if (index.shards.length !== totalShards) {
```
alanshaw, line 259: Should be a test not production code.

packages/blob-index/src/sharded-dag-index.js
```diff
@@ -165,6 +176,14 @@ export const create = (content) => new ShardedDAGIndex(content)
  * @returns {Promise<API.Result<Uint8Array>>}
  */
 export const archive = async (model) => {
+  // Check if we're dealing with a large dataset
+  const totalEntries = model.shards.size
+
+  if (totalEntries > LARGE_DATASET_ARCHIVE_THRESHOLD) {
+    return await archiveLargeDataset(model)
```
alanshaw, line 183: Why would we keep both implementations? I'd replace with this one unless there is a performance penalty.

packages/blob-index/src/sharded-dag-index.js
fforbeck: Oops. Sorry. Leftovers from the debugging process.

packages/blob-index/src/sharded-dag-index.js
fforbeck: Good point. I guess I'll replace it with the new one. I haven't run a performance check, but I was thinking more about preserving the existing behavior for smaller sets, given that using batches adds some overhead in general.

